PR #23, opened 2018-12-26
Updates to user-visible <string> class text across .ui files

StrontiumZ38:
I've gone through the .ui files found in the veil-master/src/qt/forms and veil-master/src/qt/evil/forms folders and made changes to spelling and grammar, instances of Bitcoin, and small improvement where necessary.


veildev:
Some of the files are not used but still good.

PR #156, opened 2019-01-02
[Wallet, AddressBook, QT]  bench32 stored on db fix + invoque method not parsing bench32

veildev:


presstab:
Will you remove the `std::cout`'s before we merge in?

veildev:
sure, forgot those in the rush.

PR #183, opened 2019-01-04
[QT] Settings minting - CT and Basecoin balances + css fixes.

veildev:


presstab:
Need to get rid of 3590870

veildev:
Opening a new PR to clean this.

PR #198, opened 2019-01-05
dont branch based on potentially uninit var

barrystyle:
seems to prevent segfault with local stratum; evidenced by waves of nodes reannouncing after connectblock errors..

### DIFF
--- a/src/primitives/block.h
+++ b/src/primitives/block.h
@@ -183,6 +183,7 @@ public:
         hashMerkleRoot = uint256();
         hashWitnessMerkleRoot = uint256();
         hashPoFN = uint256();
+        fSignaturesVerified = false;
     }
 
     CBlockHeader GetBlockHeader() const


src/validation.cpp
```diff
@@ -2306,7 +2306,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
             if (tx.IsZerocoinSpend()) {
                 int64_t nTimeSpendCheck = GetTimeMicros();
                 // Skip signature verification if it's already been done or if the block height is below a checkpoint height
-                bool fSkipSigVerify = block.fSignaturesVerified ? true : fSkipComputation;
+                bool fSkipSigVerify = fSkipComputation;
```
presstab: We need to skip the signature verification if the block has already had its signatures verified (slows down syncing significantly if not). If the concern is about `block.fSignaturesVerified` potentially being used while unitialized, correct thing to do would be adding:
`fSignaturesVerified = false;`
To `src/primitives/block.h` `CBlock::SetNull()`.
Good catch.

barrystyle: Corrected as suggested, much better done this way..

PR #219, opened 2019-01-07
Fix a wide range of transaction issues.

presstab:
**Fixed:**
- Sending zerocoin/RingCT/CT from the GUI send dialog
- Zerocoin mint from CT and RingCT.
- Deadlock when minting zerocoins.
- Zerocoinspend tx meta data not correctly tied to tx because of the txhash changes after signature.
- Basecoin balances
- CT Balances
- Properly select CT/RingCT output type for change situations.
- Allow Basecoin checkbox and rpc flag now used as "Use Basecoins" which will specifically only use basecoins for that tx.
- Automatically archive bad mints if possible.
- Mark anon output as spent if during spending the key image is found on the blockchain DB.
- Mark CT inputs as spent when spending them.
- Don't try to add loose coinstake transactions to mempool.

**Problems not solved:**
- Still seeing a blockchain corruption periodically when restarting, can't yet find the exact cause.

### DIFF
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -268,14 +268,11 @@ public:
     bool getPrivKey(const CKeyID& address, CKey& key) override { return m_wallet.GetKey(address, key); }
     bool isSpendable(const CTxDestination& dest) override { return IsMine(m_wallet, dest) & ISMINE_SPENDABLE; }
 
-    std::string mintZerocoin(
-            CAmount nValue,
-            std::vector<CDeterministicMint>& vDMints,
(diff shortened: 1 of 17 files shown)


StrontiumZ38:
Looks promising. Let me know if you want me testing/helping recreate on testnet wallet.

presstab:
@StrontiumZ38 sure the more help testing the better. Have tested everything many times on testnet. Problem is that is using a clean wallet.dat without some of the old data from previous builds and such.
Most help would be minting, spending, using all of the `sendxtox` rpc calls. Noticing balances using `getbalances` and making sure everything looks how it should.

PR #251, opened 2019-01-09
UI text edits

gabrielnergaard:


src/qt/veil/forms/receivewidget.ui
```diff
@@ -139,7 +140,11 @@ font-size:16px;</string>
            </property>
            <property name="text">
             <string>A sender can copy the address or scan the 
+<<<<<<< HEAD
```
presstab: Fix this.

src/qt/veil/forms/tooltipbalance.ui
```diff
@@ -134,7 +134,7 @@ font-size:12px;
 text-align:center;</string>
         </property>
         <property name="text">
-         <string>RingCT &amp; CT</string>
+         <string>Basecoin (Shielded) </string>
```
presstab: This is not basecoin.

4x13:
You cant just make changes you think should be done without understanding the differences.

src/qt/veil/forms/receivewidget.ui
gabrielnergaard: Keep in mind I use the Mac OS X GitHub Desktop app, and have very minimal experience. (For the past years I've been managing project, rather than directly working in code.)

I searched my local repository, found the thing you asked to be fixed, fixed, and committed my fix. I have no idea whether that fix will now be seen in this PR, or if there is something additional I need to do to get it into the PR.

Issue #279, opened 2019-01-12
getnewminingaddress adding unrelated new address to address book 1.0.0.9


StrontiumZ38:
**Explain issue**
creating a new mining address from console adds additional address, separate from the created mining address to address book.
**To recreate**
Open console > getnewminingaddress > go to address book and find new, different address than that just created and output in console starting with 1 (on testnet)

PR #291, opened 2019-01-14
[Wallet] Getnewminingaddress storing address on addressbook badly fix + zcspend tx creation error showed on screen.

veildev:


src/wallet/rpcwallet.cpp
```diff
@@ -249,8 +249,8 @@ static UniValue getnewminingaddress(const JSONRPCRequest& request)
 
     CKeyID dest = newKey.GetID();
     pwallet->LearnRelatedScripts(newKey, OutputType::LEGACY);
-
-    pwallet->SetAddressBook(dest, label, "receive");
+    
+    pwallet->SetAddressBook(dest, label, "basecoin");
```
presstab: `setnonautomintaddress` uses the label `basecoin` for the address that is set to not be eligible for automint. Need to use a different label here.

furszy: Hmm, the GUI has been creating miner addresses and formatting them on the correct format using the basecoin tag. Check `addressreceive.cpp` file, generateNewAddress method.

presstab: Probably the reason why automint isn't working for many people.

furszy: We need to be sync on this kind of points, the basecoin tag purpose was added 16 days ago on the GUI.

4x13: Let just remove the mining addresses from the gui entirely. It was originally created by blondfrogs for the pools, but seems the pools chose to just continue with basecoin until mining to SA's is ready

4x13: @furszy I know, I know.. I asked u to add that button.. :(

furszy: 🤷‍♂️ .

If mining pools continued with basecoin addresses and the goal is move to stealth addresses, why we are having "mining addresses" at all?

Not only GUI, should be removed from the RPC commands too.

StrontiumZ38: Would this discussion above invalidate #279 ?

furszy: Yes @StrontiumZ38

4x13: I believe the pools themselves are using the mining addresses based on seeing where the pool transactions are being distributed from, so for now,r emoving it from the GUI would be fine (and yes, that is my fault) We should check with @blondfrogs for more detailed info on the need of the Mining address.

furszy: all good, we were all under big pressure on that time.

My point goes in the direction of: 
There is no need to keep a functionality and support it when it's not part of the main goal.
Only reason that comes to my mind is that mining pools don't have bech32 support.

But well, @blondfrogs we are all ears.

blondfrogs: getminingaddress was created, because I wanted to build a simple mechanism for pools to use to generate address that they could create coinbase transactions with. Pools run a check to make sure the the addresses are base58 encoded. This means the the normal getnewbasecoinaddress wasn't working, because bech32 address don't pass this check.

As far as I know, pools should be the only ones that are using this rpc call. everyone else should be able to use getnewbasecoinaddress

blondfrogs: I think removing it from the GUI completely is a great move. However, we should keep the rpc call until mining to Stealth Addresses is complete and implemented on the pool side.

furszy:
Completed on #341 . Closing this PR.

PR #303, opened 2019-01-16
[QT] RingCT balance added to Minting Screen.

veildev:


presstab:
Need to get rid of 89b0478

presstab:
1be2813 is in a different PR

furszy:
@presstab done. 
My fault, didn't check it and just pushed it.

PR #329, opened 2019-01-24
fix for issue #323

### LINKED ISSUE #323: Addressbook in Send tab says "bitcoin addresses" and show no addresses (GUI Issue)
Compiled from the actual master:

Addressbook in Send tab says "These are your bitcoin addresses..." instead of "These are your Veil addresses..."
Beside that no adresses are listed even if the Addressbook contains addresses

bitcoin adresses

Ubuntu 18.04LTS

MisterKangaroo:
(fixes #323)


MisterKangaroo:
Change "Bitcoin address" into "Veil address"

PR #332, opened 2019-01-25
[WIP] Fixing zerocoin minting fees and dealing with dust.

ncnight:
Changed the dust limit to 1000 instead of the usual zerocoin mint fee or 3000.


ncnight:
The dust limit is now 1000 and minting zerocoin from basecoin uses that dust limit rather than the zerocoin mint fee. Additionally, CT and ringCT to zerocoin now use the minimum of the calculated zerocoin fee or transaction fee.

PR #379, opened 2019-02-10
[WIP] Add Zerocoin Precomputing

blondfrogs:
- Precomputing is enabled by **default**
- Precomputing is used to make spending zerocoin smoother and quicker
- Precomputing should decrease orphan rate of stakes by making spends take less time
- You can disable precomputing by passing **-precompute=0** as a startup flag or by putting it in the veil.conf
- Precomputing uses it own database which can be cleared manually by deleting the database file in the data directory. Or by using the rpc call **clearspendcache**
- You can view the precomputing percentages by using the rpc call **showspendcaching**
**Update: Use this branch at your own will. This branch is currently being tested, and is not fully supported by the Dev Team yet**

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -231,6 +231,7 @@ BITCOIN_CORE_H = \
   wallet/feebumper.h \
   wallet/fees.h \
   wallet/init.h \
+  wallet/lrucache.h \
   wallet/rpcwallet.h \
   wallet/wallet.h \
   wallet/walletbalances.h \
@@ -331,6 +332,7 @@ libbitcoin_wallet_a_SOURCES = \
   wallet/feebumper.cpp \
   wallet/fees.cpp \
   wallet/init.cpp \
+  wallet/lrucache.cpp \
   wallet/rpcdump.cpp \
   wallet/rpcwallet.cpp \
   wallet/rpczerocoin.cpp \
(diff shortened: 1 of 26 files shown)


CryptoMaximalist:
Created issue: https://github.com/Veil-Project/veil/issues/384

blondfrogs:
Fixed an issue where precompute thread wasn't releasing the lock needed to stakes to be created, and sometime zerocoinspends to be spent. This only happen occasionally and wasn't 100% reproducible. 
Commit **61dcfe7294e6e9383fe3b7252be12040586fcc4d**

blondfrogs:
Update: Previous fix didn't work. 
After investigation: It seems that the precompute thread which handles the precomputation of coinwittnesses is being stopped when a function of pwallet is called. For example 
**pwallet->SpendZerocoin**
This means that everytime the pwallet tries to create a spendzerocoin transaction, if the precompute thread has the cs_spendcache locked, the spendzero call will fail. To combat this I have create a temporary fix that will retreive the cs_spendcache lock before the pwallet makes its spendzerocoin transaction call. This should allow for precomputing to work with spendingzerocoin.

I have also, removed the cs_spendcache lock check on the GetSpendCache call which should now allow stakes to use the cache without having to worry about being locked out. This should make it so stakes don't fail as often as before.

CryptoMaximalist:
@blondfrogs In full recognition that this is still in development, I tried out the latest build and had a different issue.  I'm back to 15 now but I figured I should pass it along as an FYI.  All my stakes i hit were orphaned, but still on the BE, but sometimes indicated confirmations anyway.  Also they had no reward tx for winning the stake.  This happened 3+ times https://discordapp.com/channels/481997816982405130/510598374094012427/546344660554219530

blondfrogs:
Linking to #427

PR #387, opened 2019-02-12
[QT] Fix hiding of orphans.

presstab:
#381 seems to have hidden all stakes, not just orphans.


furszy:
Ok yeah, didn't see this and pushed it twice in #388 too.

PR #413, opened 2019-02-27
[Consensus] fix superblock consensus bugs

random-zebra:
This fixes two problems with contextual checks on superblock.
- bad-coinbase-vpout would be triggered for valid PoS superblocks  (mainnet and testnet) after the  first one.
- CheckBudgetTransaction does not reflect the changes introduced with https://github.com/Veil-Project/veil/pull/299 (only testnet)


furszy:
utACK ❤️

PR #427, opened 2019-03-07
Add Precomputing of Zerocoin Spends

blondfrogs:
- Precomputing is **disabled by default**
- Precomputing is used to make spending zerocoin smoother and quicker
- Precomputing should decrease orphan rate of stakes by making spends take less time
- You can start or stop it by running the rpc call **startprecomputing** or **stoprecomputing**
- View precomputing progess with rpc call **showspendcache**
- Set the precompute blocks per cycle with rpc call **setprecomputeblockspercycle**
- Get the current blocks per cycle with rpc call **getprecomputeblockpercycle**
- Clear your cache and precompute database with **clearspendcache**
- Enable precompute always by starting with or adding the flag **precompute=1** into your veil.conf
- Set the precomputeblockpercycle with by starting with or adding the flag **precomputeblockpercycle=number** into your veil.conf
- Precomputing uses it own database which can be cleared manually by deleting the database file in the data directory. Or by using the rpc call **clearspendcache**

**Update: Use this branch at your own will. This branch is currently being tested, and is not fully supported by the Dev Team yet**

Solves Issues: #447 #448 #449 #450 

**Update: Added QT checkbox to turn off an on precomputing 51b5e37495f57d12c2a11447c29c71b7423c1c99
Screenshot_20190321_100410
Screenshot_20190321_100502
Screenshot_20190321_100518

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,8 @@ include_directories(src/zmq)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Core Gui)
 
 add_executable(veil
+        src/veil/zerocoin/witness.h
+        src/veil/zerocoin/witness.cpp
         src/bench/base58.cpp
         src/bench/bech32.cpp
         src/bench/bench.cpp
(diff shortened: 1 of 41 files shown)


blondfrogs:
Cross Compiling for windows throws this error:

wallet/wallet.cpp: In function ‘void ThreadPrecomputeSpends()’:
wallet/wallet.cpp:6484:25: error: ‘interruption_point’ is not a member of ‘boost::this_thread’
     boost::this_thread::interruption_point();

MisterKangaroo:
more a feature request than an issue:

The precompute process kinda bloats the debug.log. Would it be possible or make sense to move the logs into a seperate file (ie. precompute-debug.log)?

blondfrogs:
Removed logs that were spamming the debug.log

blondfrogs:
Updated the clearspendcache rpc call. Sometimes during the call, veil would crash because of pointer access failures. beb122b94919c6ba144735133940b29667d5b928

PR #451, opened 2019-03-20
Update init.cpp

StrontiumZ38:
Spelling fix and two case fixes.


blondfrogs:
ACK - But please wait until #427 is merged, as it also changes init.cpp

StrontiumZ38:
Sorry, I didn't get around to checking this was good to go yet. Was just about to do now but looks like it's already been handled.

PR #473, opened 2019-04-03
Fix crash with -disablewallet

### LINKED ISSUE #351: Running daemon with -disablewallet kills the daemon in end of sync
When running the daemon with -disablewallet, it kills the daemon, and you have start the sync from snapshot date all over again.

No output from debug except last UpdateTip:....

Using latest source master, also happend on earlier versions

PeterL73:
Fixes #351

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -14,6 +14,7 @@
 #include <addrman.h>
 #include <amount.h>
 #include <veil/ringct/blind.h>
+#include <veil/ringct/stealth.h>
 #include <chain.h>
 #include <chainparams.h>
 #include <checkpoints.h>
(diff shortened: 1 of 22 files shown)


src/init.cpp
```diff
@@ -1871,9 +1871,12 @@ bool AppInitMain()
 
     g_wallet_init_interface.Start(scheduler);
 
+#ifdef ENABLE_WALLET
```
presstab: This is preprocessor, so only changes the way the wallet is compiled. Whereas `-disablewallet` is a runtime option and won't be able to take advantage of this change. I would suggest checking for the startup arg directly:
`gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET)` 
The above code would return true if the wallet is disabled.

src/validation.cpp
```diff
@@ -2935,11 +2935,13 @@ static void AppendWarning(std::string& res, const std::string& warn)
 /** Check warning conditions and do some notifications on new chain tip set. */
 void static UpdateTip(const CBlockIndex *pindexNew, const CChainParams& chainParams) {
 
+#ifdef ENABLE_WALLET
```
presstab: I would use the same logic described above.

presstab:
Worked for me after the new changes. Thanks @PeterL73 !

PR #481, opened 2019-04-07
Add zerocoin to coincontrol

### LINKED ISSUE #431: Add zerocoin denoms to the coin control menu
When adding zerocoin denoms to the coin control menu, there should also be a column to indicate the progress ofprecomputing for each specific denom.

blondfrogs:
- Add ability to select zerocoin denomination from coin control when spending
- Coin control will show zerocoin precomputation percentage if it is enabled and running 

**Precomputation isn't running**
image

**Precomputation is running**
image

Closes #431

### DIFF
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -24,6 +24,7 @@ enum OutputTypes
     OUTPUT_CT               = 2,
     OUTPUT_RINGCT           = 3,
     OUTPUT_DATA             = 4,
+    OUTPUT_ZC               = 5,
 };
 
 enum TransactionTypes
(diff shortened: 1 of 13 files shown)


src/primitives/transaction.h
```diff
@@ -24,6 +24,7 @@ enum OutputTypes
     OUTPUT_CT               = 2,
     OUTPUT_RINGCT           = 3,
     OUTPUT_DATA             = 4,
+    OUTPUT_ZC               = 5,
```
presstab: A zerocoin output is a standard output as of now. I think we should leave this part of the change out, unless we do a full analysis of what problems adding an additional output type could cause in the validation code.

blondfrogs: Removed in 418743623784a5ae2a366f0a9a65af1cfc921d20

PR #485, opened 2019-04-08
[Qt] Add hyperlink for txid in transaction detail view.

### LINKED ISSUE #377: Feature Request — Link transaction ID to the Explorer URL
It would be very convenient for the user if the transaction ID listed on transaction pop-up windows was hyperlinked to the associated URL at the Veil Block Explorer:

https://explorer.veil-project.com/tx/{txid}

drag0ng:
Fixes #377

### DIFF
--- a/src/qt/veil/transactiondetaildialog.cpp
+++ b/src/qt/veil/transactiondetaildialog.cpp
@@ -10,6 +10,8 @@
 #include <qt/optionsmodel.h>
 #include <QDateTime>
 #include <QSettings>
+#include <QDesktopServices>
+#include <chainparams.h>
 
 TransactionDetailDialog::TransactionDetailDialog(QWidget *parent, TransactionRecord *rec, WalletModel *walletModel) :
     QDialog(parent),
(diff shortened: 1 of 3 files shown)


src/qt/veil/transactiondetaildialog.cpp
```diff
@@ -39,8 +40,13 @@ TransactionDetailDialog::TransactionDetailDialog(QWidget *parent, TransactionRec
     ui->textSize->setProperty("cssClass" , "text-detail");
     ui->textStatus->setProperty("cssClass" , "text-detail");
     connect(ui->btnEsc,SIGNAL(clicked()),this, SLOT(onEscapeClicked()));
+    connect(ui->explorerButton,SIGNAL(clicked()),this, SLOT(onExplorerClicked()));
+
 
     if(rec) {
+        QString baseStr = "https://explorer.veil-project.com/tx/";
```
PeterL73: Shouldn't this address be chain dependent?

presstab: Good suggestion. @drag0ng the testnet explorer is: https://testnet.veil-project.com

drag0ng:
Added testnet link too.

Fuzzbawls:
the GUI's settings/options area has a user-adjustable input that can be used for this, rather than hard coding any URLs. connecting that may be a safer route (not needing to push out a new release if the URLs ever change or go down), and would allow users to link to the BE of their choosing.

drag0ng:
> the GUI's settings/options area has a user-adjustable input that can be used for this, rather than hard coding any URLs. connecting that may be a safer route (not needing to push out a new release if the URLs ever change or go down), and would allow users to link to the BE of their choosing.

Good idea. But if the text has multiple urls separated by '|', then we'd have to just open the first url.
Also without hardcoding this in settings, the users opening the wallet first time will all have to input the block explorer url here. 

Most of the wallet users just stay in the main window dealing with "send", "receive" and "address book". When they click a transaction link, they expect this to open a working block explorer url. Instead they'd think of this as a bug and ask us and then we'd have to tell them to save "this url" by going to Settings->Preferences->Display . I'm not sure that we should do this right from the start. Maybe after users getting accustomed, we can push an update explaning things?

ncrypter90:
Bug: tested environment Win 10 64
1. Issue: Hyperlink button available next txtid, however copying txtid doesnt work properly.
only if you click on the txtid and crt+S(select all) then copy using crt+C then we can copy, right click options doesnt work, no response even if you press the options of right click(copy and select all)
2.Observation: While sending coin after pressing send wallet hangs for more than 40secs.

Detailed gif : ( be patient its quite long as i recorded wallet hang time as well :))
Veil_build_485_Screengrab

crypt0-fusi0n:
@drag0ng the new commit has a conflict and is erroring out during building. can you please fix?

crypt0-fusi0n:
Hi @drag0ng , what's the status on this one? Thanks

ncrypter90:
QA passed , Tested on Win 64 , Ubuntu and Mac OS.

Issue #405, opened 2019-02-21
Corruption because of setBlockIndexCandidates.empty() assertion.


gabrielnergaard:
Running OS X Mojave, wallet version 1.15, the app frequently gets into a state (twice this week) in which simply quitting and then trying to restart results in a crash. When this happens, the only solution is to start fresh from a snapshot.

From the OS X crash reporter, this is found:

Application Specific Information:
Assertion failed: (!setBlockIndexCandidates.empty()), function PruneBlockIndexCandidates, file validation.cpp, line 3265.

I haven't found a way to reproduce the error on demand. There have, however, been other reports of this in both the Veil internal support channel, and on the public side.

onezetty:
Same happened to me at Ubuntu 16.04 

```
➜  bin veil-qt: validation.cpp:3265: void CChainState::PruneBlockIndexCandidates(): La declaración `!setBlockIndexCandidates.empty()' no se cumple.

[1]  + 15407 abort (core dumped)  ./veil-qt

```

presstab:
I saw this once today, was not sure if it was from me switching between different branches... Will see if I can get it to happen again and save my datadirectory with the faulty state for testing.

CryptoMaximalist:
I have 2 data directories with this error now, one of which is my snapshot node.  Both ubuntu 18.04 x64.  Let me know on discord if I can get these to you to help @presstab

Iconoclasta:
Likewise, compiled from master - `user@ip:~$ ./veild
veild: validation.cpp:3265: void CChainState::PruneBlockIndexCandidates(): Assertion `!setBlockIndexCandidates.empty()' failed.
Aborted (core dumped)`

presstab:
Seems to be an issue that causes the block indexes on load to have the number of tx's zeroed, then causing issues later in the loading process to arise because of it. Have not found a solution other than running with `-reindex-chainstate`

Elysiumcore:
The same just happened to me. A simple shut down in Mac OS Mohave. Wallet will not restart.  will have to resync chain

veilgets:
Can this be closed? It's fixed.

codeofalltrades:
Closing. Included in release v1.0.3.0

PR #487, opened 2019-04-09
Do not load corrupted duplicate block indexes.

### LINKED ISSUE #422: Startup Crash: chain.cpp:105: const CBlockIndex* CBlockIndex::GetAncestor(int) const: Assertion `pindexWalk->pprev' failed.
After a graceful shutdown, during a blockchain sync, the following startup error appears:

```
veil-qt: chain.cpp:105: const CBlockIndex* CBlockIndex::GetAncestor(int) const: Assertion `pindexWalk->pprev' failed.
Aborted (core dumped)
```

Expected: Wallet could start normally or handle issues

Actual: Error above, hard crash

Reliability: Every time once it happens

Version: Veil 1.0.2, Ubuntu 18.04 x64

presstab:
fixes #422
It is also possible that #405 is related, but not positive.

For some reason multiple indexes are being saved to disk for the same block. The second index will overwrite the first index on first load. The second index has `nTx` set to `0`, as well as all relevant block file locations. The second index causes all subsequent indexes to have `nChainTx` set to `0`, causing them not to be able to be considered block candidates. Also notable is that the second index has the wrong database key which is supposed to be the block hash but is not.

This change simply ignores the second block index on first load. It is still yet to be determined what causes the duplicate index.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -266,6 +266,14 @@ bool CBlockTreeDB::LoadBlockIndexGuts(const Consensus::Params& consensusParams, 
         if (pcursor->GetKey(key) && key.first == DB_BLOCK_INDEX) {
             CDiskBlockIndex diskindex;
             if (pcursor->GetValue(diskindex)) {
+                // For some reason Veil has a tendency to duplicate an index, and store the second under a different key
+                // ignore any duplicates and mark them to be erased
+                uint256 hashBlock = diskindex.GetBlockHash();
+                if (hashBlock != key.second) {
+                    pcursor->Next();
+                    continue;
+                }
+
                 // Construct block index object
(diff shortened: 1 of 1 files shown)


src/txdb.cpp
```diff
@@ -258,6 +258,8 @@ bool CBlockTreeDB::LoadBlockIndexGuts(const Consensus::Params& consensusParams,
     std::unique_ptr<CDBIterator> pcursor(NewIterator());
 
     pcursor->Seek(std::make_pair(DB_BLOCK_INDEX, uint256()));
+    
+    std::set<uint256> setBlockHash;
```
blondfrogs: This set doesn't seem to be used. Should be removed if it isn't being used

presstab: Good spot. Left over garbage. Fixed.

PR #490, opened 2019-04-10
[Build] Update TravisCI Descriptors

Fuzzbawls:
This overhauls the TravisCI descriptors with upstream changes and optimizations. Some notes:

* Running the Unit Tests is currently disabled due to pre-existing errors
* Running the Regtest Functional test suite is disabled due to it being incomplete with the veil code/network
* Linting the `src/secp256k1` and `src/univalue` subtrees are disabled due to divergences from upstream
* Linting the RPC mappings is disabled due to pre-existing failures
* Running the `lint-all.sh` script is disabled due to pre-existing failures
* Build targets that make use of `--disable-wallet` are disabled due to pre-existing compile-time errors with that configuration
* The macOS build target is being compiled without the `--enable-werror` configure flag due to pre-existing thread safety errors

These things can be re-enabled individually over time as their underlaying issues are resolved

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,21 @@
-sudo: required
-dist: trusty
+dist: xenial
 os: linux
 language: minimal
 cache:
   ccache: true
   directories:
-  - depends/built
-  - depends/sdk-sources
-  - $HOME/.ccache
+    - depends/built
+    - depends/sdk-sources
+    - $HOME/.ccache
 stages:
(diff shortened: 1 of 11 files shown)


Fuzzbawls:
@presstab had to incrementally build up the cache in travis to avoid timeouts, but it's now green across the board

presstab:
I see that now it is showing green.

PR #494, opened 2019-04-11
[Wallet][RPC] Add recoveraddresses rpc to regenerate keys and zerocoins after a seed import.

### LINKED ISSUE #453: Create functionality to restore addresses after fresh seed restore.
Generate a certain count of basecoin addresses.
Generate a certain count of stealth addresses.
Generate a certain count of zerocoins.
Scan blockchain for matches.
Delete any unused stealth addresses.

presstab:
`recoveraddresses` rpc can be used after importing a seed to the wallet. By default it will create 100 stealth addresses, 100 basecoin addresses, and 100 deterministic zerocoins (can be changed using `count`). It will then scan the full blockchain for any activity that is owned by the wallet. Since stealth addresses add computational overhead to block syncing, the rpc call will remove any of the unused stealth addresses. 

The process does keep a small buffer of up to 20 (`unused_buffer`) unused stealth addresses to accommodate the possibility that generated addresses were given out as a receiving addresses, but not used yet.

Users can also specify a block to start the search from in order to speed up the process (`scan_from_block`).

closes #453

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -387,10 +387,106 @@ bool AnonWallet::GetAddressMeta(const CStealthAddress& address, CKeyID& idAccoun
     if (mapKeyPaths.count(sxAddr.GetID())) {
         auto accountpair = mapKeyPaths.at(sxAddr.GetID());
         idAccount = accountpair.first;
+        std::string strPathToKey = BIP32PathToString(accountpair.second);
+
+        //The path of the account is what provides context
+        if (!mapKeyPaths.count(idAccount))
+            return false;
+        accountpair = mapKeyPaths.at(idAccount);
         strPath = BIP32PathToString(accountpair.second);
+        strPath += strPathToKey;
     }
 
     return true;
+}
+
+CStealthAddress GenerateStealthAddressFromIndex(const CExtKey& keyStealthAddress, int nIndex)
+{
+    //Generate the extkey for spend and scan
+    BIP32Path vPathNew;
+    vPathNew.emplace_back(nIndex, true);
+    CExtKey extKeyScan = DeriveKeyFromPath(keyStealthAddress, vPathNew);
+    vPathNew.clear();
+    vPathNew.emplace_back(nIndex+1, true);
+    CExtKey extKeySpend = DeriveKeyFromPath(keyStealthAddress, vPathNew);
+    CKey keyScan = extKeyScan.key;
+    CKey keySpend = extKeySpend.key;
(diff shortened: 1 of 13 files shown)


src/veil/ringct/anonwallet.cpp
```diff
@@ -3796,7 +3903,7 @@ bool AnonWallet::CreateAccountWithKey(const CExtKey& key)
     AnonWalletDB wdb(*walletDatabase, "r+");
     wdb.WriteAccountCounter(idAccount, (uint32_t)1);
 
-    auto idParent = mapKeyPaths.at(idAccount).first;
+    //auto idParent = mapKeyPaths.at(idAccount).first;
```
blondfrogs: Should this be being used? Could remove it if it isn't

src/veil/ringct/anonwallet.cpp
```diff
@@ -387,12 +387,106 @@ bool AnonWallet::GetAddressMeta(const CStealthAddress& address, CKeyID& idAccoun
...
+}
+
+void AnonWallet::ForgetUnusedStealthAddresses(int nBuffer)
+{
+    int nIndexLastUsed = GetLastUsedAddressIndex(idStealthAccount);
+    if (nIndexLastUsed < 0)
+        return;
+
+    // Get the last used address (this must be done after rescanning the chain)
+    int nLastGenerated = mapAccountCounter.at(idStealthAccount);
```
blondfrogs: Should check that mapAccountCounter contains the idStealthAccount first? Just as a precaution
`if (!mapAccountCounter.count(idStealthAccount) { return; }`

src/veil/ringct/anonwallet.cpp
```diff
@@ -387,12 +387,106 @@ bool AnonWallet::GetAddressMeta(const CStealthAddress& address, CKeyID& idAccoun
     if (mapKeyPaths.count(sxAddr.GetID())) {
         auto accountpair = mapKeyPaths.at(sxAddr.GetID());
         idAccount = accountpair.first;
+        std::string strPathToKey = BIP32PathToString(accountpair.second);
+
+        //The path of the account is what provides context
+        accountpair = mapKeyPaths.at(idAccount);
```
blondfrogs: How do we know that the mapKeyPaths contains the idAccount?. Should do a check to make sure the map contains it before accessing the map.

src/veil/ringct/anonwallet.cpp
presstab: `GetLastUsedAddressIndex()` returns negative if `idStealthAccount` is not in `mapAccountCounter`

PR #508, opened 2019-04-13
UI - textchange fix - correct error in first PR

MisterKangaroo:
fix issue #556

Redone since the first one was updating another file which was correct before fixing


MisterKangaroo:
closed the old PR... cause it was flawed

PR #513, opened 2019-04-13
[Wallet]Fix empty scriptpubkey causing issues creating CT transactions.

### LINKED ISSUE #486: Could not locate signing key
When minting zerocoin from RPC using CT funds:
`Error: Failed to create transaction: Failed to add ringctinputs: ringctwallet Could not locate signing key (code -4)`

Log:
```
2019-04-09T02:21:06Z ERROR: MakeSigningKeystore: Failed to extract destination
2019-04-09T02:21:06Z ERROR: AddBlindedInputs_Inner: ringctwallet Could not locate signing key
```

Other relevant information is that this wallet is encrypted, and unlocked at the time of transaction creation.

presstab:
If an output record has an empty scriptpukkey, then look it up manually instead of failing.

fixes #486

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -1702,6 +1702,19 @@ bool AnonWallet::SaveRecord(const uint256& txid, const CTransactionRecord& rtx)
     return true;
 }
 
+bool GetScriptPubKeyFromOutpoint(const COutPoint& outpoint, CScript& scriptPubKey)
+{
+    //output record does not have the signing key, find it manually
+    CTransactionRef ptx;
+    uint256 hashBlock;
+    if (!GetTransaction(outpoint.hash, ptx, Params().GetConsensus(), hashBlock, /*allowslow*/true))
+        return error("Failed to find input transaction %s", outpoint.hash.GetHex());
+    if (ptx->vpout.size() <= outpoint.n)
+        return error("Failed to find input output %s:%d does not exist", outpoint.hash.GetHex(), outpoint.n);
+
+    return ptx->vpout[outpoint.n]->GetScriptPubKey(scriptPubKey);
+}
+
 int AnonWallet::AddStandardInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std::vector<CTempRecipient> &vecSend,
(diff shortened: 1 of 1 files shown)


src/veil/ringct/anonwallet.cpp
```diff
@@ -2292,7 +2292,7 @@ int AnonWallet::AddBlindedInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx,
 
         std::vector<std::pair<MapRecords_t::const_iterator, unsigned int> > setCoins;
         std::vector<COutputR> vAvailableCoins;
-        AvailableBlindedCoins(vAvailableCoins, true, coinControl);
+        AvailableBlindedCoins(vAvailableCoins, true, coinControl, /*nMinimumAmount*/2);
```
blondfrogs: is the nMinimumAmount needed here? Should it get replaced with the 2?

src/veil/ringct/anonwallet.cpp
```diff
@@ -2653,7 +2664,17 @@ int AnonWallet::AddBlindedInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx,
...
+                if (scriptPubKey.empty()) {
+                    //output record does not have the signing key, find it manually
+                    CTransactionRef ptx;
+                    uint256 hashBlock;
+                    if (!GetTransaction(txhash, ptx, Params().GetConsensus(), hashBlock, /*allowslow*/true))
+                        return wserrorN(1, sError, __func__, strprintf("Failed to find input transaction %s", txhash.GetHex()));
+                    if (ptx->vpout.size() <= outputRecord->n)
+                        return wserrorN(1, sError, __func__, strprintf("Failed to find input output %s:%d does not exist", txhash.GetHex(), outputRecord->n));
+                    scriptPubKey = *ptx->vpout[outputRecord->n]->GetPScriptPubKey();
+                }
```
blondfrogs: Some copy pasta on this. Could be put into a function?

src/veil/ringct/anonwallet.cpp
presstab: I set it to two because I noticed some 1 sat outputs were being added when I was seeing this error pop up. I have a suspicion that there are some 1 sat outputs floating around that aren't actually 1 sat, and screw up the entire tx.

PR #519, opened 2019-04-16
[GUI] Add StakeCounter in status bar.

### LINKED ISSUE #432: Add a visual inducator showing how many stakes have occured in a user selectable time frame.
Add a visual counter displaying how many stakes occurred in a user selectable time period, from 1 hour to 7 days. 

A simple visual number should suffice somehwere on the GUI.

drag0ng:
Closes #432


Fuzzbawls:
Qt isn't compiled with the QtConcurrent feature set in `depends`, and as such any build that relies on `depends` (TravisCI non-syslib builds, and gitian release builds) will fail with this PR as-is.

If this can be done without the use of QtConcurrent, that would be ideal; otherwise changes to the `depends` system will be necessary, which may break determinism (this would need to be checked and rechecked, and would probably warrant it's own PR).

mimirmim:
Closing due to out of date and major upcoming UI changes.

PR #530, opened 2019-04-22
[Build] Fix windows build with boost thread

blondfrogs:
- Statically compile windows builds on linux was failing. This change fixes the windows build problem


MisterKangaroo:
uTack

Issue #462, opened 2019-03-30
Add syncing indicators to lower bar in GUI


4x13:
Add a visual indicator next to the lock icon that shows the wallet is syncing and the block number and an indicator that it is synced to the tip of the chain.

presstab:
By visual indicator, do you mean progress bar? This is something that is in the new GUI mockup, so best to leave this issue as untouched until the mockup is finalized and the build up of the new GUI begins.

presstab:
Bounty: 500
Estimated Hours: 6

veilgets:
Developer: MatWaller

MatWaller:
This one is pretty straight forward as we already have access to the progress percentage.

I'll see what I can come with up.

Thanks.

MatWaller:
closed with #543

codeofalltrades:
closed with #543

Issue #221, opened 2019-01-07
Is there any way to inform the user that the synchronization process is not stuck?


gabrielnergaard:
Running 1.0.0.6 on Mac OS X 10.14.2, my screen appeared to be in a stuck state for about five minutes:

![](https://clipz.in/6pmn7.png)

Just when I was about to quit the application, it started processing blocks.

During those five minutes in which the app obviously was not stuck, but the user would from the UI believe it to be so, is there any status information we can show to communicate that it's not stuck?

presstab:
How do you know it was not stuck during those 5 minutes? I am occasionally seeing periods of repetitive loops in syncing where nothing really progresses.

gabrielnergaard:
I do not know that it was not stuck during those 5 minutes. In fact, I thought it was stuck and was about to quit the app when it suddenly progressed. When it proceeded, I then presumed that it wasn’t stuck.
Jan 7, 2019, 10:18 PM by notifications@github.com:

seanPhill:
That's why I rarely leave it at that screen. I hide it and look at the Information tab of the "Advanced Options" window.

StrontiumZ38:
Is this something that can wait until new wallet? I don't undervalue this from a UX perspective whatsoever, but it may be resources better spent elsewhere for now.

presstab:
Bounty: 187.5

veilgets:
Developer: MatWaller

MatWaller:
Closed with #543

PR #543, opened 2019-04-25
[GUI] Sync progress ui additions

MatWaller:
Add a Block height indicator `syncIndicatorBtn` to `veilStatusBar` to display current block height.

This should help with the issues #462 & #221 

Update

Receiving Address
```
sv1qqpdsfcqvfcljtcq8txc8kc2dw4e9upj69wnnfv7j0ufcsrv7trs9lqpqfscke6jj92aj4aup387ggk7mathznr8hz3zde6ut87jdwphrns7qqqq00jn8w
```

Thanks,

M

### DIFF
--- a/src/qt/veil/forms/veilstatusbar.ui
+++ b/src/qt/veil/forms/veilstatusbar.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>440</width>
+    <width>500</width>
     <height>40</height>
    </rect>
   </property>
@@ -91,7 +91,7 @@
        </widget>
       </item>
       <item>
-       <layout class="QHBoxLayout" name="horizontalLayout" stretch="0,0,0,0">
+       <layout class="QHBoxLayout" name="horizontalLayout" stretch="0,0,0,0,0">
         <property name="leftMargin">
          <number>9</number>
         </property>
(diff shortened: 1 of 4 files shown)


MatWaller:
@presstab I have now cleaned up the commits.

presstab:
@MatWaller Tested this. A very nice feature to have. Plugs right in to existing logic, so very simple solution without a lot to worry about. Nice work!

In terms of visuals... I think this would fit in better if this were a `QLabel` instead of a button. And also if the label were `Height: xxxxxx`.

MatWaller:
No worries, I will get that updated now @presstab

MatWaller:
@presstab 

Hey, all done.

Update

src/qt/veil/veilstatusbar.cpp
```diff
@@ -40,6 +40,11 @@ void VeilStatusBar::updateSyncStatus(QString status){
     ui->btnSync->setText(status);
 }
 
+void VeilStatusBar::updateSyncIndicator(int height){
+    QString str;
+    ui->btnSyncIndicator->setText("Height: " + str.setNum(height));
```
blondfrogs: Text that the user can see should be wrapped with a translation method. We need to be proactive to any text changes made to the wallet. Example -> `tr("Height: ")`

MatWaller: Hey @blondfrogs thanks, done & committed.

src/qt/veil/forms/veilstatusbar.ui
```diff
@@ -186,7 +196,7 @@
            <string/>
           </property>
           <property name="icon">
-           <iconset>
+           <iconset resource="../../veil.qrc">
```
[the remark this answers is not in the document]
MatWaller: Removed & Squashed.

PR #544, opened 2019-04-27
[Consensus][Wallet]Put Zerocoin into "Limp Mode"

presstab:
## Overview
Due to disclosures made by [ZCoin](https://zcoin.io/further-disclosure-on-zerocoin-vulnerability/), and lots of analysis done with other teams such as ZCoin and PIVX, Veil will be placing its zerocoin protocol into a functional but **not anonymous** state. The transition will occur rapidly and be done using bit signalling in the block header.

Every zerocoin spend/stake will have no longer have any privacy associated with it and should be considered as transparent as a basecoin transaction. More details will follow soon that will provide a complete detailed breakdown and analysis of why these changes are being made.

### Important Notes:
- This puts zerocoin into what we call *limp mode*. 
- The privacy of both **CT and RingCT are not impacted by this.**

### Normal Zerocoin Operating
Every zerocoin mint has an identifier that gets added to the cryptographic accumulator. During normal operation of zerocoin, when someone goes to spend their mint they create a zero knowledge proof that their mint's identifier has been added to the accumulator. The proof does not reveal the mint, and the spent mint is indistinguishable from all other mints of that same denomination. This is the mechanism that makes the zerocoin protocol private and removes the link between mint and spend.

### Limp Mode
Limp Mode is being used by Veil to allow the payment protocol of zerocoin to function, while disabling the privacy aspect of zerocoin. Each zerocoin spend will also include its mint identifier so that it can be linked by Veil's blockchain database to ensure that it is a valid zerocoin that has been accumulated. This effectively removes any privacy aspect of the zerocoin protocol.

Limp Mode is useful in times that call for zerocoin to be disabled because of concerns over the ability to spend coins that have not been accumulated. For Veil this is especially critical because a large majority of the coin supply is held in zerocoins, and the proof of stake consensus system relies on zerocoin. Veil's limp mode allows staking to continue and zerocoin transactions to continue to take place.

This is should be seen as a temporary protocol that allows Veil to have more time to work on creating the best solutions to some complex problems that may not be solved in such a short time span, and especially not solved with the level of quality we would like it to be.

### Thanks To:
- The team at ZCoin, specifically [Peter](https://github.com/psolstice) and [Reuben](https://github.com/zcoinofficial/zcoin/commits?author=reubenyap) for promptly notifying members of the Veil team as well as a few other trusted developers, as well as doing the tedious work of identifying the errors as well as figuring out how to reproduce the errors.

- [random-zebra](https://github.com/random-zebra/), a contributor to both Veil and PIVX, who spent countless hours over the last week analyzing the complex and sensitive cryptography in `libzerocoin` and coming up with a solution that lets users safely send and receive zerocoin transactions without privacy aspects enabled.

- [blondfrogs](https://github.com/blondfrogs), a core developer for Veil, who helped test and implement the blockchain's transition to Limp Mode.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ project(veil)
 set(CMAKE_CXX_STANDARD 11)
 
 add_compile_definitions(ENABLE_WALLET=1)
+add_compile_definitions(USE_NUM_GMP)
 
 include_directories(src)
 include_directories(src/bench)
@@ -63,6 +64,8 @@ include_directories(src/zmq)
 find_package(Qt5 REQUIRED COMPONENTS Widgets Core Gui)
 
 add_executable(veil
+        src/libzerocoin/PubcoinSignature.h
+        src/libzerocoin/PubcoinSignature.cpp
         src/veil/zerocoin/witness.h
         src/veil/zerocoin/witness.cpp
         src/bench/base58.cpp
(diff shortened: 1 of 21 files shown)


blondfrogs:
ACK

PR #569, opened 2019-05-03
[Wallet] Add ability to autospend zerocoin denominations

### LINKED ISSUE #559: Spending zerocoins en masse
This Feature Request is for people who may have found themselves with more than ten thousand ten denom zerocoins that they cannot spend, without eating into processing power so much that spending x thousand zerocoins prevents staking for more than it is worth in moving the zerocoins to larger denominations.

If we can figure out a better way of spending ten thousand tens, ie. 100K Veil of ten denoms, it would be helpful for some people.

I have two testnet wallets that could test any solution being trialled.

One wallet has ONLY tens and a value of 157K (testnet) Veil. It has become so slow that I have had to turn OFF staking, as it forks from the chain while it calculates stakes or when trying to spend. _That is my **small** wallet!_

My other wallet is also accumulating thousands of tens, because having several hundred thousand (testnet) Veil in 10K, 1K and some 100s, it stops generating blocks as it tries to spend, and other nodes fork from it. So I've given up trying to spend the tens for now.
This wallet was created while trying to spend out the zeorocoins from the other wallet, which had bloated to hundreds of MB, but having moved the large denoms to this wallet, I have _never been able to complete the process_ of moving all the zerocoins from wallet to wallet.

I am currently only using the released 1.0.3 wallet, but this problem has been building for some time.

I also have precomputing turned off, because these testnet wallets appear not to be able to cope with precomputing on.

blondfrogs:
- Fixes regtest params to work with stealth addresses
- Added rpc call **startautospend** with three optional paramters (address, count, type)
    1. Address - The stealth address to auto spend to (if not given one will be generated)
    2. Number of spends: the number of spends to make per cycle ( 1 - 3)
    3. The denomination to spend ( 10, 100, 1000, 10000)
- Added rpc call **stopautospend** which stops the autospend thread
- Added rpc call **getautospendaddress** which shows the current address set to auto spend to
- Added new arugments
  1. **-autospend** enables the autospend thread
  2.  **-autospendcount=number** the number of spends to make per cycle (1-3)
  3. **-autospenddenom=number** the denomination to spend ( 10, 100, 1000, 10000)
  4. **-autospendaddress=string** the destination address for the zerocoin spends

Edit:
Spends zerocoin with security of 1

Closes #559

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -7019,6 +7019,14 @@ void AutoSpendZeroCoin()
                 continue;
             }
 
+            TRY_LOCK(cs_main, fLockedMain);
+            if (!fLockedMain)
+                continue;
+
+            TRY_LOCK(pwallet->cs_wallet, fLockedWallet);
+            if (!fLockedWallet)
+                continue;
+
             bool fSuccess = pwallet->SpendZerocoin(found * libzerocoin::ZerocoinDenominationToInt(denom_to_spend) * COIN, 1, receipt, vMintsSelected, false, false, denom_to_spend, &destination);
 
             if (fSuccess) {
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -4039,7 +4039,7 @@ DBErrors CWallet::ZapWalletTx(std::vector<CWalletTx>& vWtx)
     return DBErrors::LOAD_OK;
 }
 
-bool CWallet::SetAddressBook(const CTxDestination& address, const std::string& strName, const std::string& strPurpose, bool bench32){
+bool CWallet::SetAddressBook(const CTxDestination& address, const std::string& strName, const std::string& strPurpose, bool bench32) {
     bool fUpdated = false;
     {
         LOCK(cs_wallet); // mapAddressBook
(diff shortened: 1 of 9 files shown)


4x13:
Do we anticipate having many issues with spend using higher security since we are in limp mode presently?

blondfrogs:
Updated PR to spend zerocoin with security level of 1 which is okay because of limp mode

src/wallet/wallet.cpp
```diff
@@ -6826,3 +6826,159 @@ void CWallet::PrecomputeSpends()
...
+
+bool StartAutoSpend()
+{
+    if (!pthreadGroupAutoSpend) {
+        error("%s: pthreadGroupAutoSpend is null! Cannot autospend.", __func__);
+        return false;
+    }
+
+    // Close any active precomputing threads before starting new threads
+    if (pthreadGroupAutoSpend->size() > 0) {
```
presstab: The comments here don't seem to match what this is doing (not precomputing based).

src/wallet/wallet.cpp
```diff
@@ -6826,3 +6826,159 @@ void CWallet::PrecomputeSpends()
...
+                LogPrintf("%s: pwallet is null cannot auto spend\n", __func__);
+                continue;
+            }
+
+            if (pwallet->IsLocked() || pwallet->IsUnlockedForStakingOnly()) {
+                LogPrintf("%s: pwallet is locked cannot auto spend\n", __func__);
+                continue;
+            }
+
+            if (destination.empty()) {
```
presstab: What is the reason for adding a key?

blondfrogs: When spending zerocoin, we want to make sure that we spend it to an address that the wallet owns. This only happens on the first loop of the auto spend, and uses the same address for as long as the auto spend thread runs.

src/wallet/wallet.cpp
blondfrogs: Some copy paste errors. Will update comments with correct wording

src/wallet/wallet.cpp
presstab: We should only spend to stealth addresses (not basecoin), and should only create an address to spend to once. Creating a stealth address and adding it to the address book should do the trick.

blondfrogs: Updated to spend to stealth addresses. Also, instead of trying to create a new address every time the wallet loads, the wallet creates and stores a stealth address to be used for auto spending. the user can change this address by using the **startautospend** rpc call

blondfrogs:
@presstab Ready for code review

src/wallet/wallet.cpp
```diff
@@ -6844,3 +6857,181 @@ void CWallet::PrecomputeSpends()
...
+            }
+
+            std::vector<CZerocoinMint> vMintsSelected;
+            CZerocoinSpendReceipt receipt;
+
+            // Get the list of mints
+            std::set<CMintMeta> mints = pwallet->ListMints(true, true, false);
+
+            int found = 0;
+            for (auto mint : mints) {
```
presstab: Lets use `const auto& mint : mints` here.

ncrypter90:
QA Passed.

Note:As the release note says -Address - The stealth address to auto spend to (if not given one will be generated), but rpc call will not create new address instead it uses existing address in the wallet.

Veil_build_569_Screengrab

PR #580, opened 2019-05-10
Remove bit 2 warning.

### LINKED ISSUE #563: Get rid of bit 2 warning.
Ever since putting zerocoin into limp mode, the wallet has had a warning about bit 2.

presstab:
closes #563

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -198,7 +198,7 @@ public:
 
         consensus.vDeployments[Consensus::DEPLOYMENT_POS_WEIGHT].bit = 2;
         consensus.vDeployments[Consensus::DEPLOYMENT_POS_WEIGHT].nStartTime = 1548161817;
-        consensus.vDeployments[Consensus::DEPLOYMENT_POS_WEIGHT].nTimeout = 1556226440;
+        consensus.vDeployments[Consensus::DEPLOYMENT_POS_WEIGHT].nTimeout = 1576226440;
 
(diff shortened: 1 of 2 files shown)


presstab:
@blondfrogs forced pushed over what you reviewed. Added iteration over the consensus deployments to manually check if the bit has been defined. It doesn't seem like it serves a good purpose to print a warning of something that your wallet has knowledge of. On the other hand if your wallet does not have knowledge of the bit being signalled, you should be warned.

blondfrogs:
Makes sense. I like this change

codablock:
FYI, the real reason for the warning to have appeared is very likely that you have changed `nRuleChangeActivationThreshold` and `nMinerConfirmationWindow` without taking the already deployed BIP9 deployments into account. Look at https://github.com/Veil-Project/veil/commit/70de070acc7d31784cdfa0ab5c92089ccb44c0ee. This causes things to get messed up now.

The better solution for this would be to maintain these parameters per deployment so that you can vary it when needed. Look at the changes Dash did for DIP1 as an example: https://github.com/dashpay/dash/commit/cd262bf641cc560573b9cac728bf9bb606af091a#diff-bc9415451d733499187a3fe59e0fd7e6R34 (the addition of nWindowSize and nThreshold in BIP9Deployment)

presstab:
@codablock awesome reference thanks for linking it. That would be extremely useful, because (as was the case here) sometimes it is necessary to have faster than usual deployments. Will add this to the list of items to get done before we do another deployment.

PR #583, opened 2019-05-11
[GUI] Extending clickable area of Staking & Precompute toggle switch.

### LINKED ISSUE #503: GUI Bug Fix: Toggle Switches Do Not Always Respond To Clicks
It looks like the GUI Toggle buttons do not have the expected "hit box" or "click box" that would be expected of them as displayed on the screen. 

In my experience, clicking on the text associated with the toggle has always worked for me, but clicking on the toggle directly in some cases does nothing.

Because this is a cross-platform display GUI issue and I imagine it isn't a simple fix, the fix for this should be built and tested across multiple wallet builds before submitting the fix if possible.

image

Clicking the areas in red has no effect despite it being the first or second choice as user action for making the toggle to change state. Alternatively, drag and slide would be the other first user expected experience which also doesn't work.

MatWaller:
Closes #503 

- Indicator of `checkStaking` & `checkPrecompute` have been extended to cover toggle switch.
- `checkStaking` is now styled within `QtCreator`
- Updated `main.css` to include styling for `checkStaking` & `checkPrecompute` 

Bounty Address

```
sv1qqpdsfcqvfcljtcq8txc8kc2dw4e9upj69wnnfv7j0ufcsrv7trs9lqpqfscke6jj92aj4aup387ggk7mathznr8hz3zde6ut87jdwphrns7qqqq00jn8w
```

### DIFF
--- a/src/qt/res/css/main.css
+++ b/src/qt/res/css/main.css
@@ -453,7 +453,6 @@ QCheckBox[cssClass~="switch"] {
 background-image: url(":/icons/ic-switch-off") ;
 background-repeat:no-repeat;
 background-position:left center;
-padding:4px 4px 8px 30px;
 }
 
 QCheckBox[cssClass~="switch"]:checked {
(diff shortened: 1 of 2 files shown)


xo-orca:
The clickable extension is functioning as expected on all three platforms (Win, Linux and Mac).
xo

Issue #495, opened 2019-04-11
Require a destination address in the rpc call spendzerocoin


blondfrogs:
Link from issue #446

veilgets:
197.22

veilgets:
Developer: PererL73

PeterL73:
Ready for QA

PR #588, opened 2019-05-15
[RPC] SpendZeroCoin Address Validation

MatWaller:
Adding Address validation to ensure `addressTo` is both provided and a valid stealth address.

aaaa

#495


ncrypter90:
QA Passed.
Note:
Error message while sending it to incorrect address is not informative.
Error message:Transaction created (code -4)

PR #597, opened 2019-05-17
[Wallet][RPC]Fix missing RingCt Blinds. Add RingCt inputs print to getrawtransaction.

presstab:
In some situations spending RingCt would fail because of missing blinds.

`gettransaction` and `getrawtransaction` will now print the list of RingCt inputs used in a RingCt spend.

### DIFF
--- a/src/core_io.h
+++ b/src/core_io.h
@@ -11,6 +11,7 @@
 #include <vector>
 
 class CBlock;
+class COutPoint;
 class CScript;
 class CTransaction;
 struct CMutableTransaction;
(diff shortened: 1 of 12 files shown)


crypt0-fusi0n:
@presstab gettransaction is now listing ringct inputs but getrawtransaction output and hex value from gettransaction is still the same. Is this expected?

[qa-after.txt](https://github.com/Veil-Project/veil/files/3219838/qa-after.txt)
[qa-before.txt](https://github.com/Veil-Project/veil/files/3219839/qa-before.txt)

presstab:
@crypt0-fusi0n you need to use `1` after `getrawtransaction` to tell it to decode the hex.

crypt0-fusi0n:
great I see it now... confirmed working as expected.

PR #601, opened 2019-05-21
[WIP][Consensus][Zerocoin][Wallet] Lighter Public Zerocoin Spend Protocol

### LINKED ISSUE #565: Make zerocoin limp mode even lighter.
If all anonymity is lost, we should only need to have a signature from the privkey of the serial (which is a pubkeyhash). All other proofs should be able to be discarded.

Requires:
New BIP
Transaction Serialization

presstab:
closes #565


4x13:
How is this coming along??

PR #608, opened 2019-06-01
Added Veil Copyright MIT notices to a lot of files

mrmetech:


presstab:
Thanks for going through and doing this!

PR #607, opened 2019-05-30
GUI - Update VEIL logos to the actual logo version

MisterKangaroo:

PR #609, opened 2019-06-01
Update Veil logos to the actual logo version

MisterKangaroo:


xo-orca:
> Code looks fine
> 
> QA will need to make sure that all logos look perfect



> Code looks fine
> 
> QA will need to make sure that all logos look perfect

blondfrogs, PR609 and PR607 both imply that the logo is updated to the "actual state" or the "actual version" what is that means and what am I expecting differently?

MisterKangaroo:
@xo-orca
#609 replaces #607, because 607 included some merging that needed to be removed accordingly to presstab.

The logos were slightly modified (cleanup and proportion between logotype and icon) since the wallet was initially created. That PR changes the old logo version to the new one.

PR #611, opened 2019-06-04
[Qt][Wallet]Make send dialog more asynchronous.

presstab:
- Run send coins on a different thread and emit a signal when it is done.
- Make `cs_wallet` lock more fine grain on `SpendZerocoin()` so that it does halt all other items that use `cs_wallet`. This allows the UI to continue to be used while a transaction is being created.
- Add a status label that displays (and animates) the current state of transaction creation.
- Disable spend button if a transaction is being generated.
- Use a small security level for zerocoinspends since zerocoin is not anon right now anyways.
- Refactor some bad GUI code that pointlessly calls wallet methods that wait on `cs_wallet` lock.

Dev Note: This adds QtConcurrent to the depends.

Use Example:
ui_send1

Tabs Being Switch While Tx Is Being Generated:
ui_send2

Button disabled when send is pressed:
ui_send3

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -65,7 +65,7 @@ SendCoinsDialog::SendCoinsDialog(const PlatformStyle *_platformStyle, QWidget *p
     platformStyle(_platformStyle)
 {
     ui->setupUi(this);
-    prepareData = nullptr;
+    m_prepareData = nullptr;
     m_statusAnimationState = 0;
     m_timerStatus = new QTimer(this);
     connect(m_timerStatus, SIGNAL(timeout()), this, SLOT(StatusTimerTimeout()));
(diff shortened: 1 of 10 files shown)


src/qt/sendcoinsdialog.cpp
```diff
@@ -436,14 +442,76 @@ void SendCoinsDialog::on_sendButton_clicked()
...
+    fNewRecipientAllowed = false;
+    WalletModel::UnlockContext ctx(model->requestUnlock());
+    if(!ctx.isValid())
+    {
+        // Unlock wallet was cancelled
+        fNewRecipientAllowed = true;
+        return;
+    }
+
+    prepareData = new PrepareTxData();
```
presstab: This is probably a memleak here. Should change `prepareData` to be a smart pointer for automatic deletion.

blondfrogs: Agreed. Every time the send button is clicked new memory is being allocated for the pointer.

ncrypter90:
QA Failed...

Win 64: Transactions works fine as expected if send amount less or equal (including fee) to balance. 
transaction processing time is way less so I couldn't switch tabs see whether buttons disabled.

However, issues is coin transaction go through even after pressing cancel button on the dialog.

Veil_build_611_Screengrab

ncrypter90:
As discussed @blondfrogs, we can mark it Passed.

PR #614, opened 2019-06-16
Reject < v4 zerocoin from mempool.

presstab:
Make sure that zerocoinspends that are less than version 4 do not make it into the mempool.


presstab:
Travis fail is from bad build cache. Merging.

PR #621, opened 2019-06-20
[Testnet] Create new testnet chain

### LINKED ISSUE #619: Restart Testnet
- Add large amount of veil mined in the first block
- Change peer version

blondfrogs:
- Added a large amount of veil into the first block mined for regtest, and testnet chains
- Increased the Protocol Version so that only new testnet nodes connect to each other
- Add ability to turn off automint with a console parameter
- Turn on LC_LIMP and POS_WEIGHT BIPs by default of Testnet

Closes #619

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2686,9 +2686,13 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                 FormatMoney(networkReward), FormatMoney(nFounderPayment), FormatMoney(nBudgetPayment),
                 FormatMoney(nLabPayment));
 
-        return state.DoS(100, error("ConnectBlock(): coinbase pays too much (actual=%s vs limit=%s)\n %s",
-                                    FormatMoney(nCreated), FormatMoney(nCreationLimit), block.vtx[0]->ToString()),
-                         REJECT_INVALID, "bad-cb-amount");
+        // On testnet and regtest, we allow the first block to mine lots of coins to make testing easier
+        // We need to make sure this isn't allow on Mainnet.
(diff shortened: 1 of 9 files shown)


src/validation.cpp
```diff
@@ -2686,9 +2686,13 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                 FormatMoney(networkReward), FormatMoney(nFounderPayment), FormatMoney(nBudgetPayment),
                 FormatMoney(nLabPayment));
 
-        return state.DoS(100, error("ConnectBlock(): coinbase pays too much (actual=%s vs limit=%s)\n %s",
-                                    FormatMoney(nCreated), FormatMoney(nCreationLimit), block.vtx[0]->ToString()),
-                         REJECT_INVALID, "bad-cb-amount");
+        // On testnet and regtest, we allow the first block to mine lots of coins to make testing easier
+        // We need to make sure this isn't allow on Mainnet.
+        if (Params().NetworkIDString() == CBaseChainParams::MAIN || pindex->nHeight != 1) {
```
presstab: Why not just change the reward schedule for testnet instead of this roundabout way?

blondfrogs: Great Idea

blondfrogs:
@Fuzzbawls Any reason you can see why this is failing the build checks

seanPhill:
In wallet.cpp line 4787 please change:
`Nothing minted because automint if turned off.` to
`Nothing minted because automint is turned off.`
It gets printed in the debug log periodically (and quite often).

blondfrogs:
There have been lots of new additions to the code base. We should launch a newer testnet than this one. What do you think @presstab

blondfrogs:
Closing due to being outdated. Replaced by #673

PR #627, opened 2019-06-21
[Net] Add P2P version tracker & update notifier

### LINKED ISSUE #404: Feature Request — Warn user of possible wallet update needed
Knowing about a wallet update is always convenient to users, and in some cases of critical importance.

To avoid external dependencies, like checking GitHub, we could take the approach used by Dash. Whenever the wallet detects that peers on the network are using a later version, an impossible-not-to-see yellow bar is present on the wallet overview screen stating something like:

"We've noticed a majority of peers on the network running a later version of the wallet than you. You may want to check to see if there's an update available. For reference you are running 0.13.0 and the others are running 0.13.1"

They don't link to GitHub, just simply provide that advice to the user.

Can we do this in the Veil wallet?

JSKitty:
This PR closes #404 by adding a check onto every new connection that measures and saves the connection's peer subversion, this can then be used to measure the average network version, and figure out if the client is "behind", "forward" or "current" in version integers, if the client is behind, the check will trigger a messagein warnings.cpp which displays an warning of `"Over 50% of our peers are running a newer wallet version than yours. An update may be available."`

The update tracker is relatively simple, but the functions of it must be known clearly, so that it doesn't get abused by malicious actors, and doesn't false-positive on specific version integers.

- Atleast 50% of historical peers, must be a higher version before an update warning is triggered.
- Atleast 2 historical peers minimum must be a higher version before an update is considered.

This means at minimum, if you have 4 connections, two are older or matching wallets, two are new wallets higher than our own, the update warning **WILL** fire.

If you connect to only 1 peer, and they are a higher version, the update warning will **NOT** fire, due to the minimum requirement of 2 higher-ver peers, to help prevent false-positives.

It is theoretically possible to do an "attack" if an actor is to compile a wallet with a fake, higher version than the official wallet. But, due to seednodes and the natural size of the network, it would take a very high amount of 'fake version' nodes before an update notif. would fire.

No arbitrary data can be inserted into the update tracker, it will not display a URL or version text when an update triggers, it will just purely notify that an update may be possible, due to >50% of peers having higher subversions.

-------------
A possible bug with the current system, is that it cannot properly parse updates with a larger integer than 10. I'll copy-paste a comment I wrote in net_processing.cpp
```
        Due to the nature of this splicing mechanic, a version integer over 10
        will cause the updater to get 'confused'.
        
        For example, v1.2.3 would equate to 123 when spliced, but v1.10.3 would
        equate to 1103, which, even though in versioning terms is an OLDER
        version, the updater cannot see that in integer form.
        
        This problem is avoided if you do less than 10 major / minor / patch
        updates in a row, but if that cannot be done, I'd recommend we find
        a way to splice without breaking version tracking over 10 updates.
```

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -10,8 +10,10 @@
 #include <blockencodings.h>
 #include <chainparams.h>
 #include <checkpoints.h>
+#include <clientversion.h>
 #include <consensus/validation.h>
 #include <hash.h>
+#include <warnings.h>
 #include <validation.h>
 #include <merkleblock.h>
 #include <netmessagemaker.h>
@@ -80,6 +82,33 @@ std::map<uint256, COrphanTx> mapOrphanTransactions GUARDED_BY(g_cs_orphans);
 std::map<int, CBlock> mapStagedBlocks;
 static constexpr int ASK_FOR_BLOCKS = 50; //How many blocks to ask for at once
 static CCriticalSection cs_staging;
+
+bool fUpdateCheck = true;
+
+bool shouldUpgrade = false;
+
+// --------------------------------------------------
+// Upgrade Majority
+// This is the magic number for deciding when enough
+// peers have a higher version to be considered the
+// 'majority' of the network for updating the software.
+
+// Higher means more secure and less false-positives,
+// but lengthens time before the update is triggered.
+
+// Peers / nUpgradeMajority = Minimum peers for update.
+// An additional protection against false-positives
+// is the Updater will not fire unless there's atleast
+// two higher peers, minimum, even if the percentage
+// majority (With 4 or less peers) is high enough.
+
+int nUpgradeMajority = 2; // 50% of peers is 'majority'.
+// --------------------------------------------------
+
+// Track all peer version differences
+int higherVerPeers = 0;
+int currentVerPeers = 0;
+int lowerVerPeers = 0;
 
 void EraseOrphansFor(NodeId peer);
 
@@ -2065,9 +2094,12 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
 
         // Potentially mark this peer as a preferred download peer.
         {
-        LOCK(cs_main);
-        UpdatePreferredDownload(pfrom, State(pfrom->GetId()));
+            LOCK(cs_main);
+            UpdatePreferredDownload(pfrom, State(pfrom->GetId()));
         }
+        
+        // Check the peer's version and compare it to us
+        CheckForUpdates(pfrom->addr.ToString(), pfrom->strSubVer);
 
         if (!pfrom->fInbound)
         {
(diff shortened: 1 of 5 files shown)


alexeyneu:
i think if you worried about attack of this kind you can just do ssl connection to github and check last release number.  
GET on   
`https://github.com/Veil-Project/veil/releases/latest`  
it's np to code

JSKitty:
> i think if you worried about attack of this kind you can just do ssl connection to github and check last release number.
> GET on
> `https://github.com/Veil-Project/veil/releases/latest`
> it's np to code

That would then turn it into a centralized attack vector~ If the github is compromised, the update system is compromised, as well as the binary files (So the attacker can BOTH notify an update, and slip in malicious files at one time). And that's a catastrophe situation compared to the decentralized attack / defense methods written above. ^
Also, if Github goes down / maintenance / whatever, no result would return, and the update system would be stuck for a while. Having heavy reliance on Github to stay secure + online 24/7.

alexeyneu:
if you lost github you'll need to make a fork , bchsv story. warning will be last thing to care then .  
  
if github is offline you'll not receive 302 response . It's all np
```
curl https://github.com/Veil-Project/veil/releases/latest --verbose
...
< Status: 302 Found
< Vary: X-PJAX
< Location: https://github.com/Veil-Project/veil/releases/tag/v1.0.3.1
...


```

JSKitty:
@alexeyneu 
/shrug, I personally still disagree with the idea, every bit of centralization should be eradicated, adding more won't help, if anything, we should focus on strengthening the decentralized defenses instead.

Afterall, this PR was written based on the vision by @gabrielnergaard and @jakimanboy in issue #404, which also explicitly stated `"I support this idea as it doesn't create any extra polling over the network i.e. to our GitHub, and merely compares the connected peer's wallet version to see if there is a discrepancy."` So I'm just completing work requested by other team members~

xo-orca:
Hi All
This PR has been tested on all OSs and behaved as expected.

Regarding the potaintial bug that JSKitty has highlighted here in the comment and in the code is valid. 
I can confirm it from the code and will bite us later if version major exceed the limit of 10.

If everyone is happy and agree with this limitaion I can pass this PR.

However, if you are interested in a work around, maybe you can split the 3 sections "major / minor / patch" and format each number to 2 digits 00 before concatinate them back and convert it to intger. this will increase the limit to  99 which is plenty

Cheers,
orca.

xo-orca:
If I undertsnad this limitation correctly and you are interested in a work around, maybe you can split the 3 sections "major / minor / patch" and format each number to 2 digits 00 before concatenate them back and convert it to integer. this will increase the limit to 99, which is plenty.

src/net_processing.cpp
```diff
@@ -4311,6 +4343,70 @@ bool PeerLogicValidation::SendMessages(CNode* pto)
...
+  {
+    /*
+        - JSKitty
+       
+        Due to the nature of this splicing mechanic, a version integer over 10
+        will cause the updater to get 'confused'.
+        
+        For example, v1.2.3 would equate to 123 when spliced, but v1.10.3 would
+        equate to 1103, which, even though in versioning terms is an OLDER
+        version, the updater cannot see that in integer form.
```
presstab: Lets change this to parse major/minor/patch as different numbers so that they can be properly compared to each other.

presstab:
This should not hit Github's api. Github's api sucks anyways.

src/net_processing.cpp
JSKitty: Noted, will work on this soon. Need to study C++ a bit more to write a solution which isn't too messy.

JSKitty:
Closing to rebase changes on the latest Veil master and create a new PR, new PR will also fix the major/minor/patch parsing bug.

PR #630, opened 2019-06-26
[Wallet] Use one default stealth change address for CT/RingCT transactions.

presstab:
Currently the wallet generates a new stealth address everytime it needs to issue change in a transaction that has CT or RingCT outputs. Creating a new stealth address each time is bulky and increases the amount of scanning that needs to happen when checking the blockchain for transactions that belong to the wallet.

Since a new stealth destination is derived from a stealth address each time it is used, it provides the same utility if the wallet uses a dedicated change address.

This adds a dedicated change address that is in a specific location on the seed's key path.

This also adds `getstealthchangeaddress` RPC which will print out the default change address.

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -130,10 +130,10 @@ bool AnonWallet::Initialise(CExtKey* pExtMaster)
     std::string sError;
 
     bool fFirstRun = pExtMaster != nullptr;
+    AnonWalletDB wdb(*walletDatabase);
 
     {
         LOCK(pwalletParent->cs_wallet);
-        AnonWalletDB wdb(*walletDatabase);
 
         //Load all keys and accounts
         if (!fFirstRun) {
@@ -141,9 +141,23 @@ bool AnonWallet::Initialise(CExtKey* pExtMaster)
             if (!wdb.ReadNamedExtKeyId("master", idMaster))
                 return error("%s: failed to read masterkey id from db", __func__);
             if (!wdb.ReadNamedExtKeyId("defaultaccount", idDefaultAccount))
-                return error("%s: failed to read defaultaccount id from db", __func__);
+                return error("%s: failed to read default account id from db", __func__);
             if (!wdb.ReadNamedExtKeyId("stealthaccount", idStealthAccount))
-                return error("%s: failed to read stealthaccount id from db", __func__);
+                return error("%s: failed to read stealth account id from db", __func__);
+            //Load the change account
+            idChangeAccount.SetNull();
+            if (!wdb.ReadNamedExtKeyId("stealthchange", idChangeAccount)) {
+                //If the change account is not created yet, then create it
+                if (!IsLocked() && !CreateStealthChangeAccount(&wdb))
+                    return error("%s: failed to create stealth change account ", __func__);
+            }
+            //Load the change address
+            if (!idChangeAccount.IsNull() && !wdb.ReadNamedExtKeyId("stealthchangeaddress", idChangeAddress)) {
+                auto address = GetStealthChangeAddress();
+                idChangeAddress = address.GetID();
+                if (!wdb.WriteNamedExtKeyId("stealthchangeaddress", idChangeAddress))
+                    return error("%s: Failed to write stealth change address to wallet db", __func__);
+            }
 
             // Load all accounts, keys, stealth addresses from db
             if (!LoadAccountCounters())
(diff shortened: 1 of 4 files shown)


src/veil/ringct/anonwallet.cpp
```diff
@@ -2452,12 +2459,7 @@ int AnonWallet::AddBlindedInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx,
 
                 //If no change address is set, then generate a new stealth address to use for change
```
blondfrogs: Need to update the comment to be the same as the above one

src/veil/ringct/anonwallet.cpp
```diff
@@ -3172,10 +3174,7 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
...
```
blondfrogs: NIT:  update comment

blondfrogs:
QA needs to make sure they do testing with wallets that
- Already encrypted before this code is executed. 
- Not encrypted 
- New Wallet creation
- Make sure all new transactions use the default changes address
- Make sure if the custom change address is given, that it uses that instead. 
- Should be able to use the rpc call to verify that correct addresses are used

ncrypter90:
@blondfrogs Could you explain me more about the change?

Already encrypted before this code is executed. -> Is this about wallet being encrypted? before transaction?
Not encrypted-> Same as above with wallet being not encrypted?
New Wallet creation-> what to validate with new wallet ?
Make sure all new transactions use the default changes address-> How to check default change address?
Make sure if the custom change address is given, that it uses that instead.-> How to provide custom change address?
Should be able to use the rpc call to verify that correct addresses are used-> What is the RPC call to check ?

ncrypter90:
Hi @blondfrogs @presstab 

Could you provide more details as requested earlier?

I have tested below on this PR:
1. Made multiple transactions from WIn wallet (build 630) to another Mac wallet (build 630) and vice versa, checked change address not being changed whereas with wallet 1.3 release, observed change addresses were different in each transactions. Test Passed. 

However, I still have few questions, Could you provide me further details on below?

Make sure all new transactions use the default changes address-> How to check default address?
Make sure if the custom change address is given, that it uses that instead.-> How to provide custom address ?
Should be able to use the rpc call to verify that correct addresses are used -> What is the RPC cmd ?

@crypt0-fusi0n

seanPhill:
> Make sure all new transactions use the default changes address-> How to check default address?

I noticed _this_ above: ```This also adds getstealthchangeaddress RPC which will print out the default change address.```

> Made multiple transactions from WIn wallet (build 630) to another Mac wallet (build 630) and vice versa, checked new change address are being changed . Test Failed.
Is it a new change address, or just a different stealth destination?

> Should be able to use the rpc call to verify that correct addresses are used -> What is the RPC cmd ?
I guess it should be `gettransaction TXID`

ncrypter90:
Thanks @seanPhill , appreciate it.

Yes i was looking at stealth destination address in the transaction response.
as per our conversation on discord, stealth address when is_change is true is same, and i could check the default stealth change address can be retrieved getstealthchangeaddress.

Could you tell me how to change the custom change stealth address ?

ncrypter90:
@seanPhill @blondfrogs 

One more observation:

I have created a new wallet with PR630, transferred some coins from other wallet to the new and closed the wallet. When I tried to open again, it error out saying error in one of cpp file and I had to remove all the files and resync whole wallet again to make it work.
Finally when I tried to send coin across the wallet hangs while Generating transaction. I waited for an hour. I had 10 zero coins and i tried sending 9. Please find the screen shot below.

Should I ignore this? Do you think it might be related to PR630?
 
image

presstab:
@ncrypter90 yes since this PR adds more databasing to the wallet, if you were to use that same wallet.dat after using this code then it would likely have an error reading the new databasing that was added.

Do you have the error print out?

ncrypter90:
@presstab I have created new wallet for PR630 that seemed to be unstable. later when i switched my old wallet file it rescaned wallet for 10mins and it working fine. Just new wallet creation with code seems to be have some problem.

Sadly i didn't save the error.

I have tested the fix on same change address which is working on WIn ,Mac and Ubuntu with default change address. could you tell me how to use custom address ?
Make sure if the custom change address is given, that it uses that instead.-> **How to provide custom address ?**

presstab:
@ncrypter90 to use a custom change address, you would have to enter it into the change address field on the coincontrol portion of the sending dialog.

ncrypter90:
@presstab Thank you! I will test few transaction with custom address and let you know.

ncrypter90:
@presstab I could not find custom change address in coin control.
Could you confirm whether I can pass without testing custom change address and test separately  when its available in coin control?
@crypt0-fusi0n

crypt0-fusi0n:
yeah looks like the change address field is missing... I've checked older builds as well and not there either...

@presstab What do you think? How should we proceed with this PR?

presstab:
I think it is a missing feature that was already missing in the wallet. If we could not use a custom change address before this PR, then it wasn't exactly part of the scope of this particular code to do so.

I think we should add a custom change address as a new issue to be worked on in a different PR.

crypt0-fusi0n:
ok, we will close this one as passed and will log a new issue re: missing change address

@ncrypter90

ncrypter90:
Thanks @crypt0-fusi0n @presstab 

I will close this as QA passed with below observations.

1.When coins transferred from one wallet to another , destination wallet(630) needs manual rescan zero and rescan ringct to reflect the transaction and amount. Does not happens all the time.
2.Coin control window in MAC is misaligned.

ncrypter90:
Created a new issue:
https://github.com/Veil-Project/veil/issues/662

Issue #509, opened 2019-04-13
GUI Bug: Generate Address Icon doesn't work


Rock-N-Troll:
Clicking the "Generate Address" circle arrow icon in the bottom left of the Receive screen shows a pop up for "address generated" but it doesn't do anything

![image](https://user-images.githubusercontent.com/34344520/56085776-0ef97900-5e17-11e9-95c8-62641f232a13.png)

but the debug log shows:

> 
> 2019-04-13T22:06:13Z GUI: QObject::connect: No such slot AddressesWidget::hideThisWidget() in qt/veil/addresseswidget.cpp:283
> 2019-04-13T22:06:13Z GUI: QObject::connect:  (receiver name: 'AddressesWidget')
> 2019-04-13T22:06:15Z GUI: QObject::connect: No such slot ReceiveWidget::hideThisWidget() in qt/veil/receivewidget.cpp:203
> 2019-04-13T22:06:15Z GUI: QObject::connect:  (receiver name: 'ReceiveWidget')
> 2019-04-13T22:06:18Z GUI: QObject::connect: No such slot AddressesWidget::hideThisWidget() in qt/veil/addresseswidget.cpp:283
> 2019-04-13T22:06:18Z GUI: QObject::connect:  (receiver name: 'AddressesWidget')
> 
>

veilgets:
Bounty: 166.67

veilgets:
Developer: rikublock

rikublock:
ready for QA

veilgets:
re-opened

CaveSpectre11:
Ready for QA/Code Review - PR #652

veilgets:
Developer: CaveSpectre11

codeofalltrades:
closed via #652

PR #652, opened 2019-08-12
[QT] Fix address creation on Receive tab

CaveSpectre11:
### Problem
Clicking the "Generate Address" circle arrow icon in the bottom left of the Receive screen shows a pop up for "address generated" but doesn't generate a new address.

### Root Cause
`ReceiveWidget::generateNewAddress()` has multiple uses.  It's called by `setWalletModel()`, `refreshWalletStatus()`, and when you click on the "Generate Address" icon.  The first use is to generate a new address if you don't already have one.  The second is to get the label of your initial address.  These two uses required the original `generateNewAddress()` to need a change; so that it wouldn't create a new address every time you started the wallet.

That fix introduced the issue in question.  Because the fix was made by not generating a new address if an address already exists.  Since the nature of having the wallet up and running means you have an address; a new address wasn't created.

### Solution
Add a parameter to the function, `bOnDemand`, that defaults to false so that the other two cases don't have to change their call.  For the click action, add the "true" parameter; and conditionalize the address existence check.  Now the other two uses will pass through the same code it had before; but the on clicked address creation will go down the path intended.

### Bounty PR
#509 

### Bounty Payment Address
`sv1qqpj9l2f883k7ucs88xdzrcuus8wr6jzrnk85quv3p2u90zwyy6d5tqpq0fl2ksmvlsf5hfddf8ezqnph5q6zf05cmac28g6ytafdzp8aufsuqqqwk9ep3`

### Unit Testing Results
Very simple
1. Click the generate address icon on the Receive tab
2. Observe the "Address generated" result
3. Observe the QR code and address change.
3. Change to the Address Book tab to confirm the newly created address has been added.

### DIFF
--- a/src/qt/veil/receivewidget.cpp
+++ b/src/qt/veil/receivewidget.cpp
@@ -81,17 +81,17 @@ void ReceiveWidget::on_btnCopyAddress_clicked() {
 }
 
 void ReceiveWidget::generateNewAddressClicked(){
-    if(generateNewAddress()) openToastDialog("Address generated", mainWindow->getGUI());
+    if(generateNewAddress(true)) openToastDialog("Address generated", mainWindow->getGUI());
     else openToastDialog("Wallet Encrypted, please unlock it first", mainWindow->getGUI());
 }
 
-bool ReceiveWidget::generateNewAddress(){
+bool ReceiveWidget::generateNewAddress(bool isOnDemand){
     // Address
     interfaces::Wallet& wallet = walletModel->wallet();
 
     std::string strAddress;
     std::vector<interfaces::WalletAddress> addresses = wallet.getLabelAddress("stealth");
-    if(!addresses.empty()) {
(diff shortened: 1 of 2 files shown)


CaveSpectre11:
Closes #509 
Ready for Code Review/QA

CaveSpectre11:
Ready for QA/Code Review - PR #652

CaveSpectre11:
> Minor change request. The codebase typically uses `f` or `is` as the prefix for a boolean. Will you update these to use either `is` or `f`?

No problem.  Since it's crossing two files I'll take care of it when I have access to git and rebase it.

CaveSpectre11:
> Minor change request. The codebase typically uses `f` or `is` as the prefix for a boolean. Will you update these to use either `is` or `f`?

Changed to `isOnDemand` and ready for re-review

ncrypter90:
QA passed!

PR #654, opened 2019-08-16
[Mining] Create placeholder tx for CoinStake tx

### LINKED ISSUE #590: codablock Code Fix Suggestion: Add Additional Placeholder Entry To vtx For PoS (Currently Works For PoW)
[codablock](https://github.com/codablock) via Discord:

> going through a few changes/fixes I did when I tried out some staking a few months ago. One issue was that my staked blocks were rejected by my own node due to TX inputs missing. Looking at how you guys fixed in the mean time brings me to this: https://github.com/Veil-Project/veil/blame/master/src/miner.cpp#L311
> 
> The actual reason for the issue was a different thing however. When a PoS block is created (vs a PoW block), the transaction at index 1 is replaced with the coinstake TX: https://github.com/Veil-Project/veil/blame/master/src/miner.cpp#L393
> 
> But the original TX is actually a TX added by the block template code
> So if you replace it, TXs that depend on this one (spend an output from that one) and were also included in the same block would become invalid after the TX at index 1 is replaced
> Making the whole block invalid
> The correct fix would be to add an additional place holder entry to vtx, the same as done for PoW here: https://github.com/Veil-Project/veil/blame/master/src/miner.cpp#L144
> 
> So that you later replace the placeholder
> This should also allow you to revert the hotfix you did in line 311
> This should also fix the "issue" of the mempool not getting fully cleared when PoS blocks come in :wink:
> (btw, don't have time for a PR, so just writing this here so that someone can take it)

codeofalltrades:
### Problem ###
PoS blocks are rejected by own wallet due to missing TX inputs

### Root Cause ###
When a PoS block is created the transaction at index 1 is replaced with the coinstake TX. This is replacing a valid tx from the block template. This replacement causes the rejected block. 

### Solution ###
When a PoS block is created a placeholder tx at index one that will be replaced with CoinStake tx.

### Bounty PR ###
fixes #590

### Bounty Payment Address ##
`sv1qqpj0zn0pyd2gp7w5vs93ja2rq0s6yrwvmtcvvd6lg5awdpuggsh32cpq2seytp7tg6mlqvzn2zjl9ug63gsga0q5kgzh5sjmg0uz32r2krh6qqq4tpdvx`

### Unit Testing Results ###
Mine blocks without additonal txs
Mine blocks with additonal txs
Stake blocks with only the stake and reward
Stake blocks with stake, reward, with additonal txs
From each test verify you can spend the rewards


codeofalltrades:
@veilgets please tag ready for code review

src/miner.cpp
```diff
@@ -143,6 +143,7 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
     // Add dummy coinbase tx as first transaction
     pblock->vtx.emplace_back();
     pblock->fProofOfStake = fProofOfStake;
+	if (fProofOfStake) { pblock->vtx.emplace_back(); } // Add placeholder tx - will be replaced with CoinStake tx.
```
codablock: You also need to add a dummy entry to vTxFees and vTxSigOpsCost

codablock:
FYI, this PR allows reverting 1403229e5f9ed6f877dd9b5ab7d0767ece4782ae as the bug hot-fixed by it is correctly fixed in this PR.

src/miner.cpp
presstab: @codeofalltrades was this part of the code meant to be replaced? Now it only add to fees and sigopscost. I gave a positive code review because I was not sure the intention, without testing this myself, it seems like it may no longer be doing what was intended?

ncrypter90:
@codeofalltrades 
As discussed, Validation of this fix requires latest testnet needs to be running which is not ready yet.

I will setup testnet once it is available.

@presstab @crypt0-fusi0n

codeofalltrades:
Closing, there have been too many changes to master since this change. I will create a new pull with the changes.

PR #669, opened 2019-10-01
[RPC] getzerocoinsupply height not required

mimirmim:
While working through the RPC, I had noticed in the RPC command getzerocoincoinsupply, height is assumed to be required when it isn't. Small fix.


codeofalltrades:
utACK

mimirmim:
I will be adding more as I see them but I worry too much about messing with it too much. The RPC inherited from Bitcoin is just an absolute mess. Specifically the mixed cases. You got snakeCase then_this and thanthat.

PR #676, opened 2019-10-10
[Testnet] Update testnet params and fix amount

blondfrogs:
When syncing with the latest code. Nodes were rejecting the block at height 1. This was because of a miss calculation on the amount. This PR fixes that, as well as adds another seed node for testnet


codeofalltrades:
utACK

PR #679, opened 2019-10-17
Add P2P version tracker & update warnings

### LINKED ISSUE #404: Feature Request — Warn user of possible wallet update needed

JSKitty:
This PR closes #404 by adding a check onto every new connection that measures and saves the connection's peer subversion, this can then be used to measure the average network version, and figure out if the client is "behind", "forward" or "current" in version integers, if the client is behind, the check will trigger a message in warnings.cpp which displays a warning of `"Over 50% of our peers are running a newer wallet version than yours. An update may be available."`

The update tracker is relatively simple, but the functions of it must be known clearly, so that it doesn't get abused by malicious actors, and doesn't false-positive on specific version integers.

- Atleast 50% of historical peers, must be a higher version before an update warning is triggered.
- Atleast 2 historical peers minimum must be a higher version before an update is considered.

This means at minimum, if you have 4 connections, two are older or matching wallets, two are new wallets higher than our own, the update warning **WILL** fire.

If you connect to only 1 peer, and they are a higher version, the update warning will **NOT** fire, due to the minimum requirement of 2 higher-ver peers, to help prevent false-positives.

No arbitrary data can be inserted into the update tracker, it will not display a URL or version text when an update triggers, it will just purely notify that an update may be possible, due to >50% of peers having higher subversions.

*This is a rebase and upgrade to the closed #627 PR which had two vulnerabilities. The new code removes the possibility to abuse such parsing bugs.*

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
+// Copyright (c) 2018-2019 The Veil Developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -10,8 +11,10 @@
 #include <blockencodings.h>
 #include <chainparams.h>
 #include <checkpoints.h>
+#include <clientversion.h>
 #include <consensus/validation.h>
 #include <hash.h>
+#include <warnings.h>
 #include <validation.h>
 #include <merkleblock.h>
 #include <netmessagemaker.h>
@@ -80,6 +83,39 @@ std::map<uint256, COrphanTx> mapOrphanTransactions GUARDED_BY(g_cs_orphans);
 std::map<int, CBlock> mapStagedBlocks;
 static constexpr int ASK_FOR_BLOCKS = 50; //How many blocks to ask for at once
 static CCriticalSection cs_staging;
+
+bool fUpdateCheck = true;
+
+bool fShouldUpgrade = false;
+
+// --------------------------------------------------
+// Upgrade Majority
+// This is the magic number for deciding when enough
+// peers have a higher version to be considered the
+// 'majority' of the network for updating the software.
+
+// Higher means more secure and less false-positives,
+// but lengthens time before the update is triggered.
+
+// Peers / nUpgradeMajority = Minimum peers for update.
+// An additional protection against false-positives
+// is the Updater will not fire unless there's atleast
+// two higher peers, minimum, even if the percentage
+// majority (With 4 or less peers) is high enough.
+
+int nUpgradeMajority = 2; // 50% of peers is 'majority'.
+// --------------------------------------------------
+
+// The quantity of version types. E.g: major.minor.patch.build.
+// Veil generally uses four of them, so this system will parse
+// 4 of them, but ignore the "build" version to prevent pre-release
+// false-positives.
+int nVersionTypes = 4;
+
+// Track all peer version differences
+int nHigherVerPeers = 0;
+int nCurrentVerPeers = 0;
+int nLowerVerPeers = 0;
 
 void EraseOrphansFor(NodeId peer);
 
@@ -2067,9 +2103,12 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
 
         // Potentially mark this peer as a preferred download peer.
         {
-        LOCK(cs_main);
-        UpdatePreferredDownload(pfrom, State(pfrom->GetId()));
+            LOCK(cs_main);
+            UpdatePreferredDownload(pfrom, State(pfrom->GetId()));
         }
+
+        // Check the peer's version and compare it to us
+        CheckForUpdates(pfrom->addr.ToString(), pfrom->strSubVer);
 
         if (!pfrom->fInbound)
         {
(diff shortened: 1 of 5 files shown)


JSKitty:
**Ready for QA**

Bounty Address: `sv1qqpx292fphza99xahxs0lrqwsjjjvgly53aggm37rw3tj4qt5jhd5yqpqfje7wmm9utzjynq85yyhpcka47knqzkgnj5clal8s8dy5nqq24hjqqquky79f`

src/net_processing.cpp
```diff
@@ -81,6 +83,39 @@ std::map<int, CBlock> mapStagedBlocks;
...
+// --------------------------------------------------
+
+// The quantity of version types. E.g: major.minor.patch.build.
+// Veil generally uses four of them, so this system will parse
+// 4 of them, but ignore the "build" version to prevent pre-release
+// false-positives.
+int versionTypes = 4;
+
+// Track all peer version differences
+int higherVerPeers = 0;
```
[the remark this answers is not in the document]
JSKitty: Change made in d4619d9

codeofalltrades:
Please squash into 1 commit.

CaveSpectre11:
Still processing through the code.  Not so much the surface functionality, but the potential risk (if there is one).    However two things of minor note.
- Add the Veil copyright to the top of the files you're changing.
- And rebase down to one commit (after you `git commit` the copyright edits, you should be able to `git rebase -i HEAD~2`, which will pull up an editor with the 3 commits and you can change the two trivial commits from `pick` to `fixup`.  You might have to `push -f` after the rebase)

JSKitty:
> 
> 
> Still processing through the code. Not so much the surface functionality, but the potential risk (if there is one). However two things of minor note.
> 
>     * Add the Veil copyright to the top of the files you're changing.
> 
>     * And rebase down to one commit (after you `git commit` the copyright edits, you should be able to `git rebase -i HEAD~2`, which will pull up an editor with the 3 commits and you can change the two trivial commits from `pick` to `fixup`.  You might have to `push -f` after the rebase)

Alrighty! Thanks for the git commands, will get that done later today.

JSKitty:
@CaveSpectre11 Rebase done ✌

src/net_processing.cpp
```diff
@@ -4314,6 +4353,125 @@ bool PeerLogicValidation::SendMessages(CNode* pto)
...
+        // Minor
+        if (i == 1) {
+            if (ourVer[i] > peerVer[i])
+                fHigherMinor = true;
+        } else
+        // Patch
+        if (i == 2) {
+            if (ourVer[i] > peerVer[i])
+                fHigherPatch = true;
+        }
```
CaveSpectre11: This logic has a flaw.  It assumes that the more significant version is equal if it's not higher.  It doesn't handle if the more significant version is actually lower.  e.g. if you have 1.0.4 and they have 1.0.5; it works.  Major and Minor will be false, Patch will be true.  But if you have 1.3.9 and they have 1.4.0 (for example), major is false, minor is false, patch is true; therefore `nLowerVerPeers` is incremented; even though they have a higher version.

src/net_processing.cpp
```diff
@@ -4314,6 +4353,125 @@ bool PeerLogicValidation::SendMessages(CNode* pto)
...
+
+std::vector<int> ParseVersion(const std::string subver)
+{
+    try {
+    /*
+        This parses a subversion string into an array of four
+        integers that represent the subversion's raw versioning.
+
+        e.g: "/Veil:1.0.3.99/" = [1, 0, 3, 99]
+
```
CaveSpectre11: I would rather see a method here that returns a 32 bit number representing the version:
```
version = major << 24;
version += minor << 16;
version += patch << 8;
```
This way you have a number that is easily comparable and isn't at risk of the problem I outlined where a higher order version is greater than a peer, but a lower order version is less than the peer.  It makes for simpler logic in the comparisons.

src/net_processing.cpp
```diff
@@ -2067,10 +2103,13 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
...
         {
-        LOCK(cs_main);
-        UpdatePreferredDownload(pfrom, State(pfrom->GetId()));
+            LOCK(cs_main);
+            UpdatePreferredDownload(pfrom, State(pfrom->GetId()));
         }
 
+        // Check the peer's version and compare it to us
+        CheckForUpdates(pfrom->addr.ToString(), pfrom->strSubVer);
+
```
CaveSpectre11: I'm not sure about this here.  I could be wrong, but looking at the whole logic flow, it looks like we're going to check each one individually but start setting things after we have the first few (see later)

src/net_processing.cpp
```diff
@@ -4314,6 +4353,125 @@ bool PeerLogicValidation::SendMessages(CNode* pto)
...
+    int nPeersForUpgrade = (nHigherVerPeers + nCurrentVerPeers + nLowerVerPeers) / nUpgradeMajority;
+
+    // Ensure minimum is atleast 2 peers
+    if (nPeersForUpgrade < 2)
+        nPeersForUpgrade = 2;
+
+    // Check if majority consensus is met
+    ((nHigherVerPeers >= nPeersForUpgrade) ? fShouldUpgrade = true : fShouldUpgrade = false);
+
+    SetfUpdateFound(fShouldUpgrade);
```
CaveSpectre11: This is where I'm having trouble since we're going to run this each time we look at a peer.  So if the first two are higher, we're going to set ShouldUpgrade before we have looked at everyone (presumably we will rescind that back to "false" as we see more peers).

I would much rather see logic that calls into GetNodeStats() and processes the whole list of peers, and then makes a determination after it's processed the whole list; rather than doing all these calculations each time we see a peer.

The other part that makes  me nervous is a few lines above (and I haven't researched the full context of the call to this routine), but the fact that we keep incrementing the numbers without any bounds is risky.  In other words, we don't really have any concept of the peer has been looked at before or not.  So if this is called each time a peer shows up; if that peer keeps disappearing and re-appearing, they may get counted multiple times.  I would have to dig more, but I think using the GetNodeStats() would be a better route to go; however if that's the case (that someone can get counted multiple times), they could actually manipulate it by bouncing in and out enough to throw your numbers off.  Perhaps as a test, you can look into the logic that causes these routines to call, and watch nPeersForUpgrade as it relates to the length of the vector derived from the GetNodeStats() list (or some other way to compare to the current number of known peers).

src/net_processing.cpp
JSKitty: This review seems correct, although I haven't been able to test this in-action yet.

src/net_processing.cpp
JSKitty: Using GetNodeStats() will definitely cause this mechanic to be more "realtime" instead of running off of historical data, would also remove the need for extra variables, at the cost of _slightly_ more processing on every new connection from having to scan through the full peers list.
It would also solve the issue of peers purposefully connecting/disconnecting to increment version integers essentially infinitely.
Good catch and suggestion, I think this is worthwhile to swap to, just need to find the time to rewrite it.

src/net_processing.cpp
JSKitty: Yep.

src/net_processing.cpp
JSKitty: This is something a little outside of my scope of knowledge in C++ 😓 I guess this would reduce the amount of logic needed to make the comparisons?
Have never written anything like this previously.

CaveSpectre11: Yes.  What that does is it "bit shifts" the major version over 24 bits (3 bytes), the minor over 16 bits (2 bytes), etc.

So you end up (in hex) with 01000408 (for 1.0.4.8) and 01000406 (for 1.0.4.6).  Now your comparison is simple:   01000408 > 01000406.  

Long story short.  The bug you have is that 1.3.9 will be "higher" than 1.4.0; but this logic.
1030900 < 1040000.  Rendering the whole for loop to check major, minor and patch unnecessary.

Let me know (here or discord) if the above is clear or if you have more questions.

CaveSpectre11:
Just noticed that validate.cpp has a check as well; it's not as elaborate, but it might make sense to work off what's already existing:

```
        // Check the version of the last 100 blocks to see if we need to upgrade:
        for (int i = 0; i < 100 && pindex != nullptr; i++)
        {
            int32_t nExpectedVersion = ComputeBlockVersion(pindex->pprev, chainParams.GetConsensus());
            if (pindex->nVersion > VERSIONBITS_LAST_OLD_BLOCK_VERSION && (pindex->nVersion & ~nExpectedVersion) != 0)
                ++nUpgraded;
            pindex = pindex->pprev;
        }
        if (nUpgraded > 0)
            AppendWarning(warningMessages, strprintf(_("%d of last 100 blocks have unexpected version"), nUpgraded));
        if (nUpgraded > 100/2)
        {
            std::string strWarning = _("Warning: Unknown block versions being mined! It's possible unknown rules are in effect");
            // notify GetWarnings(), called by Qt and the JSON-RPC code to warn the user:
            DoWarning(strWarning);
        }

```

ianGre:
Before this is looked at again, is have two concerns, having read the comments, not the code. 
1. Potential adverse effect on performance.
2. Unofficial version numbers (not proven by PGP signature, for example) could be used to trick peers into displaying the update warning and may be used as part of a spearfishing attack or a  campaign to get malware accepted as an official update.

CaveSpectre11:
> Before this is looked at again, I have two concerns, having read the comments, not the code.

> 2. Unofficial version numbers (not proven by PGP signature, for example) could be used to trick peers into displaying the update warning and may be used as part of a spearfishing attack or a  campaign to get malware accepted as an official update.

Thank you for your feedback.  Yes, the "nPeersForUpgrade = 2" definitely struck me as unsafe.  I thought I had a comment to make it the majority of nodes, but I don't see it when I browse quickly through.  Seeing as there is already a check in validate.cpp; this might not be even necessary at all.

Issue #270, opened 2019-01-11
Disappearing Text in 1.0.0.9


Larcea:
I have this issue with 1.0.0.9.

However I have pegged it to the "Dark" appearance for the MacOS Theme.

When I switch to "Light" the text all comes back.

Dark Theme:
![screen shot 2019-01-11 at 3 47 23 pm](https://user-images.githubusercontent.com/38234120/51061819-29bf1600-15ba-11e9-9689-171a003376d1.png)

Light Theme:
![screen shot 2019-01-11 at 3 52 03 pm](https://user-images.githubusercontent.com/38234120/51061840-3ba0b900-15ba-11e9-921c-12357dd890f9.png)

Issue: Text visually disappears.
OS: MacOS (Mojave)
Build: 1.0.0.9
Workaround: Change from Dark to Light theme.
Related Issue: #60

Additional Information:
- Compiled wallet with `Make Deploy` option.

Attempted resolutions:
- Recompiled wallet
- Recompiled wallet with `sudo` rights
- Opened wallet with `-resetguisettings` command line.

seanPhill:
This has now also been observed by an Arch Linux user who had Dark Mode enabled. The wallet displayed white text on a white background and was therefore illegible.

The user will try the CLI wallet instead.

codeofalltrades:
Bitcoin core disabled the dark setting on macOS
https://github.com/bitcoin/bitcoin/commit/df5131bd41ea86c5b404cd6091284d9ac62368e7

Issue #671, opened 2019-10-05
Blank page in wallet on checking v1.0.4.7-007651e15 on Settings page.


seanPhill:
When the Veil v1.0.4.7-007651e15 wallet is redisplayed (having been minimised for a while) and the settings page is already selected, the central (main content) section is completely blank.

To display the settings page it is necessary to firstly select "Overview", then re-select "Settings."

I am using MacOS 10.13.6 (High Sierra), NOT using Dark Mode.
The wallet was not self-compiled, so far as I can remember. 

I am using a Mac Pro with 32GB RAM and an SSD.

The previous behaviour, of display the content window should be maintained.

For comparison I am launching the Veil wallet (same version) on my iMac (8GB, HDD, same OS version) which has been turned off for a couple of days.
It opened up normally, and is showing the Syncing dialog, with the Overview screen visible behind it (non-transparent), just around the outside of the synchronising dialog.

I clicked "Hide" and the overview page shows as normal, and tapping the Settings page shows as it normally does. (I'll check back here when I get that wallet to be not seen for a while and then see if the central section is visible when I come back to it.)

seanPhill:
Yes, I redisplayed the wallet from the Dock, where it had been minimised for only a minute (with the Settings screen active) and the central section was blank.

So far it has not happened when the Overview screen was selected when it was minimised (for a couple of minutes).

seanPhill:
Even after a long time with the screen asleep and the wallet minimised it does not happen if you had the Overview screen selected when it was minimised, only with the Settings screen.
This only started with version 1.0.4.7. I would have noticed it earlier if it had been a present behaviour in 1.0.3.x.

seanPhill:
It also happens even if not minimised, but out of focus.

CaveSpectre11:
Fixed with PR #687

PR #681, opened 2019-10-23
[GUI] Explicitly disable Dark Mode appearance on macOS

codeofalltrades:
### Problem ###
Bitcoin Core's gui is pretty broken for a someone using macOS in "Dark Mode"; the biggest issue being lots of white text on a white background, leaving most inputs, the rpc console etc unusable.

This is likely something we'll have to wait for Qt to fix/support, so it's probably easiest to just disable for now, rather than provide a broken UI; 

### Root Cause ###
Apple documentation on "Opting out of Dark Mode" is here.

>    If you need extra time to work on your app's Dark Mode support, you can temporarily opt out by including the NSRequiresAquaSystemAppearance key (with a value of true) in your app’s Info.plist file. Setting this key to true causes the system to ignore the user's preference and always apply a light appearance to your app.

### Solution ###
Disable dark mode

### Testing ###
1. Enable dark mode
2. Launch and use the wallet
3. Verify text is visible

fixes: #270

### DIFF
--- a/share/qt/Info.plist.in
+++ b/share/qt/Info.plist.in
@@ -99,6 +99,9 @@
 
   <key>LSAppNapIsDisabled</key>
     <string>True</string>
+
+  <key>NSRequiresAquaSystemAppearance</key>
+    <string>True</string>
   
   <key>LSApplicationCategoryType</key>
     <string>public.app-category.finance</string>


CaveSpectre11:
utACK https://github.com/Veil-Project/veil/pull/681/commits/b7af393411ac758b7b917aa895b5ad29c8765ff5

@codeofalltrades perhaps post screen grabs of before and after?  Not necessary, but it would help for those without the necessary MacOS version.  

Also, is this general for all supported MacOS versions, or does it apply to "version x+"; that would help QA get it looked at by someone with the correct version.

ncrypter90:
@codeofalltrades @crypt0-fusi0n 
Few observations:
1. Floating "View explorer" looks odd.
2. Watch the screen record video, Setting goes blank after closing other windows (ex: debug window).

Screen Shot 2019-11-11 at 9 19 52 pm

[Veil_681.mov.zip](https://github.com/Veil-Project/veil/files/3830958/Veil_681.mov.zip)

crypt0-fusi0n:
Thanks @ncrypter90 

@codeofalltrades Not serious issues but not sure if these are easy to resolve?

codeofalltrades:
Re @ncrypter90, @crypt0-fusi0n 
> 1. Floating "View explorer" looks odd
Yes looks odd. It looks like a tooltip styling problem. In the windows version the background is also most black. Do you have a screenshot from a wallet without this fix? 

> 2. Setting goes blank after closing other windows (ex: debug window).
This issue started occurring with wallet v1.0.4.6 or v1.0.4.7. Sean logged it here #671

codeofalltrades:
Please log floating view in explorer tooltip as separate issue. I checked the code and I have a fix for that.

ncrypter90:
@codeofalltrades I have logged a new issue for floating view explorer.

https://github.com/Veil-Project/veil/issues/685

@crypt0-fusi0n

PR #651, opened 2019-08-11
[GUI][QT] Fix "no such slot" and "no matching signal" log errors

CaveSpectre11:
### Problem
Clicking on an address in the Address Book page to pull up the Copy|Edit menu generates "No matching signal" errors in debug.log the first time you use the menu.

### Root Cause
The Qt MetaObject `connectSlotsByName` searches for `connect()` code with slot objects that conform to the format `on_<object-name>_<signal-name>`.  Since the signals for that pulldown (`on_btnCopyAddress_clicked()`, `on_btnEditAddress_clicked()` and `on_btnDeleteAddress_clicked()`) conformed to that format; connectSlotsByName picked them up when they shouldn't have been.

### Solution
Rename these slots in the format `on<Object-name>Clicked` to avoid moc from picking them up inadvertently.

### Additional Scope Included
Several other instances of the same issue were found and corrected in this PR.

_Also included in this PR:_ (https://github.com/Veil-Project/veil/commit/ed31991faf0acddd65e48fe1e22ab06730a5ff5b)
**Problem**
Clicking away from certain pages in the QT wallet generates a "no such slot[...]hideThisWidget()" error in the log file.

**Root Cause**
The fade out logic added to the Settings, Receive and Address Book pages was found on the internet, and the comment of "now implement a slot called hideThisWidget()" was removed and not implemented.   In one instance, an additional "->hide()" call was added. 

**Solution**
Rather than remove the connect for the finish signal, `hideThisWidget()` was implemented in each of the three cases to implicitly hide the widget on the finish signal; and thus provides the infrastructure for further effects to be added in the future if desired.

### Bounty PR
#541 

### Bounty Payment Address
`sv1qqpj9l2f883k7ucs88xdzrcuus8wr6jzrnk85quv3p2u90zwyy6d5tqpq0fl2ksmvlsf5hfddf8ezqnph5q6zf05cmac28g6ytafdzp8aufsuqqqwk9ep3`

### Unit Testing Results
**With v1.0.3 Qt Wallet**
1. Start QT Wallet
2. observe during startup, near the end of loading" connect slots by name errors for the copy address button and the address book button.
```
2019-08-11T14:43:38Z CleanBlockIndexGarbage: Erasing 37481 irrelivant indexes
2019-08-11T14:43:40Z GUI: QMetaObject::connectSlotsByName: No matching signal for on_btnCopyAddress_clicked()
[...]

Issue #541, opened 2019-04-24
GUI Bug: Debug Log Error for Copy Address Functionality


Rock-N-Troll:
![image](https://user-images.githubusercontent.com/34344520/56689311-cb213200-66a8-11e9-8e4e-cf63413da720.png)

Occurred when using the right click copy functionality in the address book

MatWaller:
Seems to be another Windows specific bug, working as intended on;

```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

veilgets:
Bounty: 187.5

veilgets:
Developer: CaveSpectre11

CaveSpectre11:
Ready for QA/Code Review - PR #651

codeofalltrades:
closed with pull #651

PR #687, opened 2019-11-17
[Qt] Remove hide() on hideEvents

### LINKED ISSUE #671: Blank page in wallet on checking v1.0.4.7-007651e15 on Settings page.

CaveSpectre11:
### Problem
When restoring the wallet from a minimized state, or when completely covered by other windows; the receive, addresses, and settings tabs don't redraw.

### Root Cause
This problem was introduced with PR #651 in response to Issue #541, where "no such slot" errors were generated when switching between windows.  The problem with the errors was a missing hideThisWidget() call.  That routine traditionally calls hide() to hide the widget.  However, when a widget is hidden on minimize, the showEvent() handler doesn't get called on restoration; so it didn't automatically redraw the widget.  One had to click away and back to get the showEvent() to trigger.

### Solution
After some research, it seems that hiding the widget isn't really necessary; so instead of putting conditions to prevent the hide when minimizing (and then dealing with all the other corner cases) the slot has been removed.  This was done to prevent the problem introduced with #651, and to also fix the same problem (no such slot) for the Overview and Send Coins tabs as well.

### Unit Testing Results
**With v1.0.4.7 Qt Wallet**
1. Start QT Wallet
2. Minimize wallet when on the Overview tab
3. observe "No such slot" errors in the log file.
4. restore focus to the window.
5. click to the "receive" tab
6. minimize and then restore the window
7. observe the screen is now blank
8. repeat for "addresses" and "settings" tabs.

**With PR code loaded**
Restart wallet and repeat 1-8 above; observe none of the errors and issues are seen.

closes #671

### DIFF
--- a/src/qt/overviewpage.cpp
+++ b/src/qt/overviewpage.cpp
@@ -569,5 +569,4 @@ void OverviewPage::hideEvent(QHideEvent *event){
     a->setEndValue(0);
     a->setEasingCurve(QEasingCurve::OutBack);
     a->start(QPropertyAnimation::DeleteWhenStopped);
-    connect(a,SIGNAL(finished()),this,SLOT(hideThisWidget()));
 }
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -1051,7 +1051,6 @@ void SendCoinsDialog::showEvent(QShowEvent *event){
     a->setEndValue(0);
     a->setEasingCurve(QEasingCurve::OutBack);
     a->start(QPropertyAnimation::DeleteWhenStopped);
-    connect(a,SIGNAL(finished()),this,SLOT(hideThisWidget()));
 }
 
 SendConfirmationDialog::SendConfirmationDialog(const QString &title, const QString &text, int _secDelay,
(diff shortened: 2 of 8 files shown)


CaveSpectre11:
This PR addresses Issue #671

mimirmim:
utACK. Looks good, all UI based and conceptually from what I am looking at it should fix it. I would like one or two others to verify on the GUI that this has indeed fixed the issue posted.

Issue #682, opened 2019-11-03
RPC Calls Need to be Checked and Fixed for RingCT Integration


CaveSpectre11:
### Short Description
A number of RPC commands need to be checked for Ring CT Integration.

### Detailed Description
@4x13 has flagged a number of RPC calls that are likely to have issues with RingCT.  They need to individually be evaluated and any issues corrected.  These may be split into separate issues, but until the individual details are known about each one, this will serve as a placeholder.

- [ ] createrawtransaction
- [ ] decoderawtransaction
- [ ] getblock
- [ ] getblockchaininfo
- [ ] getnetworkinfo
- [ ] getinfo
- [ ] getblockhash
- [ ] getnewaddress
- [ ] getrawmempool
- [ ] getrawtransaction
- [ ] gettransaction
- [ ] gettxout
- [ ] listunspent
- [ ] sendrawtransaction
- [ ] signmessage
- [ ] signrawtransaction
- [ ] signrawtransactionwithwallet
- [ ] verifymessage

mimirmim:
I had done some work on this with https://github.com/Veil-Project/veil/pull/688.

This has enabled the following to be updated:
`getrawtransaction`
`decoderawtransaction`
`getblock`

`gettransaction` additionally had the `debug:` section removed.

mimirmim:
`createrawtransaction` is basecoin
`createrawbasecointransaction` is actually for CT

Need to change:
* `createrawtransaction` new code for RingCT
* `createrawcttransaction` move the code from `createrawbasecointransaction`
* `createrawbasecointransaction` for basecoin for the interim.

mimirmim:
These also need to be added / changed:
* `fundrawtransaction`
* `fundrawcttransaction`
* `fundrawbasecointransaction`

mimirmim:
Looks like this only works for basecoin so variants need to be created too:
* `sendrawtransaction`
* `sendrawcttransaction`
* `sendrawbasecointransaction`

mimirmim:
sendrawtransaction only goes to user mempool. Needs fix asap.

PR #696, opened 2019-11-30
[RPC] correct listunspent for ct, ringct and zerocoin

CaveSpectre11:
### Problem
`listunspent` only returns basecoin results

### Root Cause
`listunspent` specifically skipped anything that wasn't a standard transaction (skipping CT and RingCT) and skipped zerocoin transactions.

### Solution
Add support for CT, RingCT and zerocoin into `listunspent`.

### Release Notes
- [RPC] listunspent now returns the list of all unspent transactions, rather than just basecoin.
- [RPC] listunspent now properly functions when specifying value based query options.
- [RPC] listunspent now returns unspent immature transactions, not just mature transactions.

### Developer Notes
A few different places in the code use `AvailableCoins()` to collect the unspent list of basecoins, and use other methods to collect the list of other coin  types.  To handle this, a parameter was added with a bitmask to select what coin types you would like `AvailableCoins()` to collect.  During refactoring efforts, these alternate methods of collecting the coin lists should be migrated to use `AvailableCoins()`.  For now the default for the bitmask is `FILTER_BASECOIN`, so that the other callers aside of `listunspent` function as they have been.

### Known Issues
- Codewise: a few places should eventually have the base code updated to handle zerocoin and stealth transactions.  Things like "isSpent" and "GetValue", etc.   This work can be done as a later refractoring PR.
- The help documentation for `listunspent` needs some work, including the additional fields added.  This also isn't high priority at the moment, and there's a lot of other RPC commands that need revisions to their help documentation.  That can be rolled into a separate PR as well.

### Fixes of note
A couple issues were uncovered with this work.  
- First, the value calculations used for the query_options were limited to a signed 32 bit number; and as such likely don't work when the value of the unspent transaction is greater than 22 veil.  Likewise; values that would wrap and fall into a negative number would be skipped by the default filter (0 to 300m).
- Secondly; a parameter was missing in the creation of the COutput object; thus effecting when watched outputs are being used.

### Issue
Part of #682 

### Unit Testing Results
Rather than blasting with a ton of outputs; some things to note:
1) watch only addresses were not tested
2) zerocoin spends had to be handled specifically.  It has been tested, but could probably use more.
3) spent ringct transactions also had to be handled specifically; likewise tested.
4) AvailableCoins is used several different places, like selecting the outputs for spending, as well as some places where it collects the wallet value.  Aside of a full regression test, some of these functions should be checked as well.

Some techniques used
1) filtering by block and comparing with `gettransaction` for that block
2) insuring zerocoin has the proper `pubcoinhash` and `serialhash`, and accurately reflects `listmintedzerocoins`
3) insuring ringct and ct have the proper `stealth_address` and ct has `stealth_destination`.
4) comparing the amounts of each type to `getamounts`

Some of the linux commands used
`./veil-cli listunspent | grep -B 1 -A 12 \"ct`
`./veil-cli listunspent | grep -B 1 -A 12 \"ringct | grep -e lth_ad -e amount`
`./veil-cli listunspent | grep -B 1 -A 12 \"zero | grep serial | sort`
` ./veil-cli listunspent 0 999999 [] true '{ "maximumAmount": 0.05 }' | grep -B 1 -A 12 \"ringct | grep -e stealth_address -e amou`
`./veil-cli listunspent 0 100 | grep txid`

### DIFF
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -1,5 +1,6 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
+// Copyright (c) 2018-2019 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -433,6 +434,8 @@ public:
 
     CTxOut ToTxOut() const { return CTxOut(nValue, scriptPubKey); }
 };
+
+static const uint32_t EPHEMERAL_PUBKEY_LENGTH = 33;
 
 class CTxOutCT : public CTxOutBase
 {
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
+// Copyright (c) 2018-2019 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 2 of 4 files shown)


CaveSpectre11:
Getting asserts with coin sends, maybe it was short cutting the coin selection to availablecoins.  Will investigate (and change to WIP when i get back to my computer)

CaveSpectre11:
> Currently seeing asserts in `wallet/coinselection.h:28` when sending basecoin to ringct. Perhaps there is code that is assuming `AvailableCoins` returns only basecoin. Will investigate; but in the meantime, "code review" tag is on for the code that's been updated thus far.

This problem has been corrected with https://github.com/Veil-Project/veil/pull/696/commits/baeab980b771c6ecad45b7151e636c04e94b8212 .  Given the distinct delineation between the two commits, I would like to keep them separate rather than rebase them together; as some development may need to cherry pick the latter commit, but not necessarily need the former commit.

codeofalltrades:
@mimirmim Please review. If ok, it is ready for QA.

CaveSpectre11:
@mimirmim btw: this is backwards compatible and doesn't have to wait for the v2.0 branch in it's current state.  Actually might be a good idea to get it rolling so if we do need some specific v2.0 changes, they can be done on top of this.

Issue #710, opened 2019-12-16
qt: Set AA_EnableHighDpiScaling attribute early


sinetek:
`Qt docs: This attribute must be set before QGuiApplication is 
constructed.`
Fixes an issue with Qt.

sinetek:
This should have been a pull request. NVM.

sinetek:
Reopening until it's merged.

PR #711, opened 2019-12-16
[Qt] AADPI fix - Enables high-DPI scaling in Qt on supported platforms

sinetek:
This takes care of a Qt issue on more recent versions.

fixes: #710

### DIFF
--- a/src/qt/veil.cpp
+++ b/src/qt/veil.cpp
@@ -590,13 +590,12 @@ int main(int argc, char *argv[])
     Q_INIT_RESOURCE(veil);
     Q_INIT_RESOURCE(veil_locale);
 
-    BitcoinApplication app(*node, argc, argv);
 #if QT_VERSION > 0x050100
     // Generate high-dpi pixmaps
     QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
 #endif
 #if QT_VERSION >= 0x050600
-    QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+    QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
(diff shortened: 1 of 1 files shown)


CaveSpectre11:
@sinetek can you describe what the symptom of this problem was?

sinetek:
> @sinetek can you describe what the symptom of this problem was?

See bitcoin issue #16254.

src/qt/veil.cpp
```diff
@@ -590,13 +590,12 @@ int main(int argc, char *argv[])
     Q_INIT_RESOURCE(veil);
     Q_INIT_RESOURCE(veil_locale);
 
-    BitcoinApplication app(*node, argc, argv);
 #if QT_VERSION > 0x050100
     // Generate high-dpi pixmaps
     QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
 #endif
 #if QT_VERSION >= 0x050600
```
CaveSpectre11: Not enough to hold it up, but I would prefer if we start using the syntax `QT_VERSION_CHECK(5,6,0)`, especially considering QT_VERSION is actually built from a QT_VERSION_CHECK call.

crypt0-fusi0n:
@mimirmim Saw this one with QA: Pending... do you mean it's ready for QA?

src/qt/veil.cpp
```diff
@@ -609,6 +608,9 @@ int main(int argc, char *argv[])
     QSslConfiguration::setDefaultConfiguration(sslconf);
 #endif
 
+    BitcoinApplication app(*node, argc, argv);
+
```
MatWaller: Remove extra white space here, other than that all good.

sinetek: done ✍✍

PR #713, opened 2019-12-17
[Qt][Build] Fix Mint Zerocoin amount

CaveSpectre11:
### Problem
Build generates overflow warning in `SettingsMinting`
```
qt/veil/settings/settingsminting.cpp: In constructor ‘SettingsMinting::SettingsMinting(QWidget*, WalletView*, WalletModel*)’:
qt/veil/settings/settingsminting.cpp:77:73: warning: overflow in implicit constant conversion [-Woverflow]
     ui->editAmount->setValidator(new QIntValidator(0, 100000000000, this) );
                                                                         ^
```

### Root Cause
The validator value (100000000000) is greater than an integer; so it only allows an integer value to be placed.

### Solution
The simple solution was to reduce the integer value to something that fits into a 32 bit integer; but that will also allow a negative number to be input.  Instead, this fix sets it to a regular expression that will require a positive digit followed by 1 to 11 other digits (allowing for the size that was allowed before).  Since it technically should only be a multiple of 10, two digits are required and the edit window was changed to mention it needs to be a multiple of 10.  This way they don't enter a number and then are warned that it's not a multiple of 10.

Settings->Zerocoin Minting
image

While there's a lot of other work that could be done on the page; the intent was to remove the warning on the build, so other things on the page were skipped as the page is now only temporary.

### DIFF
--- a/src/qt/veil/settings/settingsminting.cpp
+++ b/src/qt/veil/settings/settingsminting.cpp
@@ -7,7 +7,7 @@
 #include <qt/walletmodel.h>
 #include <qt/guiutil.h>
 #include <qt/optionsmodel.h>
-#include <QIntValidator>
+#include <QRegExpValidator>
 
 #include <QSettings>
 #include <QStandardPaths>
@@ -29,7 +29,7 @@ SettingsMinting::SettingsMinting(QWidget *parent, WalletView *mainWindow, Wallet
 
     ui->btnEsc->setProperty("cssClass" , "btn-text-primary-inactive");
 
-    ui->editAmount->setPlaceholderText("Enter amount here");
+    ui->editAmount->setPlaceholderText("Enter amount here (multiple of 10)");
     ui->editAmount->setAttribute(Qt::WA_MacShowFocusRect, 0);
     ui->editAmount->setProperty("cssClass" , "edit-primary");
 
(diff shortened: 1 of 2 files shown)


sinetek:
utACK

src/qt/veil/settings/settingsminting.cpp
```diff
@@ -74,7 +74,10 @@ SettingsMinting::SettingsMinting(QWidget *parent, WalletView *mainWindow, Wallet
     //
```
MatWaller: Can you remove the empty comment?

sinetek:
can you open an issue instead if you care? theres more whitespace to fix etc and it wastes time to do small edits like that

CaveSpectre11:
> can you open an issue instead if you care? theres more whitespace to fix etc and it wastes time to do small edits like that

It's definitely not worth an extra issue; and it's out of scope of the change (since it wasn't introduced), but it's pretty simple to tweak and rebase and work towards getting the code clean.

crypt0-fusi0n:
* Confirmed build warning is no longer there - Pass
* Minor issue with the UI... - Needs fixing

**When my basecoin balance is greater than 10.1 - UI text shows fine.**

image

**When my basecoin balance is less than 10.1 - UI text is cut off**

image

CaveSpectre11:
> * Minor issue with the UI... - Needs fixing

as I stated in the original PR writeup:

> While there's a lot of other work that could be done on the page; the intent was to remove the warning on the build, so other things on the page were skipped as the page is now only temporary.

This is one of the things I was referring to.  While I would love to take the time to fixup that whole page, the intent with this one is to pretty much clear up the build warning.  So it's my opinion that this request, while valid, is out of scope for this PR and should be put in as an issue (if we care to fix it before zerocoin is phased out)

What do you think @crypt0-fusi0n?

crypt0-fusi0n:
That's fine. I logged a new issue #738 for the UI change..

Can we get a bounty dev on it before next public release?

Issue #709, opened 2019-12-16
Software Versioning issue in Releases.


sinetek:
This one is trivial.
We might want to rename the releases to specify the build version.

![image](https://user-images.githubusercontent.com/5908875/70923811-cb4fb280-2028-11ea-8a6b-a63123f6c708.png)

1.0.4 -> 1.0.4.7

codeofalltrades:
Agreed. Gitian it the problem from my understanding.

mimirmim:
The problem is that we have x.x.x.x, not x.x.x, this was discussed in dev chat already where we remarked that we will be going to sane semantic versioning from 2.0.

PR #715, opened 2019-12-17
[Build] Fix automake's $VERSION var

sinetek:
* For the Veil project we use a different versioning scheme, xxxx. Declare this for correct release naming.

* This behaviour might change in the future.

This fixes issue #709.


mimirmim:
Just note, this will probably absolutely change back to x.x.x in the future from 2.0.

PR #657, opened 2019-08-22
Fix formatting of ztracker remove pending

blondfrogs:
The following segfault occurred on a node running veil. This is a potential fix to this segfault

```
[Thread 0x7ffef37fe700 (LWP 36705) exited]
[New Thread 0x7ffef37fe700 (LWP 36914)]
[Thread 0x7ffef37fe700 (LWP 36914) exited]
[New Thread 0x7ffef37fe700 (LWP 37562)]
[Thread 0x7ffef37fe700 (LWP 37562) exited]
[New Thread 0x7ffef37fe700 (LWP 38217)]
[Thread 0x7ffef37fe700 (LWP 38217) exited]
double free or corruption (fasttop)

Thread 26 "veil-httpworker" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fff98ff9700 (LWP 45833)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51        ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
#0  0x00007ffff2f22e97 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff2f24801 in __GI_abort () at abort.c:79
#2  0x00007ffff2f6d897 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff309ab9a "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007ffff2f7490a in malloc_printerr (str=str@entry=0x7ffff309c828 "double free or corruption (fasttop)") at malloc.c:5350
#4  0x00007ffff2f7c004 in _int_free (have_lock=0, p=0x7ffef8155530, av=0x7ffef8000020) at malloc.c:4230
#5  0x00007ffff2f7c004 in __GI___libc_free (mem=0x7ffef8155540)
    at malloc.c:3124
#6  0x000055555597b93b in __gnu_cxx::new_allocator<std::_Rb_tree_node<std::pair<uint256 const, uint256> > >::deallocate(std::_Rb_tree_node<std::pair<uint256 const, uint256> >*, unsigned long) (this=<optimized out>, __p=0x7ffef8155540)
    at /usr/include/c++/7/ext/new_allocator.h:125
[...]

PR #723, opened 2019-12-27
Get rid of precompute.

sinetek:
* Remove various precompute functionality.

* UI: remove precompute related buttons.


mimirmim:
Looks to be failing, did it build for you @sinetek?

sinetek:
It builds for me indeed.

codeofalltrades:
> Looks to be failing, did it build for you @sinetek?

TravisCI builds are passing now.

src/consensus/tx_verify.cpp
```diff
@@ -235,35 +234,16 @@ bool CheckZerocoinSpend(const CTransaction& tx, CValidationState& state)
...
-                cacheValidatedPubcoin.set(hashPubcoin.GetHex(), true);
-            }
-        }
-
-
-    }
+    if (!pubCoin.validate())
+        return state.DoS(100, error("CheckZerocoinMint() : PubCoin does not validate"));
 
+    bnValue = pubCoin.getValue();
```
CaveSpectre11: This does change a behavior.... before: regardless of if pubCoin validated or not, bnValue is set.  Now, if the validation fails, bnValue remains unchanged.  But we need to make absolutely sure that there isn't any code that uses bnPubCoin this filters back up unchanged.

We should document the routine to explain the conditions bnValue is set, and the conditions bnValue is not set.

The change appears to be fine (initial inspection seems that CheckZerocoinMint() is only called from CheckStandardOutput, which is only called from CheckTransaction() which returns false in CheckStandardOutput fails....   so it -should- be okay.  But that due diligence is needed (both in the details of this PR, and the addition of a little documentation on this routine to explain the conditions that change the arguments.

src/veil/zerocoin/ztracker.h
```diff
@@ -55,13 +53,6 @@ class CzTracker
     bool UpdateZerocoinMint(const CZerocoinMint& mint);
     bool UpdateState(const CMintMeta& meta);
     void Clear();
-    mutable CCriticalSection cs_modify_lock;
```
CaveSpectre11: when resolving the conflict with ztracker.h; note that the conflict is due to https://github.com/Veil-Project/veil/commit/0a0a1159dd596e2337c62f2c74194bd59ac7d527 adding the `cs_remove_pending` critical section.  so that needs to be preserved here.

src/consensus/tx_verify.cpp
```diff
@@ -235,35 +234,16 @@ bool CheckZerocoinSpend(const CTransaction& tx, CValidationState& state)
...
         return state.DoS(100, error("CheckZerocoinMint(): TxOutToPublicCoin() failed"));
 
-    bnValue = pubCoin.getValue();
-    uint256 hashPubcoin = GetPubCoinHash(bnValue);
-
-    //CheckZerocoinMint can be done from different threads at the same time. Lock static containers. If try_lock fails,
-    //don't wait, just do full validation.
-    TRY_LOCK(cs_pubcoinCache, fLocked);
-    bool value;
-    if (!fSkipZerocoinMintIsPrime) {
```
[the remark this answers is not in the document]
sinetek: OK agreed it can go back in. let me see what i can do

src/veil/zerocoin/ztracker.h
sinetek: I'd rather we move cs_remove_pending to just above UpdateStatusInternal  since it is used internally there.
Arguments are: it is polluting the class public space and this is meant for local usage.
Can I?

sinetek: on second thought let's keep the remove lock in the declaration but move it to the private area.
Good?

src/consensus/tx_verify.cpp
sinetek: I agree.. I reverted the function back to the current logic and propose we go ahead with this.

src/veil/zerocoin/ztracker.h
CaveSpectre11: Thank you for asking.  Excellent question.  And if this was in a code review for #657 I would welcome the idea.

However this is just making sure the conflict is resolved.   As such, we really need to keep this PR focused to what this PR is for, and we can follow up with a refactoring PR that addresses the pollution of the class public space (as I'm sure there are other places were improvements under that title can be made).

So for now, merge where #657 put it; and if you want to throw in an issue for refactoring public/private space and reference it.. by all means do so.  Then we can look at that then.

sinetek:
Addressed the collision in ztracker.h.
Addressed the changes in tx_verify

PR #727, opened 2020-01-02
[Chain][Performance] Rework orphan pruning

CaveSpectre11:
### Problem
The PoS algorithm has a much greater potential to have several stakers submitting a valid block, even though eventually only one will end up being chosen by the consensus to be the official chain.  All others that submit a block that doesn't end up in the chain will see orphans.  These orphans reside in the memory of the core daemons until they are purged.  As a result, the number of tips (as seen with `veil-cli getchaintips`) will continue to grow until it reaches a threshold of 1000.

The code searches for stale tips frequently, and has an observed cost of about 1.3 seconds at the current block height for each time the search is done.  This would have grown slowly as the chain grows.

### Root Cause
Code was added to Veil to clean up these tips.  Other projects have many tips in their database (one popular coin was observed to have over 16k.  Veil cleaned the tips up when they reached 1000.  This is believed to be wholly inefficient.

Additionally, this added code has efficiency problems.  One change in this PR is an initial pass at an efficiency change to help reduce the burden of this cleanup.

### Solution
Two changes are included in this PR.  The first, which is easily confirmed, is the pruning of these chain tips.  Instead of the pruning occurring when the number of tips exceeds 1000; it now will prune them when the tip has been stale for more than 12 blocks.

Additionally, the order of the tests was changed.  The bulk of the time goes to the checking of the ancestors of a block.   If that block is in the current chain, the ancestor check is unnecessary.  So instead of checking the ancestors for every block in the database, then checking if it's in the current chain (and pruning it if neither occurs), the bulk of the blocks are in the chain, so we will rule that out first (much faster) and only check ancestry if the block is not in the current chain.

### Testing
Testing the first part is very simple.  After running the wallet for a little while, anyone can see numerous tips; many of them dozens, if not hundreds or thousands of blocks behind the current height when executing `getchaintips`.   These are not necessary to be retained.

With this PR, you will not see any chain tips with a height less than 12 from the current height.  You will see the pruning occurring in the debug log file as well:

```
2020-01-02T02:15:58Z CleanBlockIndexGarbage: Erasing 53584 of 53585 irrelevant indexes
2020-01-02T02:16:12Z CleanBlockIndexGarbage: Erasing 0 of 2 irrelevant indexes
2020-01-02T02:16:22Z CleanBlockIndexGarbage: Erasing 1 of 1 irrelevant indexes
2020-01-02T02:18:29Z CleanBlockIndexGarbage: Erasing 0 of 1 irrelevant indexes
2020-01-02T02:18:30Z CleanBlockIndexGarbage: Erasing 0 of 1 irrelevant indexes
2020-01-02T02:18:49Z CleanBlockIndexGarbage: Erasing 0 of 1 irrelevant indexes
```

(side note; why there are 53k sitting in the block database on startup is unknown currently, and will be investigated at a later date; as they are pruned on startup it doesn't have an adverse effect on the running system.  This is the reason that the number of chain tips starts at zero; because the large amount in the block database triggers them to be pruned on startup).

What the log shows is how many tips there were, and how many were pruned.  When you see "Erasing 0 of 2", that means there are two that are believed to be irrelevant, but their height requirement is not to a point where they will be pruned.

The second issue is not something that can be easily tested without adding timing into the code.  putting timers in the routine at hand, and keeping track of the amount of time the routine took, reveals the effect of that code change (about a 92% performance improvement on that section of code, and shaving off about 1.2 seconds of time in a critical section locked by `cs_main` for each block processed)

# Before
```
2020-01-02T01:45:34Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.632087 seconds
2020-01-02T01:46:47Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.253449 seconds
2020-01-02T01:46:50Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.251874 seconds
2020-01-02T01:47:33Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.292188 seconds
2020-01-02T01:47:45Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.321275 seconds
2020-01-02T01:47:51Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.293556 seconds
2020-01-02T01:47:54Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.286087 seconds
2020-01-02T01:48:01Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.360860 seconds
2020-01-02T01:48:48Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.303014 seconds
2020-01-02T01:49:12Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.243704 seconds
2020-01-02T01:49:18Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.251913 seconds
2020-01-02T01:49:33Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.242712 seconds
2020-01-02T01:49:41Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.284505 seconds
2020-01-02T01:49:50Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.301958 seconds
2020-01-02T01:50:10Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.318217 seconds
2020-01-02T01:51:00Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.304591 seconds
2020-01-02T01:51:17Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.272602 seconds
2020-01-02T01:51:34Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.257356 seconds
2020-01-02T01:51:59Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.284783 seconds
2020-01-02T01:52:02Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.265133 seconds
2020-01-02T01:53:00Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.300298 seconds
2020-01-02T01:53:28Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.248319 seconds
2020-01-02T01:54:19Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.294210 seconds
2020-01-02T01:55:28Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.302487 seconds
2020-01-02T01:56:22Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.255330 seconds
2020-01-02T01:57:22Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.268904 seconds
2020-01-02T01:58:11Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.293706 seconds
2020-01-02T01:58:39Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.354431 seconds
2020-01-02T01:58:56Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.291397 seconds
2020-01-02T01:59:03Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.353625 seconds
2020-01-02T02:00:59Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.309790 seconds
2020-01-02T02:01:12Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.246629 seconds
2020-01-02T02:01:15Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.350599 seconds
2020-01-02T02:01:28Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.303686 seconds
2020-01-02T02:01:40Z (debug) validation.cpp:4731 - CleanBlockIndexGarbage took 1.303095 seconds
```

# After
```
2020-01-02T03:28:17Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099894 seconds
2020-01-02T03:28:18Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099980 seconds
2020-01-02T03:30:10Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099467 seconds
2020-01-02T03:31:27Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100021 seconds
2020-01-02T03:31:32Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099398 seconds
2020-01-02T03:31:38Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100509 seconds
2020-01-02T03:32:52Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.098997 seconds
2020-01-02T03:35:15Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.098915 seconds
2020-01-02T03:35:34Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099278 seconds
2020-01-02T03:36:09Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099287 seconds
2020-01-02T03:36:38Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099494 seconds
2020-01-02T03:37:43Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099135 seconds
2020-01-02T03:37:50Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.138882 seconds
2020-01-02T03:39:11Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100209 seconds
2020-01-02T03:39:22Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.125681 seconds
2020-01-02T03:39:23Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099585 seconds
2020-01-02T03:39:33Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100035 seconds
2020-01-02T03:39:48Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099417 seconds
2020-01-02T03:40:47Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100063 seconds
2020-01-02T03:40:47Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099346 seconds
2020-01-02T03:41:16Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.105898 seconds
2020-01-02T03:41:18Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099293 seconds
2020-01-02T03:42:19Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.101253 seconds
2020-01-02T03:42:20Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.138483 seconds
2020-01-02T03:42:24Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100591 seconds
2020-01-02T03:42:34Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100140 seconds
2020-01-02T03:43:10Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.138943 seconds
2020-01-02T03:43:45Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100185 seconds
2020-01-02T03:43:53Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.098239 seconds
2020-01-02T03:44:55Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100541 seconds
2020-01-02T03:46:53Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099540 seconds
2020-01-02T03:46:59Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100139 seconds
2020-01-02T03:47:04Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100110 seconds
2020-01-02T03:47:07Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099737 seconds
2020-01-02T03:47:20Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.098150 seconds
2020-01-02T03:47:24Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099493 seconds
2020-01-02T03:47:37Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100032 seconds
2020-01-02T03:48:08Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100015 seconds
2020-01-02T03:48:26Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099422 seconds
2020-01-02T03:48:40Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099060 seconds
2020-01-02T03:48:49Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.098883 seconds
2020-01-02T03:48:50Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.099002 seconds
2020-01-02T03:48:55Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.098913 seconds
2020-01-02T03:49:13Z (debug) validation.cpp:4751 - CleanBlockIndexGarbage took 0.100458 seconds
```

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1,7 +1,7 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
 // Copyright (c) 2015-2019 The PIVX developers
-// Copyright (c) 2019 The Veil developers
+// Copyright (c) 2019-2020 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -3542,6 +3542,80 @@ static void NotifyHeaderTip() LOCKS_EXCLUDED(cs_main) {
     }
 }
 
+bool IsAncestor(const CBlockIndex* pindexChain, const CBlockIndex* pindexCheck)
+{
+    if (!pindexChain || !pindexCheck)
+        return false;
+    if (pindexChain->nHeight <= pindexCheck->nHeight)
+        return false;
+
+    return pindexChain->GetAncestor(pindexCheck->nHeight)->GetBlockHash() == pindexCheck->GetBlockHash();
+}
+
+static const unsigned int PRUNE_DEPTH = 120;
+static const unsigned int PRUNE_COUNT = 12;
+
+/*
+** Periodically scan the block index map for stale tips and clean them up.  Note that both
+** chainActive.Contains and IsAncestor are expensive as the block index grows, so we want
+** to take special care to not spend a lot of extra time doing useless work.
+*/
+void PruneStaleBlockIndexes()
+{
+    static int64_t lastPrunedHeight = 0;
+
+    // This isn't going to change while we're processing, so just leave and try again later.
+    if (!pindexBestHeader) return;
+
+    // If mapBlockIndex isn't bloated, don't bother taking the time.
+    if (chainActive.Height() > (mapBlockIndex.size() - PRUNE_COUNT)) {
+        return;
+    }
+
+    LogPrintf("%s: Checking for stale indexes. Block index size=%d; Chain height=%d\n",
+              __func__, mapBlockIndex.size(), chainActive.Height());
+
+    uint32_t irrelevantIndexes = 0;
+    int64_t lowestPrunedHeight = lastPrunedHeight;
+    std::set<uint256> setDelete;
+
+    for (const auto& p : mapBlockIndex) {
+        const CBlockIndex* pindex = p.second;
+        // If we've already checked to this height, don't waste time
+
+        if (pindex->nHeight < lastPrunedHeight)
+            continue;
+
+        // If it's not in the active chain, and not in our best header ancestor list.
+        if (!chainActive.Contains(pindex) && (!IsAncestor(pindexBestHeader, pindex))) {
+            irrelevantIndexes++;
+
+            // if it's also old enough, add it to the prune list.
+            if (pindex->nHeight + PRUNE_DEPTH < chainActive.Height()) {
+                setDelete.emplace(p.first);
+
+                // save the lowest height that we're purging
+                if ((pindex->nHeight < lowestPrunedHeight) || (lowestPrunedHeight <= lastPrunedHeight)) {
+                    lowestPrunedHeight = pindex->nHeight;
+                }
+            }
+        }
+    }
+
+    if (irrelevantIndexes > 0) {
+         LogPrintf("%s: Erasing %d of %d irrelevant indexes.  LastPrunedHeight now %d.\n",
+                   __func__, setDelete.size(), irrelevantIndexes, lowestPrunedHeight);
+
+         // Purge the ones we flagged
+         for (const uint256& hash : setDelete) {
+             mapBlockIndex.erase(hash);
+         }
+
+         // Step back a little for safety
+         lastPrunedHeight = lowestPrunedHeight;
+    }
+}
+
 /**
  * Make the best chain active, in multiple steps. The result is either failure
  * or an activated best chain. pblock is either nullptr or a pointer to a block
@@ -3647,6 +3721,8 @@ bool CChainState::ActivateBestChain(CValidationState &state, const CChainParams&
     if (!FlushStateToDisk(chainparams, state, FlushStateMode::PERIODIC)) {
         return false;
     }
+
+    PruneStaleBlockIndexes();
 
     return true;
 }
@@ -4590,42 +4666,12 @@ bool CChainState::AcceptBlockHeader(const CBlockHeader& block, CValidationState&
     return true;
 }
 
-bool IsAncestor(const CBlockIndex* pindexChain, const CBlockIndex* pindexCheck)
-{
-    if (!pindexChain || !pindexCheck)
-        return false;
-    if (pindexChain->nHeight <= pindexCheck->nHeight)
-        return false;
-
-    return pindexChain->GetAncestor(pindexCheck->nHeight)->GetBlockHash() == pindexCheck->GetBlockHash();
-}
-
-// If there are too many loose block indexes, then delete the extras.
-void CleanBlockIndexGarbage()
-{
-    std::set<uint256> setDelete;
-    for (const auto& p : mapBlockIndex) {
-        const CBlockIndex* pindex = p.second;
-        if (pindexBestHeader && !chainActive.Contains(pindex) && (!IsAncestor(pindexBestHeader, pindex)))
-            setDelete.emplace(p.first);
-    }
-
-    if (setDelete.size() > 1000) {
-        LogPrintf("%s: Erasing %d irrelevant indexes\n", __func__, setDelete.size());
-        for (const uint256& hash : setDelete) {
-            mapBlockIndex.erase(hash);
-        }
-    }
-}
-
 // Exposed wrapper for AcceptBlockHeader
 bool ProcessNewBlockHeaders(const std::vector<CBlockHeader>& headers, CValidationState& state, const CChainParams& chainparams, const CBlockIndex** ppindex, CBlockHeader *first_invalid)
 {
     if (first_invalid != nullptr) first_invalid->SetNull();
     {
         LOCK(cs_main);
-        if (!IsInitialBlockDownload())
-            CleanBlockIndexGarbage();
         int nHeightMaxNonPoW = chainActive.Height() + Params().MaxHeaderRequestWithoutPoW();
         nHeightMaxNonPoW = std::max(nHeightMaxNonPoW, Checkpoints::GetLastCheckpointHeight(chainparams.Checkpoints()));
 


mimirmim:
Going to give this a whirl now and check the improvements over the next couple of days. Thanks for your work on this.

CaveSpectre11:
> * `CleanBlockIndexGarbage` runs more times than it should be. Upwards to 8 times between blocks.

This isn't changed from the original implementation.  It's called from ProcessNewBlockHeaders().  While it makes sense to move it to somewhere perhaps more appropriate, to reduce the number of calls we would have to investigate why ProcessNewBlockHeaders() is being called so much (I suspect it's due to more "solutions" coming in for a block...  An area that needs some focus anyway, to make sure we're not submitting worse solutions then we already have submitted (for instance); as well as potentially looking if there's a way to triage solutions coming in and getting rid of them sooner if we know it won't be the accepted block.

Given the time spent in the routine is significantly less now, I suspect the reason you see it more frequently is due to that effect... you have more time to do other things, like accept more block headers, since you're not blocked up in there as much :).

While I believe there is significantly more to do, I chose to snap the line on this PR due to the immediate improvements it provides.  I'm not sure tackling a frequency reduction is scope that would be required for this to move into the mainline.

> * Since it cleans up to block 16, does that force the chain to be set in stone after 16 blocks?

This is merely cleaning out what it believes to be stale tips.  I believe that the ancestry check will keep them around longer if they're progressing, and if one does surface as the longest chain, we'll pull the blocks (we deleted) again.   A little bit about the original logic; it pruned when 1000 'irrelevant' blocks were found; regardless of how old they are.  In other words; when that 1000th is determined, regardless if it's one back, or an uncle to the current tip you have, it's going to throw it away.  So basing the purge on height rather than the number of them is (in my opinion) infinitely better.

I'm not sure the best way to test this, and confirm if 12 blocks is the right number.  I'm considering splitting the two, since there will still be significant improvement if we're not going through the ancestry check on over 1/2 million blocks in the current chain; the ancestry check on 2 vs. 1000 shouldn't be very significant from a percentage of the total in mapBlockIndex.

> * It seems to consider the tip to be an "irrelevant index", then deletes it when a new block is produced.

The original code considered any that aren't in the current chain (or passing the ancestor check) as "irrelevant".  The change now is that we wait 12 blocks before we remove it; it will keep coming up as irrelevant on each check until either it becomes not irrelevant, or it becomes part of the active chain.  Watching the debug message reveals that this is the case... sometimes 0 of 2 are removed, and then later 1 of 1 is removed... implying the other one became part of the active chain.

> However, it seems like the bottleneck is now fixed and the RPC calls are super fast again.

This specifically is why I don't want to make riskier changes in this PR, and leave it as is.  However, I just split out the performance change into another PR; since I suspect that is the bulk of the bottleneck (although having upwards of 1000 stale tips may be the actual root cause of the bottleneck, the amount of time locked in that check definitely would have some effect).

#728 should be tested to see how much of an improvement that adds without the question of "is 12 too small of a chain deviation to trim" holding up the merge.

CaveSpectre11:
I did a lot of rework on this:
1) Moved it from the headers to when we actually attach a new block; which should reduce the frequency of checks.
2) Greatly reworked the conditions for checking.  It now checks when the mapBlockIndex is "bloated".  It's considered bloated when there's 12 or more extra blocks in the map.
3) Reworked the depth of pruning from 12 to 120.  So if it's more than 120 blocks stale, it will prune it then.
4) Added more conditions to what we'll check.  If it's deep enough that we've checked it before, don't bother since it's likely part of the active chain.
5) We set "deep enough" to be the depth (120) below the deepest one we pruned on the last pass.  So the first pass will take a little while (~.5 seconds right now).  The second pass will check fairly deep as well, but most of those should already be cleaned.  Passes after the second pass should be much faster.

mimirmim:
Testing it out now. Code looks fine, but, I for some reason feel like this may have unforeseen consequences. Give me a day for running this.

src/validation.cpp
```diff
@@ -3542,6 +3542,80 @@ static void NotifyHeaderTip() LOCKS_EXCLUDED(cs_main) {
...
+{
+    if (!pindexChain || !pindexCheck)
+        return false;
+    if (pindexChain->nHeight <= pindexCheck->nHeight)
+        return false;
+
+    return pindexChain->GetAncestor(pindexCheck->nHeight)->GetBlockHash() == pindexCheck->GetBlockHash();
+}
+
+static const unsigned int PRUNE_DEPTH = 120;
```
CaveSpectre11: @mimirmim wondering if what I just found will address your underlying concern.  
we have `DEFAULT_MAX_REORG_DEPTH = 100`; and is settable with -maxreorg.  So I'm going to look at removing `PRUNE_DEPTH` and using nMaxReorgDepth in it's place.

sinetek:
Maybe we should set an upper limit on the number of checks that are performed so as to remain deterministic wrt/ runtime.

|     for (const auto& p : mapBlockIndex) { 

set an upper limit to the number of items to consider.
The idea is to clean out over time, it needn't be done in one shot necessarily.

sinetek:
Maybe we should set an upper limit on the number of checks that are performed so as to remain deterministic (and bounded) wrt/ runtime.

| for (const auto& p : mapBlockIndex) {

set an upper limit to the number of items to consider at any one time.
The idea is to clean out over time, it needn't be done in one shot necessarily.

CaveSpectre11:
> Maybe we should set an upper limit on the number of checks that are performed so as to remain deterministic (and bounded) wrt/ runtime.

The trouble is finding the stale ones.  Reducing the frequency of the passes is what the PRUNE_COUNT is meant to do:

>     // If mapBlockIndex isn't bloated, don't bother taking the time.
>     if (chainActive.Height() > (mapBlockIndex.size() - PRUNE_COUNT)) {
>         return;
>     }

The trouble is the biggest one is actually on startup; there's somewhere on the order of 54,000 stale tips on our disks.  At some point we will figure out how to prune the on disk data so we can clean out the snapshots, but for now that first pass at startup cleans out a significant number.

After that point, the bulk of the time is processing the loop looking for the stale tips.  This (currently) takes 90-120ms (on my machine, with our current 520k blocks).  This of course will grow over time, but I'm predicting 100ms a year give or take. 

So we're at 90-120ms per block received (and linearly growing), only when we're bloated enough, is exponentially better than 1.2-1.4 seconds per header received.

I did take some time to look at where in the map the stale indexes were (after the first cleanup).  I know some places in the code sort the map, so if it was sorted, there was the opportunity to run it until we hit the point where we've already cleared out; knowing the rest of the blocks are in our chain.  However their locations are random; and sorting would not be efficient.

In a nutshell; it's now to a point where the bulk of the time is spent finding the stale blocks.

src/validation.cpp
mimirmim: Yeah this is great. I am not seeing issues of unknown block headers of peers anymore.

src/validation.cpp
```diff
@@ -3542,6 +3542,80 @@ static void NotifyHeaderTip() LOCKS_EXCLUDED(cs_main) {
...
+                // save the lowest height that we're purging
+                if ((pindex->nHeight < lowestPrunedHeight) || (lowestPrunedHeight <= lastPrunedHeight)) {
+                    lowestPrunedHeight = pindex->nHeight;
+                }
+            }
+        }
+    }
+
+    if (irrelevantIndexes > 0) {
+         LogPrintf("%s: Erasing %d of %d irrelevant indexes.  LastPrunedHeight now %d.\n",
```
mimirmim: Seeing this in my debug log, I'm almost thinking it is better not to have it on the regular debug level, but some other. I just want the debug log a bit more cleaned up and seeing this spammed per block can get a bit annoying.

src/validation.cpp
```diff
@@ -3542,6 +3542,80 @@ static void NotifyHeaderTip() LOCKS_EXCLUDED(cs_main) {
...
+
+    // This isn't going to change while we're processing, so just leave and try again later.
+    if (!pindexBestHeader) return;
+
+    // If mapBlockIndex isn't bloated, don't bother taking the time.
+    if (chainActive.Height() > (mapBlockIndex.size() - PRUNE_COUNT)) {
+        return;
+    }
+
+    LogPrintf("%s: Checking for stale indexes. Block index size=%d; Chain height=%d\n",
```
mimirmim: Seeing this in my debug log, I'm almost thinking it is better not to have it on the regular debug level, but some other. I just want the debug log a bit more cleaned up and seeing this spammed per block can get a bit annoying.

src/validation.cpp
CaveSpectre11: At this point I'm not sure I want to suppress the logging.  It doesn't start logging until the bloat exceeds the threshold to start looking for ones to prune.  The reason I don't want to remove the logging yet?  Because this PR is a symptom repair, not root cause.  We still have efficiency issues with the network itself, and therefore still far too many side chains competing.  When those issues start getting unraveled; the number of tips will begin to reduce and we will be able to monitor that progress somewhat by watching the frequency of this log message.

Quite simply; there is going to be PRs coming where I ask you for a log file that includes these two lines before the change, and after the change, to help measure the improvement.

Do you disagree or concur?

codeofalltrades: I would rather over log a bit at this point.

ncrypter90:
@codeofalltrades Could you let me know how can i test the in the wallet?

i have the wallet fully synced with 727 binary. Monitored the debug log with wallet unlocked, but did not find anything related to the fix.

crypt0-fusi0n:
@ncrypter90 Have you observed what has been described in the testing info above from @CaveSpectre11 ?

image

ncrypter90:
@crypt0-fusi0n Yeah that's what i have tested but i dont see those entries in debug.log so i wanted to check what am i missing.

crypt0-fusi0n:
Do you see **CleanBlockIndexGarbage** in the log at all? How long has the wallet been running?

ncrypter90:
@crypt0-fusi0n 

I had started wallet two weeks back, however, Wallet was not running continuously since then, whenever I test I will re-sync and check.

Do i have to run any commands in console to trigger anything?

image

CaveSpectre11:
> @crypt0-fusi0n
> 
> I had started wallet two weeks back, however, Wallet was not running continuously since then, whenever I test I will re-sync and check.
> 
> Do i have to run any commands in console to trigger anything?
> 
> image

The routine name was changed between the time I ran the test logs and the time the code was finally committed.  Instead of searching for the routinename; look for the actual log messages; which are very prevalent in the screenshot you provided.

CaveSpectre11:
Per @seanPhill : ACK https://github.com/Veil-Project/veil/pull/727/commits/82546bd0621d55825a296dcbfbb6d75374cb4a66 - Ready to Merge

PR #729, opened 2020-01-05
[Refactor] Replace boost::bind with std::bind and others

sinetek:
* See BTC cb53b825c26af6e628ba88d72b2000e75bedbbc6

*       Replace boost::bind with std::bind

      - In `src/rpc/server.cpp`, replace `std::transform` with simple loop.
      - In `src/validation.cpp`, store the `boost::signals2::connection` object and use it to disconnect.
      - In `src/validationinterface.cpp`, use 2 map to store the `boost::signals2::scoped_connection` object.

* Get rid of a leak in threaded batch verify (zchain.cpp).

---

### Problem ###
This is a maintenance patch; now that std::bind is available in the standard, we want to
use this instead of boost. That way we avoid potential issues with third party code.
Furthermore, this patch fixes a memory leak in boost::thread_group usage. We won't touch boost::thread_group for now for the sake of small incremental updates.

### Root Cause ###
The C++ standard keeps evolving. This is part of our move to C++11.

### Solution ###

* For boost::bind, → std::bind.
* For the memory leak, use an object on the stack instead of the heap.

### Unit Testing Results ###
Testing on all platforms. 

1)  The build is successful.
3)  This patch touches the core signaling primitives: test that blocks are syncing correctly, that the sync progress bar works, etc.

### DIFF
--- a/src/bench/block_assemble.cpp
+++ b/src/bench/block_assemble.cpp
@@ -76,7 +76,7 @@ static void AssembleBlock(benchmark::State& state)
         ::pcoinsTip.reset(new CCoinsViewCache(pcoinsdbview.get()));
 
         const CChainParams& chainparams = Params();
-        thread_group.create_thread(boost::bind(&CScheduler::serviceQueue, &scheduler));
+        thread_group.create_thread(std::bind(&CScheduler::serviceQueue, &scheduler));
         GetMainSignals().RegisterBackgroundSignalScheduler(scheduler);
         LoadGenesisBlock(chainparams);
         CValidationState state;
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -72,7 +72,6 @@
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/algorithm/string/split.hpp>
 #include <boost/filesystem.hpp>
-#include <boost/bind.hpp>
 #include <boost/interprocess/sync/file_lock.hpp>
 #include <boost/thread.hpp>
 #include <openssl/crypto.h>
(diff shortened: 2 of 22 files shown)


codeofalltrades:
The first step of the travis ci build is failing.
`$ set -o errexit; source .travis/lint_05_before_script.sh
$ set -o errexit; source .travis/lint_06_script.sh
Error: missing script for: a33cc7fdb5504b3c34cf84d27d90476b6a438434
Failed`

sinetek:
> The first step of the travis ci build is failing.
> `$ set -o errexit; source .travis/lint_05_before_script.sh $ set -o errexit; source .travis/lint_06_script.sh Error: missing script for: a33cc7fdb5504b3c34cf84d27d90476b6a438434 Failed`

I have no idea, I didn't cause it?

sinetek:
okay, i didn't know that travis does any verification for scripted diffs.

CaveSpectre11:
@sinetek what I would really like to see is if these types of commits can be cherry-picked from the bitcoin path; so we have the bitcoin commit reference built in; that way the reference isn't lost, or the commit comments.  For ones that are cherry-picked, they are fine as is... if we're making modifications, or even merging by hand, we really should update the copyright if there isn't a veil one in there.

sinetek:
Done rebase.

sinetek:
I'll merge this if there's no opposition?

codeofalltrades:
> 
> 
> I'll merge this if there's no opposition?

Not ready to merge yet.  We need to have it QA'd first.
Would you please add some instructions for QA testing to the PR description?

sinetek:
Yes I see, I thought that might apply to GUI modifications only.

sinetek:
updated description

PR #730, opened 2020-01-05
GUI: Get rid of some deprecated calls.

sinetek:
* QSslSocket -> QSslConfiguration

* Use std::lower_bound() and std::upper_bound() instead of Qt ones.

* Use std sorts instead of Qt ones.

* TextWidth will use the right Qt function depending on version.

* primaryScreen()->geometry() is now what we use.

* desktop() -> primaryScreen().


sinetek:
Ready for review.

src/qt/paymentserver.cpp
```diff
@@ -648,7 +648,7 @@ void PaymentServer::fetchPaymentACK(WalletModel* walletModel, const SendCoinsRec
         qWarning() << "PaymentServer::fetchPaymentACK: Error getting refund key, refund_to not set";
     }
 
-    int length = payment.ByteSize();
+    int length = payment.ByteSizeLong();
```
CaveSpectre11: This change isn't building... 

> qt/paymentserver.cpp:651:26: error: ‘class payments::Payment’ has no member named ‘ByteSizeLong’; did you mean ‘ByteSize’?
     int length = payment.ByteSizeLong();
                          ^~~~~~~~~~~~
                          ByteSize

sinetek: what version of protobuf do you have?

sinetek: fixed it

src/libzerocoin/bignum.h
```diff
@@ -10,6 +10,7 @@
 #include "veil-config.h"
 #endif
 
+#include <limits.h>
```
CaveSpectre11: This looks to be successful (limits must have been getting pulled in somewhere down the path of compat.h).

Let's put this into a separate PR so we can fast-track it through before anything else.

mimirmim:
Travis looks to be failing.

CaveSpectre11:
> Travis looks to be failing.

It's failing for the timeout (test taking too long).  For some reason my request to requeue the tests aren't taking.  I'm not sure why.

codeofalltrades:
utACK
All travis build have passed

CaveSpectre11:
utACK https://github.com/Veil-Project/veil/pull/730/commits/fec6e0aea2664313a11ef157511a91be3b71a6c9

PR #735, opened 2020-01-10
Address mutex requirements.

sinetek:
* Correct missing mutex locks.

* Correct one path where the main mutex was locked twice.

### DIFF
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -234,12 +234,14 @@ WalletTx MakeWalletTx(CHDWallet& wallet, MapRecords_t::const_iterator irtx)
 }
 */
 //! Construct wallet tx status struct.
-WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx)
+WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
 {
     WalletTxStatus result;
     auto mi = ::mapBlockIndex.find(wtx.hashBlock);
+
     CBlockIndex* block = mi != ::mapBlockIndex.end() ? mi->second : nullptr;
     result.block_height = (block ? block->nHeight : std::numeric_limits<int>::max());
+
     result.blocks_to_maturity = wtx.GetBlocksToMaturity();
     result.depth_in_main_chain = wtx.GetDepthInMainChain();
     result.time_received = wtx.nTimeReceived;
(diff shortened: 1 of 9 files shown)


sinetek:
Draft, but please review the concept and current code.

Note to self: I need to go over tests.

src/interfaces/wallet.cpp
```diff
@@ -236,10 +236,13 @@ WalletTx MakeWalletTx(CHDWallet& wallet, MapRecords_t::const_iterator irtx)
 //! Construct wallet tx status struct.
 WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx)
 {
+    LOCK(cs_main);
```
CaveSpectre11: Both callers are locking cs_main before calling in here.  while it probably doesn't matter since I believe we can recursively take the lock; I don't think it's necessary.

src/interfaces/wallet.cpp
```diff
@@ -254,13 +257,14 @@ WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx)
 
 WalletTxStatus MakeWalletTxStatus(AnonWallet* pAnonWallet, const uint256 &hash, const CTransactionRecord &rtx)
 {
+    LOCK(cs_main);
```
CaveSpectre11: This one doesn't have the caller locking it... however the caller is commented out.  So this routine might not ever be called.  Since this is an anonwallet version, it definitely raises the question as to why it's not used.  This might be a bigger problem then just the lack of the cs_main lock.

The question to investigate is "how does the different coin types function through the code calling these two routines"?

src/interfaces/wallet.cpp
```diff
@@ -275,6 +279,7 @@ WalletTxStatus MakeWalletTxStatus(AnonWallet* pAnonWallet, const uint256 &hash,
 //! Construct wallet TxOut struct.
 WalletTxOut MakeWalletTxOut(CWallet& wallet, const CWalletTx& wtx, int n, int depth)
 {
+    LOCK(cs_main);
```
CaveSpectre11: Likewise, all the callers to this have cs_main locked already as well.

src/wallet/rpczerocoin.cpp
```diff
@@ -1365,6 +1365,7 @@ UniValue showspendcaching(const JSONRPCRequest& request)
     }
 
     auto ztracker = pwallet->GetZTrackerPointer();
+    LOCK(ztracker->cs_readlock);
```
CaveSpectre11: What makes this one need the lock, and other callers to ListMints not?

sinetek: only this one calls GetSpendCache() which needs the lock.
but tbh this is going away in another commit anyhow.

sinetek: another PR*

src/interfaces/wallet.cpp
sinetek: that looks right.

sinetek:
I have a better solution for this ..

PR #736, opened 2020-01-11
Refactor + fix mutex locking in wallet and ztracker.

sinetek:
**Problem**
We are having trouble with locking issues.

**Root Cause**
Locking is used to protect information from getting corrupted when multiple threads access it at the same time. The issue is that locking must be done in the correct order, and it's pretty tricky.

**Solution**
I use Clang's static analysis features to find a few places where locking is done twice, incorrectly. I found a few inconsistencies in the locking paths using Clang's tools.
Finally, by perusing the bitcoin sources, I added in correct locking annotations to a bunch of critical wallet functions.
This solution is part of a greater incremental push to fix locking issues. The locking issues are admittedly hairy, this patch is meant as a small improvement over the current situation.

The patch's description is as follows.

* Correct missing mutex locks.

* Correct paths where the mutex is locked twice.

* Add mutex annotations to some declarations, prototypes.

**Unit Testing Results**
Testing on all platforms.

1) The build is successful.
2) The user succeeds in launching the application to the main window.
3) This patch touches a lot of core components like the wallet, and mining, and so on.. For this patch it is requested that tests with transactions be done, what we are looking for is hard crashes, and performance issues.

### DIFF
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -234,12 +234,14 @@ WalletTx MakeWalletTx(CHDWallet& wallet, MapRecords_t::const_iterator irtx)
 }
 */
 //! Construct wallet tx status struct.
-WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx)
+WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
 {
     WalletTxStatus result;
     auto mi = ::mapBlockIndex.find(wtx.hashBlock);
+
     CBlockIndex* block = mi != ::mapBlockIndex.end() ? mi->second : nullptr;
     result.block_height = (block ? block->nHeight : std::numeric_limits<int>::max());
+
     result.blocks_to_maturity = wtx.GetBlocksToMaturity();
     result.depth_in_main_chain = wtx.GetDepthInMainChain();
     result.time_received = wtx.nTimeReceived;
@@ -252,7 +254,7 @@ WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx)
     return result;
 }
 
-WalletTxStatus MakeWalletTxStatus(AnonWallet* pAnonWallet, const uint256 &hash, const CTransactionRecord &rtx)
+WalletTxStatus MakeWalletTxStatus(AnonWallet* pAnonWallet, const uint256 &hash, const CTransactionRecord &rtx) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
 {
     WalletTxStatus result;
     auto mi = ::mapBlockIndex.find(rtx.blockHash);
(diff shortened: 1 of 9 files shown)


sinetek:
OK I'm happy with this now. Up for review.

codeofalltrades:
@sinetek please update the PR description with links to the bitcoin core commits these changes came from.

sinetek:
This patch is a combination of various Bitcoin commits and my own original work using Clang's static analysis tools.
As such, I can't give one single commit for the base work, however one could glean more information about the locking changes in upstream with:

`git log --all --grep="EXCLUSIVE_LOCKS_REQUIRED"`

See also the commit.
`0dec5b5af4e8531bf4d50f74eafe8e5e60d2fbc3`

sinetek:
What would you like to see for the PR message?

codeofalltrades:
Can you please update with the completed PR template?

sinetek:
Updated.

CaveSpectre11:
rebased with current master

src/interfaces/wallet.cpp
```diff
@@ -234,12 +234,14 @@ WalletTx MakeWalletTx(CHDWallet& wallet, MapRecords_t::const_iterator irtx)
 }
 */
 //! Construct wallet tx status struct.
-WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx)
+WalletTxStatus MakeWalletTxStatus(const CWalletTx& wtx) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
```
mimirmim, line 237: Shouldn't we be doing this in all places that need to lock cs_main?

src/wallet/wallet.cpp
```diff
@@ -1448,7 +1448,7 @@ void CWallet::BlockUntilSyncedToCurrentChain() {
         // We could also take cs_wallet here, and call m_last_block_processed
         // protected by cs_wallet instead of cs_main, but as long as we need
         // cs_main here anyway, it's easier to just call it cs_main-protected.
-        LOCK(cs_main);
+        LOCK2(cs_main, cs_wallet);
```
mimirmim, line 1451: With the above comment, is this necessary?

CaveSpectre11: Good catch; yea.  It would take some digging to see if cs_wallet is actually needed, or to do the action called out for.

CaveSpectre11:
Needs a rebase of sorts since ztracker was removed with the precompute removal.

mimirmim:
Says on the check that there are no conflicts with base branch when rebasing?

CaveSpectre11:
> Says on the check that there are no conflicts with base branch when rebasing?

hence the "of sorts".  I rebased it before, but I think the changes to ztracker overrode the deletion of the file (I think); need to look at it.

mimirmim:
I wish we had our benchmarks working, could really get insight if was made slower / faster.

seanPhill:
Bump

PR #739, opened 2020-01-13
[RPC][Logging] First round of RingCT logging cleanup

CaveSpectre11:
### Problem
As a precursor to working through the various strangities with sending ringCT transactions, the error logging needed drastic improvements.  Not the least of these being the plethora of cases where an RPC command returns with:
```
error code: -4
error message:
AddAnonInputs failed: .
```

### Root Cause
The different routines were very confusing.  Some returning a boolean, some returning 0 for good and 1 for bad.  It made it very confusing and caused a number of different methods to be used to return the error message through the logging utility.  All the routines returning 0 for pass only ever returned 1 for fail, so there was no real reason for them to not be changed to bool, and the return codes reversed.  This allowed for the standard `error()` method to be able to be used (which will report the module and message to the debug log and return false).  The other major problem was `sError` was not consistently set, and when it was, it was often overwritten by callers.  Now the original sError will be passed back through the code, as well as the original generator of the error to be reported into the log file.

### Solution
Major cleanup of error messages and returning.  Still more to be done, but this should start a path to get to the bottom of some of the confusing error codes.  This gives us two benefits.  First, the debug message is going to tie back to exactly the routine where the failure occurred (I may even go so far as to add line number into the message as well) in a future PR.  For example, `AddAnonInputs_Inner` was being reported as a failure location rather than the true routine `PickHidingInputs`.  Secondly, the user doesn't need to be bothered with where the problems occurred.  They need to know "why".  Taking off the preambles so they can see precisely what went wrong should be a step in the right direction.

Still need to go through other places; especially `AddStandardInputs` and `AddBlindedInputs` (as well as changing that to `AddSteathInputs`.  But RingCT (obviously) has priority.

```
$ ./veil-cli -regtest sendringcttobasecoin "mqw2957UraAYyPuYNuM8THYiLEZjvpZrbZ" 15000
error code: -4
error message:
Insufficient funds.
```
```
2020-01-13T03:17:41Z ERROR: AddAnonInputs_Inner: Insufficient funds.
```
instead of
```
$ ./veil-cli -regtest sendringcttobasecoin "mqw2957UraAYyPuYNuM8THYiLEZjvpZrbZ" 15000
error code: -4
error message:
AddAnonInputs failed: .
```
```
2020-01-13T03:15:28Z ERROR: AddAnonInputs_Inner: Insufficient funds.
```
and
```
$ ./veil-cli -regtest sendringcttobasecoin "mqw2957UraAYyPuYNuM8THYiLEZjvpZrbZ" 150
error code: -4
error message:
Not enough anonymous outputs exist, last: 132, required: 264.
```
```
2020-01-13T03:18:39Z ERROR: PickHidingOutputs: Not enough anonymous outputs exist, last: 132, required: 264.
```
instead of
```
$ ./veil-cli -regtest sendringcttobasecoin "mqw2957UraAYyPuYNuM8THYiLEZjvpZrbZ" 150
error code: -4
error message:
AddAnonInputs failed: ringctwallet Not enough anon outputs exist, last: 132, required: 297.
```
```
2020-01-13T03:14:01Z ERROR: PickHidingOutputs: ringctwallet Not enough anon outputs exist, last: 132, required: 297
2020-01-13T03:14:01Z ERROR: AddAnonInputs_Inner: failed to pick hiding outputs: ringctwallet Not enough anon outputs exist, last: 132, required: 297
```

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -1305,7 +1305,8 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                 }
 
                 if (k >= nTries) {
-                    return error("%s: Could not generate receiving public key.", __func__);
+                    sError = strprintf("Could not generate receiving public key.");
+                    return error("%s: %s", __func__, sError);
                 }
 
                 CPubKey pkEphem = r.sEphem.GetPubKey();
@@ -1316,13 +1317,16 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                 CTempRecipient rd;
                 rd.nType = OUTPUT_DATA;
 
-                if (0 != MakeStealthData(r.sNarration, sx.prefix, sShared, pkEphem, rd.vData, r.nStealthPrefix, sError)) {
-                    return error("%s: Failed to make stealth data: %s", __func__, sError);
+                if (!MakeStealthData(r.sNarration, sx.prefix, sShared, pkEphem, rd.vData, r.nStealthPrefix, sError)) {
+                    // TODO: MakeStealthData currently won't return failure
+                    return false;
                 }
 
                 if (r.isMine) {
-                    if (!RecordOwnedStealthDestination(sShared, sx.GetID(), idTo))
-                        return error("%s: failed to record stealth destination", __func__);
+                    if (!RecordOwnedStealthDestination(sShared, sx.GetID(), idTo)) {
+                        sError = strprintf("Failed to record stealth destination.");
+                        return error("%s: %s", __func__, sError);
+                    }
                 }
 
                 vecSend.insert(vecSend.begin() + (i+1), rd);
@@ -1336,7 +1340,8 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                     if (!r.fScriptSet) {
                         r.scriptPubKey = GetScriptForDestination(r.address);
                         if (r.scriptPubKey.empty()) {
-                            return wserrorN(1, sError, __func__, "Unknown address type and no script set.");
+                            sError = strprintf("Unknown address type and no script set.");
+                            return error("%s: %s", __func__, sError);
                         }
                     }
                 }
@@ -1362,7 +1367,8 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                 }
 
                 if (k >= nTries) {
-                    return error("%s: Could not generate receiving public key.", __func__);
+                    sError = strprintf("Could not generate receiving public key.");
+                    return error("%s: %s", __func__, sError);
                 }
 
                 r.pkTo = CPubKey(pkSendTo);
@@ -1374,8 +1380,10 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                 }
 
                 if (r.isMine) {
-                    if (!RecordOwnedStealthDestination(sShared, sx.GetID(), idTo))
-                        return error("%s: failed to record stealth destination", __func__);
+                    if (!RecordOwnedStealthDestination(sShared, sx.GetID(), idTo)) {
+                        sError = strprintf("Failed to record stealth destination.");
+                        return error("%s: %s", __func__, sError);
+                    }
                 }
             } else if (r.address.type() == typeid(CExtKeyPair)) {
                 throw std::runtime_error(strprintf("%s: sending to extkeypair", __func__));
@@ -1386,7 +1394,8 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
             } else if (!r.fScriptSet) {
                 r.scriptPubKey = GetScriptForDestination(r.address);
                 if (r.scriptPubKey.size() < 1) {
-                    return error("%s: Unknown address type and no script set.", __func__);
+                    sError = strprintf("Unknown address type and no script set.");
+                    return error("%s: %s", __func__, sError);
                 }
             }
 
@@ -1410,7 +1419,8 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                     sEphem.MakeNewKey(true);
                 }
                 if (k >= nTries) {
-                    return error("%s: Could not generate receiving public key.", __func__);
+                    sError = strprintf("Could not generate receiving public key.");
+                    return error("%s: %s", __func__, sError);
                 }
 
                 r.pkTo = CPubKey(pkSendTo);
@@ -1421,11 +1431,14 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                 }
 
                 if (r.isMine) {
-                    if (!RecordOwnedStealthDestination(sShared, sx.GetID(), idTo))
-                        return error("%s: failed to record stealth destination", __func__);
+                    if (!RecordOwnedStealthDestination(sShared, sx.GetID(), idTo)) {
+                        sError = strprintf("Failed to record stealth destination.");
+                        return error("%s: %s", __func__, sError);
+                    }
                 }
             } else {
-                return error("%s: Only able to send to stealth address for now.", __func__); // TODO: add more types?
+                sError = strprintf("Only able to send to stealth address for now."); // TODO: add more types?
+                return error("%s: %s", __func__, sError);
             }
 
             r.sEphem = sEphem;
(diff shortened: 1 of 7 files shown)


src/veil/ringct/anonwallet.cpp
```diff
@@ -1317,12 +1318,15 @@ bool AnonWallet::ExpandTempRecipients(std::vector<CTempRecipient> &vecSend, std:
                 rd.nType = OUTPUT_DATA;
 
                 if (0 != MakeStealthData(r.sNarration, sx.prefix, sShared, pkEphem, rd.vData, r.nStealthPrefix, sError)) {
```
mimirmim: Change this from `0 != `. Usual Particl weirdness.

CaveSpectre11: ahh yes, of course.  I made the decision to flip them for the ones that I was fixing their sError definitions of; which was after I went through MakeStealthData.  I made sure putting the commented out code back in will actually not break things, but I didn't go through and define sError in those places because that needs more attention anyway.

mimirmim: Eventually should be a goal to revisit and rework much of the RingCT work Particl had done as it isn't Object Orientated, and has repeated code everywhere. No harm is fixing some as it comes :)

src/veil/ringct/anonwallet.cpp
```diff
@@ -3317,8 +3370,10 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
                 nFeeRet = 0;
                 std::vector<uint8_t> &vData = ((CTxOutData*)txNew.vpout[0].get())->vData;
                 vData.resize(1);
-                if (0 != PutVarInt(vData, 0))
-                    return error("%s: failed to add CTFee of 0 to transaction", __func__);
+                if (0 != PutVarInt(vData, 0)) {
```
mimirmim: Another one here.

src/veil/ringct/anonwallet.cpp
```diff
@@ -3437,8 +3501,10 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
         //Add actual fee to CT Fee output
         std::vector<uint8_t> &vData = ((CTxOutData*)txNew.vpout[0].get())->vData;
         vData.resize(1);
-        if (0 != PutVarInt(vData, nFeeRet))
-            return error("%s: PutVarInt %d failed\n", __func__, nFeeRet);
+        if (0 != PutVarInt(vData, nFeeRet)) {
```
mimirmim: And here

src/veil/ringct/anonwallet.cpp
```diff
@@ -3459,17 +3525,24 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
...
+                    if (!GetKey(idk, key)) {
+                        sError = strprintf("No key for output: %s", HexStr(anonOutput.pubkey.begin(), anonOutput.pubkey.end()));
+                        return error("%s: %s", __func__, sError);
+                    }
 
                     // Keyimage is required for the tx hash
-                    if (0 != (rv = secp256k1_get_keyimage(secp256k1_ctx_blind, &vKeyImages[k * 33], anonOutput.pubkey.begin(), key.begin())))
-                        return error("%s: secp256k1_get_keyimage failed %d", __func__, rv);
+                    rv = secp256k1_get_keyimage(secp256k1_ctx_blind, &vKeyImages[k * 33], anonOutput.pubkey.begin(), key.begin());
+                    if (0 != rv) {
```
mimirmim: And here

src/veil/ringct/anonwallet.cpp
```diff
@@ -3526,16 +3602,19 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
...
                             }
                             vpsk[k] = vsk[k].begin();
 
                             vpBlinds.push_back(&vInputBlinds[l][k * 32]);
                             /*
                             // Keyimage is required for the tx hash
-                            if (0 != (rv = secp256k1_get_keyimage(secp256k1_ctx_blind, &vKeyImages[k * 33], &vm[(i+k*nCols)*33], vpsk[k])))
-                                return errorN(1, sError, __func__, _("secp256k1_get_keyimage failed %d").c_str(), rv);
+                            rv = secp256k1_get_keyimage(secp256k1_ctx_blind, &vKeyImages[k * 33], &vm[(i+k*nCols)*33], vpsk[k]);
+                            if (0 != rv) {
```
mimirmim: I guess, might as well change this one too.

src/veil/ringct/anonwallet.cpp
```diff
@@ -3587,34 +3669,39 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
...
-
                     memcpy(&vDL[(1 + (nSigInputs+1) * nSigRingSize) * 32], splitInputCommit.data, 33);
 
                     vpBlinds.emplace_back(vSplitCommitBlindingKeys[l].begin());
                     const uint8_t *pSplitCommit = splitInputCommit.data;
-                    if (0 != (rv = secp256k1_prepare_mlsag(&vm[0], blindSum,
-                        1, 1, nCols, nRows,
-                        &vpInCommits[0], &pSplitCommit, &vpBlinds[0]))) {
-                        return error("%s: secp256k1_prepare_mlsag failed %d", __func__, rv);
+                    if (0 != (rv = secp256k1_prepare_mlsag(&vm[0], blindSum, 1, 1, nCols, nRows,
```
mimirmim: Here

src/veil/ringct/anonwallet.cpp
```diff
@@ -3587,34 +3669,39 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
...
+                                                           &vpInCommits[0], &pSplitCommit, &vpBlinds[0]))) {
+                        sError = strprintf("Failed to prepare mlsag with %d.", rv);
+                        return error("%s: %s", __func__, sError);
                     }
 
                     vpBlinds.pop_back();
                 };
 
                 uint256 hashOutputs = txNew.GetOutputsHash();
                 if (0 != (rv = secp256k1_generate_mlsag(secp256k1_ctx_blind, &vKeyImages[0], &vDL[0], &vDL[32],
```
mimirmim: Here

src/veil/ringct/anonwallet.cpp
```diff
@@ -3587,34 +3669,39 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
...
+                                                        randSeed, hashOutputs.begin(), nCols, nRows, vSecretColumns[l],
+                                                        &vpsk[0], &vm[0]))) {
+                    sError = strprintf("Failed to generate mlsag with %d.", rv);
+                    return error("%s: %s", __func__, sError);
                 }
 
                 // Validate the mlsag
-                if (0 != (rv = secp256k1_verify_mlsag(secp256k1_ctx_blind, hashOutputs.begin(), nCols, nRows, &vm[0], &vKeyImages[0], &vDL[0], &vDL[32])))
-                    return error("%s: secp256k1_verify_mlsag failed on initial generation %d", __func__, rv);
+                if (0 != (rv = secp256k1_verify_mlsag(secp256k1_ctx_blind, hashOutputs.begin(), nCols,
```
mimirmim: Here

src/veil/ringct/anonwallet.cpp
```diff
@@ -3552,9 +3631,10 @@ int AnonWallet::AddAnonInputs_Inner(CWalletTx &wtx, CTransactionRecord &rtx, std
                     vpBlinds.insert(vpBlinds.end(), vpOutBlinds.begin(), vpOutBlinds.end());
 
                     if (0 != (rv = secp256k1_prepare_mlsag(&vm[0], blindSum,
```
mimirmim: Here

src/veil/ringct/stealth.cpp
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2014 The ShadowCoin developers
 // Copyright (c) 2017-2019 The Particl developers
-// Copyright (c) 2018-2019 Veil developers
+// Copyright (c) 2018-2019 The Veil developers
```
mimirmim: This is better.

sinetek:
few things ...

We shouldn't abuse the Trivial tag, this commit changes the logic of error values. a trivial commit would be a name change or a comment change.

Second, let's be consistent. ...  If int errors are returned elsewhere in the code base, we should use that.  I am more in favor of an INT because we can return different error codes, a bool limits you to a not very helpful "fail" status.

sinetek:
> few things ...
> 
> We shouldn't abuse the Trivial tag, this commit changes the logic of error values. a trivial commit would be a name change or a comment change.
> 
> Second, let's be consistent. ... If int errors are returned elsewhere in the code base, we should use that. I am more in favor of an INT because we can return different error codes, a bool limits you to a not very helpful "fail" status.

```
bool DirIsWritable(const fs::path& directory)
{
    fs::path tmpFile = directory / fs::unique_path();

    FILE* file = fsbridge::fopen(tmpFile, "a");
    if (!file) return false;

    fclose(file);
    remove(tmpFile);

    return true;
}
```
Base uses bools, I'm okay with this change.

CaveSpectre11:
> We shouldn't abuse the Trivial tag, this commit changes the logic of error values. a trivial commit would be a name change or a comment change.

I do disagree with this somewhat.  The code flow isn't changing, so it's pretty much a trivial change.  I would be more inline to argue that it's not trivial because error messages are changing (and thus scripts might break, or automated tests).  Given that return values are flipped doesn't necessarily make it non-trivial; but it does put more importance on code review.   So while I disagree, it's a fair point so I'll give you it :)

> Second, let's be consistent. ... If int errors are returned elsewhere in the code base, we should use that. I am more in favor of an INT because we can return different error codes, a bool limits you to a not very helpful "fail" status.

Note a couple things about the ones I changed, and really the 'why'.  The bulk of the changes are in calls that pass in `sError`, which is returning a much more robust error status then return codes.  They are also routines that only return 0 or 1.  Should we feel the need to put in more complex error coding, we can always change them back to `int`, but at this point in time, int is overkill, confusing, and... 

... to the main point; encouraging sError to not be properly loaded in many places.  Given the flow of sError back through the code, you were getting things that were chaining together their pass back through the code.  This cleans it up some to provide groundwork for a separation of what is returned to the user (`sError`), and what is put in the debug log (`error()`).  These will continue to separate as time moves on... `sError` for what the user needs to know (e.g. "Insufficient fee") and what a developer needs to know to assess if it's incorrect.

So while I agree that there is a place for multiple error codes; the current code base isn't using those, and not only they weren't using them, they were either not loading `sError` (like the Insufficient Funds error that was suppressed), or overwriting it to obscure the root of the failure.  I didn't want to creep so much to start writing new tools to load sError and also log.. cleanly.

Hope that helps with the perspective of where this PR is coming from.

CaveSpectre11:
> Sorry, I missed a number of them, forgot to check a diff when reviewing initially.

To summarize our morning/evening discussion:
- Intent here was to lay groundwork for improvements, not to fully solve it all.  It was a speedbump that was impeding investigations into failing ringct spend commands.
- Further improvements will be made in future PRs; most notably where `sError` is used, to make sure those are loaded properly and logged properly (the primary drive down this path).
- with this PR as groundwork, we can have bounty issues to go further if we desire

With that stated; I'm going to mark your feedback resolved and kick it back for another review so it's not sitting with 'changes requested'.

PR #743, opened 2020-01-17
[RPC] Fix sendrawtransaction Dandelion default

CaveSpectre11:
### Problem
The RPC documentation for `sendrawtransaction` shows:

`3. useDandelion     (boolean, optional, default=false) Use dandelion protocol to broadcast transaction
`

However it's actually "true" by default.

### Root Cause
Code review fail.

### Solution
Make the actual default "false".

This is important for now, because there appears to be a problem with a segmentation fault when sending a raw transaction through dandelion.  While dandelion is being investigated, we will correct the default behavior (and since `sendtoaddress` defaults to false correctly; we will go with default false.

In the future, we will work towards changing the default to true; but for now we will correct the default to match the help.


sinetek:
utACK

mimirmim:
ACK

PR #746, opened 2020-01-22
[Build] Increase proc count and timeout length

codeofalltrades:


.travis.yml
```diff
@@ -12,7 +12,7 @@ stages:
   - test
 env:
   global:
-    - MAKEJOBS=-j3
+    - MAKEJOBS=-j8
```
CaveSpectre11: "You see, most blokes will be bulding at 3. You’re on 8, all the way up, all the way up...Where can you go from there? Nowhere. What we do, is if we need that extra push over the cliff...Eleven. One faster."

sinetek: 😂😂

codeofalltrades:
Closing. Issue was fixed with pr #754

PR #749, opened 2020-02-18
Add missing header.

sinetek:
This is required for building with GCC 9.2.

### DIFF
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -14,6 +14,7 @@
 #include <ui_interface.h>
 
 #include <memory>
+#include <deque>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>


codeofalltrades:
utACK. Please look at the Travis build. All are passing except the macOS one.

CaveSpectre11:
> utACK. Please look at the Travis build. All are passing except the macOS one.

Agreed.  I wanted to make sure we didn't need compiler conditional around the include to only include it if using a compiler at or later then the one needed; so travis and gitlian success is important on this one.  I tried to restart the travis test but it didn't.

@sinetek can you update the PR submission content to explain what is needed from that include file for GCC 9.2?

sinetek:
Travis is timing out on the macOS build, the usual.

httpserver is directly using std::deque, so we have to include it.
everything is fine with this patch.

mimirmim:
Yeah, that's bizarre to fail a build when you were just adding a missing header.

crypt0-fusi0n:
This seems to be building fine for me... all Windows, Mac and Linux builds working perfectly.

codeofalltrades:
@crypt0-fusi0n based on your comments this is QA complete?

crypt0-fusi0n:
> @crypt0-fusi0n based on your comments this is QA complete?

Almost... need to just double-check a couple of things and will mark as approved.

PR #751, opened 2020-02-25
Update block header to remove veildatahash

blondfrogs:


src/chain.h
```diff
@@ -600,7 +608,14 @@ class CDiskBlockIndex : public CBlockIndex
         // block header
         READWRITE(this->nVersion);
         READWRITE(hashPrev);
-        READWRITE(hashVeilData);
+        bool fCheckVeilDataHash = false;
+        if (nVersion >> 28 != 3) {
```
CaveSpectre11: Can you explain the logic here; may need more comments.  From what I can tell, current version would match this check (0x20000000), and so I presume new version is 0x30000000.  I would much prefer to see this the VERSIONBITS values and masks.  I also am curious to understand more about what "version bits" are compared to regular "pre-version bits" versions.

src/chainparams.cpp
```diff
@@ -443,6 +442,11 @@ class CTestNetParams : public CChainParams {
         nDefaultPort = 58811;
         nPruneAfterHeight = 1000;
 
+        /** Timestamp when to switch to ProgPow, RandomX, Sha256D. UTC based **/
+        // TODO update when ready for testnet launch
+        nPowUpdateTimestamp = 8578938400; // Monday, January 13, 2020 6:00:00 PM
+        nPowTimeStampActive = 8578938400; // Used by block.h for serialization
```
CaveSpectre11: These need to be reduced to 4294967295 so that we don't actually inadvertently turn it on due to the value exceeding a 32 bit boundary.

src/primitives/block.h
```diff
@@ -99,24 +103,54 @@ class CBlockHeader
...
             READWRITE(mixHash);
         } else if (nVersion & RANDOMX_BLOCK && nTime >= nPowTimeStampActive) {
             READWRITE(nHeight);
             READWRITE(nNonce);
+        } else if (nVersion & SHA256D_BLOCK && nTime >= nPowTimeStampActive) {
+            READWRITE(nNonce64);
         } else {
             READWRITE(nNonce);
         }
+
```
CaveSpectre11: Rather than all these extra checks, this might be better all contained inside an nTime check:
```
if (nTime >= nPowTimeStampActive) {
  if (fCheckVeilDataHash) {
    ... return...
    }
  read merkle root
  read witness
  read accumulators
}
```
Since it's going to be one of the 3, I would think a switch would be better served.  Since you check the bit, what happens if you have a block that set more than one of the bits?  Should you not mask off the the type and switch it?
```
nPowType = (nVersion & (PROGPOW_BLOCK | RANDOMX_BLOCK | SHA256D_BLOCK))
switch (nPowType)
  case PROGPOW_BLOCK
...
  case 0
     READWRITE (nNonce)
  default: 
    // someone's setting more than one!
```
```

File: src/primitives/block.h [hunk 4]
```diff
@@ -99,24 +103,54 @@ class CBlockHeader
...
+            // This is the version that the new PoW started at. This is when we stop using hashVeilData.
+            // We would normally use the nTime of the block. However, that is serialized after the hashVeilData
+            // So we don't have access to it
+            READWRITE(hashVeilData);
+        }
+
         READWRITE(nTime);
         READWRITE(nBits);
+
+        if (fCheckVeilDataHash && nTime >= nPowTimeStampActive) {
```
CaveSpectre11: What about the inverse?  Should we not:
```
if (nTime >= )
else
  if !fCheckVeilDataHash
    // fail to accept premature new PoW block?
```

src/primitives/block.h
```diff
@@ -99,24 +103,54 @@ class CBlockHeader
 
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action) {
+        bool fCheckVeilDataHash = false;
         READWRITE(nVersion);
         READWRITE(hashPrevBlock);
-        READWRITE(hashVeilData);
+        if (nVersion >> 28 != 3) {
```
CaveSpectre11: See above.  Want to make sure we don't re-enable this when the block version goes up.

src/chain.h
```diff
@@ -613,6 +628,12 @@ class CDiskBlockIndex : public CBlockIndex
         READWRITE(vMintDenominationsInBlock);
         READWRITE(fProofOfFullNode);
 
+        if (fCheckVeilDataHash && nTime >= nPowTimeStampActive) {
```
mimirmim: I worry about using nTime instead of a block height. May cause issues? One thing that we are for sure about is that height is consistently incrementing by 1. Time can be give or take. And I believe that our PoS intentionally adds variation to the time when I went through it. Could cause valid blocks to fail?

src/chainparams.cpp
```diff
@@ -255,6 +255,12 @@ class CMainParams : public CChainParams {
         nDefaultPort = 58810;
         nPruneAfterHeight = 100000;
 
+        // TODO update timestamp with mainnet activation time
```
mimirmim: Actually, have we always used timestamps for activation? Block time, or height, is always better.

src/primitives/block.h
```diff
@@ -99,24 +103,54 @@ class CBlockHeader
...
         READWRITE(nTime);
         READWRITE(nBits);
+
+        if (fCheckVeilDataHash && nTime >= nPowTimeStampActive) {
+            // We want to fail to accept this block. setting it NUll should fail all checks done on it
+            SetNull();
+            return;
+        }
+
+        if (nTime >= nPowTimeStampActive) {
```
mimirmim: Same issue outlined above.

src/chain.h
CaveSpectre11: @mimirmim I am referring to:
```
static const int32_t VERSIONBITS_LAST_OLD_BLOCK_VERSION = 4;
/** What bits to set in version for versionbits blocks */
-static const int32_t VERSIONBITS_TOP_BITS = 0x04000000UL;
+static const int32_t VERSIONBITS_TOP_BITS = 0x01000000UL;
+static const int32_t VERSIONBITS_OLD_POW_VERSION = 0x20000000UL;
+static const int32_t VERSIONBITS_NEW_POW_VERSION = 0x30000000UL;
/** What bitmask determines whether versionbits is in use */
-static const int32_t VERSIONBITS_TOP_MASK = 0xFC000000UL;
+static const int32_t VERSIONBITS_TOP_MASK = 0xFF000000UL;
/** Total bits available for versionbits */
-static const int32_t VERSIONBITS_NUM_BITS = 26;
+static const int32_t VERSIONBITS_NUM_BITS = 24;
```

So this code `if (nVersion >> 28 != 3)` appears to be built on the tribal knowledge that comes from versionbits.h.  So I'm raising one core review comment, and one possible issue:

- This logic should be using the defines presented in versionbits.h and not magic numbers.
- Next time nVersion changes, this section of code will then execute.  It is not limited to cover "all the nVersions that will have hashVeilData and not any other future versions".

src/chainparams.cpp
blondfrogs: We use timestamps for Proof of Work changes because at certain times in the code, we don't know the block height. We can use block height for other types of changes though. We need to use values that currently reside inside the CBlockHeader class. Currently the height isn't in the CBlockHeader, so we use the next best thing (nTime)

src/chain.h
blondfrogs: I will update this not to use magic number. 
Also, I will update this to check for version 2. The current version and write it when only it is version 2.

src/primitives/block.h
blondfrogs: Great catch. I am going to add code into the CheckBlockHeader that will enforce only one bit to be present in each block header.

src/chain.h
blondfrogs: No this wont allow for valid block to fail. Time is always changing between blocks. We have a max time drift allowed. So about after 10 minutes after the time that we choose to activate. All blocks must be in ahead of activation time. At time point we will know that all blocks going forward will be mined using the new 3 Pow Algos.

src/chain.h
blondfrogs: I have updated the code to not use magic number. Also, I change the check to look for version less than 2. Which includes the genesis block which has a version of 1. All new version going forward will now be supported. 

It isn't often that you need to change the block verison though. So that is great news.

src/primitives/block.h
blondfrogs: Updated the formating. Also added an additional check that wont allow blocks to have multiple algo bits set.

src/primitives/block.h
blondfrogs: Resolved jjust as above

src/primitives/block.h
blondfrogs: Fixed

src/primitives/block.h
blondfrogs: Format updated

src/chainparams.cpp
blondfrogs: Done

src/versionbits.h
```diff
@@ -20,6 +20,7 @@ static const int32_t VERSIONBITS_TOP_MASK = 0xFF000000UL;
 /** Total bits available for versionbits */
 static const int32_t VERSIONBITS_NUM_BITS = 24;
 
+
```
codeofalltrades: remove extra line.

src/chain.h
```diff
@@ -627,10 +644,30 @@ class CDiskBlockIndex : public CBlockIndex
...
+                    READWRITE(nNonce64);
+                    READWRITE(mixHash);
+                    break;
+                case CBlockHeader::RANDOMX_BLOCK:
+                    break;
+                case CBlockHeader::SHA256D_BLOCK:
+                    READWRITE(nNonce64);
+                    break;
+                default:
+                    break;
```
CaveSpectre11: Should this also `SetNull(); return;` as it would imply they had either none of the bits set, or more than one?

src/chain.h
```diff
@@ -627,10 +644,30 @@ class CDiskBlockIndex : public CBlockIndex
...
+            }
+
+            READWRITE(hashAccumulators);
         }
     }
```
CaveSpectre11: What of the case of an "early" newly mined block?  the case where nTime < PowTimeStamp active, but 
fCheckVeilDataHash != true.

So the above if block:
```
if (nTime >= nPowTimeStampActive) {
}
```
should we not have
```
else {
    // Check for an early new block
    if (!fCheckVeilDataHash)
        SetNull(); return;
}
```

CaveSpectre11:
Still need to sort out the travis issue with the jit compiler and `asm/program_prologue_linux.inc`.

src/chain.h
blondfrogs: we can do that. however because of the consensus rules this can't actually happen. We will add the SetNull() as an extra redundancy though

src/chain.h
blondfrogs: @CaveSpectre11 That shouldn't ever happen. As if it does, it will reject the block because it contains a bad version.

CaveSpectre11:
@blondfrogs looking pretty good.  There's a conflict on block.cpp that needs to be addressed, and I want to take some time to look at the callers into the serialization to make sure I convince myself that consensus won't allow my concerning blocks to get in there.

mimirmim:
Looks good.

blondfrogs:
Fix merge conflict @mimirmim @CaveSpectre11 @codeofalltrades

Issue #697, opened 2019-12-01
Wallet segfaults when clicking Send without inputs


crypt0-fusi0n:
**Describe the issue**
Wallet segfaults if you click on the Send button in send dialog without any inputs.

**What behavior did you expect?**
A prompt asking me to input the address and amount

**What was the actual behavior (provide screenshots if the issue is GUI-related)?**
![segfault](https://user-images.githubusercontent.com/48879483/69910119-2e9ee980-145a-11ea-83be-6ad60dc4a6f7.gif)

**How reliably can you reproduce the issue, what are the steps to do so?**

1. Open wallet
2. Go to Send tab
3. Click **Send** without any inputs

**What version of the Veil client are you using, where did you get it (website, self-compiled, etc)?**
1.0.4.7 (website) and current master (self-compiled)

**This is normally the contents of a `debug.log` or `config.log` file. Raw text or a link to a pastebin type site are preferred.**
Backtrace here [2019.12.01 - Segfault on Send.txt](https://github.com/Veil-Project/veil/files/3907861/2019.12.01.-.Segfault.on.Send.txt)

codeofalltrades:
Similar to #684

sinetek:
Managed to reproduce this on v1.0.4.7. but not on master compiled locally.

I suggest if this is an issue we might have to bump the Qt version in our builds.

sinetek:
v1.0.4.8 also affected.

Issue #684, opened 2019-11-09
Send in GUI with EMPTY address field instant segfault crash


seanPhill:
If you put an amount in the GUI Send page and hit the Send button (without also putting an address in the Address field it instantly crashes.
(I did not notice this before, because 99.99% of the time I use the Debug Console.)

I expected it to check that a required field has been filled (with an appropriate content type) and to move the cursor there, and display a red box around it, with a message saying that "an amount is required".

Instant crash.
[In the Terminal window]
$ ./veil-qt -testnet
PublicCoin::validate value is not prime
Segmentation fault: 11

[In the Apple Problem Report dialog]
Process:               veil-qt [3317]
Path:                  /Applications/Veil-newTestnet4-pr680/*/veil-qt
Identifier:            veil-qt
Version:               0
Code Type:             X86-64 (Native)
Parent Process:        bash [3309]
Responsible:           veil-qt [3317]
User ID:               501

Date/Time:             2019-11-09 17:44:04.465 -0500
OS Version:            Mac OS X 10.13.6 (17G9016)
Report Version:        12
Anonymous UUID:        98807FAC-5A43-94B1-AF1F-9E55DC64D4BA

Time Awake Since Boot: 160000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]

Tested multiple times between the reporting user and mine, resulting in 100% the same result.

A recent master (end user on testnet) and the testnet PR (mine).

Windows 10 (end user) MacOS 10.13.6 (mine) iMac, i7 cpu, 8 GB RAM, HDD.

MatWaller:
Does not occur on Ubuntu

```
Distributor ID:	Ubuntu
Description:	Ubuntu 19.04
Release:	19.04
Codename:	disco
```

Booting into Windows now and checking there.

seanPhill:
Great to hear!

PR #755, opened 2020-03-01
[GUI] Fix sending page crashes

CaveSpectre11:
### Problem
As seen in issue #684 and #697; the GUI crashes when you do things it's not expecting in the send screen.

### Root Cause
The error path in the GUI send screen tried to write the error message to the wrong object; which triggered the crash.  A number of other things were found while it was being investigated; the code to check for no recipients wasn't proper; it would never execute since the previous error catch would catch those as well.  Many things would cause the crash:  no recipient address, no value, a basecoin recipient address...

### Solution
Please note this is not a deep dive into all the problems in the send code; but was a means to enable reproduction of a different problem being investigated.  More work will go into this code when the time comes.  For now, a few improvements:

- Send page now notifies you what is wrong with your send; "Amount can't be zero", "only stealth addresses accepted", etc.
- Send Page now states "Enter a Veil stealth address" to help the user understand before they start that they can't send to a basecoin address.
- Several typos in properties were revealed when debugging was enabled
- As debugging revealed the crash was related to duplicate painter objects, investigation into that revealed several locations where QPainter was included and not necessary.

Ideally this should be done as two separate PRs; one to clean up the little stuff, and one for the crash.  However since it's a means to an end fix, for something more pressing; I'm pushing it as one.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2011-2019 The Bitcoin Core developers
+// Copyright (c) 2019-2020 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -121,7 +122,7 @@ void setupAddressWidget(QValidatedLineEdit *widget, QWidget *parent)
     //widget->setFont(fixedPitchFont());
     // We don't want translators to use own addresses in translations
     // and this is the only place, where this address is supplied.
-    widget->setPlaceholderText(QObject::tr("Enter a Veil address (e.g. %1)").arg(
+    widget->setPlaceholderText(QObject::tr("Enter a Veil stealth address (e.g. %1)").arg(
         QString::fromStdString(DummyAddress(Params()))));
     widget->setValidator(new BitcoinAddressEntryValidator(parent));
     widget->setCheckValidator(new BitcoinAddressCheckValidator(parent));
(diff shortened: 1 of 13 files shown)


src/qt/guiutil.cpp
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) 2011-2019 The Bitcoin Core developers
+// Copyright (c) 2019-2020 The Bitcoin Core developers
```
codeofalltrades: Change to Veil developers.

CaveSpectre11:
> The UI strings that you capitalized were made lower case by this PR in Dec 2019. @StrontiumZ38 was the original author.

Problem was they weren't all lowercase; some were lowercase, some were uppercase.  Looked sloppy.  Frankly I think the uppercase looks cleaner, but I don't hold a strong opinion on it.  @StrontiumZ38; what drove the previous change?  (adding as a reviewer)

src/qt/guiutil.cpp
CaveSpectre11: oops.  nice catch.  joys of distractions :)  All set.

StrontiumZ38:
On case:
For Veil/veil/VEIL, please follow same standards as set by Bitcoin ie. Veil = protocol/project, veil = individual units of currency, VEIL = ticker/currency.

Official documentation for public such as articles, announcements etc. should use sentence case "This is a heading" rather than "This is a Heading".

In a product, I'm less concerned as long as there's a logical, consistent style across everything. It's usually easier to follow a set style guide otherwise you're forever playing catch-up trying to make things consistent, and as this PR shows that's an uphill battle!

Hope that helps, but let me know if there're more questions.

CaveSpectre11:
> In a product, I'm less concerned as long as there's a logical, consistent style across everything. It's usually easier to follow a set style guide otherwise you're forever playing catch-up trying to make things consistent, and as this PR shows that's an uphill battle!

Ok, taken care of

PR #756, opened 2020-03-02
[WIP][RingCT] Increase max inputs to 50

CaveSpectre11:
### Problem
Sending ringct transactions with more than 10 inputs results in a failure.
`Transaction commit failed: bad-witness-nonstandard (code 64)`

### Root Cause
The maximum witness stack item size was coded to 4096.  Particl changed theirs to 8192.  That number is actually 384 + 352 per input.  So 10 inputs == 3904; where 11 is 4256; which exceeds the 4096 and creates the error.

### Solution
Increase the number in a set way.  Instead of using magic numbers in the code, defines have been made to define them in policy.h.  The limit is changed to include 50 input transactions.  The error message cleanup will be done as a separate PR.

### DIFF
--- a/src/policy/policy.cpp
+++ b/src/policy/policy.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
+// Copyright (c) 2019-2020 The Veil developers
 // Distributed under the MIT software license, see the accompanying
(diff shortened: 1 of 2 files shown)


CaveSpectre11:
Back to  WIP; as it needs to be tied to a fork due to the consensus rule change (just because we can create it with more inputs doesn't mean it will be accepted without the consensus needing to change).

Need to target for 1.1 and tie in the check to the POW enforcement.

CaveSpectre11:
Replaced with #763

Issue #661, opened 2019-08-31
Fatal error when choosing zerocoin in coin control


SvenJonas:
When choosing specific zerocoin in coin control a fatal error accours, se attached picture.
This has happend multiple times in several different wallet versions.
This picture is from wallet version 1.0.4.5.
Also attaching end of the logfile, if more is needed let me know.

2019-08-31T07:53:00Z GUI: Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

2019-08-31T07:53:00Z GUI: Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

2019-08-31T07:53:00Z 

************************
UNKNOWN EXCEPTION       
C:\Program Files\Veil\veil-qt.exe in Runaway exception 

![Coin control error](https://user-images.githubusercontent.com/37953479/64061321-98002980-cbd9-11e9-8e93-9baf33a181ab.jpg)

Issue #660, opened 2019-08-30
Pressing ESC in Coin control crashes wallet


Veil-Grumpy:
Pressing the ESC button in Coin control crashes the wallet.
Wallet closes and unable to open it without replacing the blockchain with a snapshot.

Windows 10
v1.0.4.4

related to #661

PR #757, opened 2020-03-07
[GUI] Fix coin control crashes

CaveSpectre11:
### Problem
GUI Coin Control crashes in a couple instances:
- When on the list of utxos (either by clicking on the line or checking the box) and then escaping, wallet crashes (Issue #660).
- When checking a UTXO 10,000 veil or larger will crash the wallet (Issue #661)

### Root Cause
First issue, the `Key_Escape` catch in `CoinControlTreeWidget::keyPressEvent` was getting the wrong object to close.

Second issue, the Coin Control display used the `BitcoinUnits::format` method for the value; which puts whitespace in values greater than 10,000.  This was not being accounted for when converting the value of the selection to a `CAmount`.

### Solution
Fix both issues.  The first by addressing the proper object; the second by using the `BitcoinUnits::parse` method to convert the amount column back to a `CAmount`.

### Testing ###
Ideally, test against a UTXO that is greater than 10000 and not a zerocoin; but also test against zerocoins.  Also make sure to click around the coin control page before hitting escape.

### QA ###
There may be other instances of this type of thing.  Anywhere you find a veil amount in excess of 10000, test around it to see if there's places where it's not converted properly.  Likewise; hit escapes in places to see if there's other areas where it is improperly escaped.  These would be separate issues.  Note that the receive page, and the debug window are the only other two places that catch keyboard input.

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2011-2019 The Bitcoin Core developers
+// Copyright (c) 2019-2020 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -436,7 +437,8 @@ void CoinControlDialog::viewItemChanged(QTreeWidgetItem* item, int column)
             coinControl()->nCoinType = CoinControlDialog::nCurrentCoinTypeSelected;
             coinControl()->fZerocoinSelected = CoinControlDialog::fSpendingZerocoin;
 
-            CAmount nValue = AmountFromValue(item->text(COLUMN_AMOUNT).toStdString());
+            CAmount nValue;
+            BitcoinUnits::parse(BitcoinUnits::VEIL, item->text(COLUMN_AMOUNT), &nValue);
 
             if (CoinControlDialog::fSpendingZerocoin)
                 coinControl()->Select(serialHash, nValue);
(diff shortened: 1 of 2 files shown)


sinetek:
Will test once I fix this keyboard situation.

sinetek:
tested, all good.

seanPhill:
Okay, I've tried a bunch of things. Coin Control is at least uncrashable.
I have had trouble when trying to spend the full amount after fee.
I can spend amounts over 10000 when typing them, but NOT when trying to PASTE the amount, as the copied amount includes either a space as the thousands separator, which is an acceptable number format in Australia (although not the dominant format), and presumably other places.

I do _not_ believe that this should stop this fix being committed.
The pasting with non-numerals in the clipboard, and full amount after fee issue should be addressed in a later PR.

VirtualBox_veil-testenv_default_1545463730648_38346_19_06_2020_08_33_31
VirtualBox_veil-testenv_default_1545463730648_38346_19_06_2020_08_11_30

No problem otherwise. Able to be worked around by typing the amount.

CaveSpectre11:
> I can spend amounts over 10000 when typing them, but NOT when trying to PASTE the amount, as the copied amount includes either a space as the thousands separator, which is an acceptable number format in Australia (although not the dominant format), and presumably other places.
> 

Yes that textual amount is graphical to be easier for the eyes, but does not cut and paste into the CAmount.  It's actually a different twist to the root problem in this PR (the second issue).  The problem was that the code wasn't converting the textual representation to the actual usable number, thus causing the problem.  The ability to paste those textual representations would be a nice feature, but as you noted, is a separate issue.

PR #759, opened 2020-03-18
[RPC] Update getblockchaininfo - Add a difficulty value for each algo type

codeofalltrades:
### Problem ###
The RPC command `getblockchaininfo` only contains two values for difficulty, one for PoS and one for PoW. With the change to 3 PoW algos we need to return a difficulty value for each algo.

### Solution ###
`getblockchaininfo` was updated to return
```
"difficulty_pow": 0,
"difficulty_randomx": 0,
"difficulty_progpow": 0,
"difficulty_sha256d": 0,
"difficulty_pos": 0,
```

### Logic ### 
Search from the chaintip backward. For each block determine the algo type and if a difficulty value > 0 have already been saved for that type. Exit the difficulty lookup once a difficultly for each block type has been found or after checking the last 1440 blocks.

### Testing ###
Run the `getblockchaininfo` cmd via rpc or the console. Verify each difficulty matches the block explorer. You can also check the difficulty of each block with `getblock(getblockhash(180795))`
```
"proof_type": "Proof-of-Stake",
"difficulty": 2426835.587045338,
```

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1318,7 +1318,11 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
             "  ],\n"
             "  \"headers\": xxxxxx,            (numeric) the current number of headers we have validated\n"
             "  \"bestblockhash\": \"...\",       (string) the hash of the currently best block\n"
-            "  \"difficulty\": xxxxxx,         (numeric) the current difficulty\n"
+            "  \"difficulty_pow\": xxxxxx,             (numeric) the current X16RT difficulty\n"
+            "  \"difficulty_randomx\": xxxxxx,         (numeric) the current RandomX PoW difficulty\n"
+            "  \"difficulty_progpow\": xxxxxx,         (numeric) the current ProgPow difficulty\n"
+            "  \"difficulty_sha256d\": xxxxxx,         (numeric) the current SHA256D difficulty\n"
+            "  \"difficulty_pos\": xxxxxx,         (numeric) the current PoS difficulty\n"
             "  \"mediantime\": xxxxxx,         (numeric) median time for the current best block\n"
             "  \"verificationprogress\": xxxx, (numeric) estimate of verification progress [0..1]\n"
             "  \"initialblockdownload\": xxxx, (bool) (debug information) estimate of whether this node is in Initial Block Download mode.\n"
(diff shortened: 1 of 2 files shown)


src/rpc/blockchain.cpp
```diff
@@ -1363,26 +1367,30 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
...
+            nDiffRandomX = GetDifficulty(pindex);
+        } else if (pindex->IsSha256DProofOfWork() && nDiffSha256D == 0) {
+            nDiffSha256D = GetDifficulty(pindex);
+        } else if (pindex->IsX16RTProofOfWork() && nDiffX16rt == 0) {
+            nDiffX16rt = GetDifficulty(pindex);
         }
-    } else {
-        nDiffPoS = GetDifficulty(pindex);
-        while (pindex->pprev) {
-            pindex = pindex->pprev;
```
CaveSpectre11: @codeofalltrades ; where does pindex iterate down the chain?  I think you're missing the `pindex= pindex->pprev`; at the bottom of the loop.

I'm guessing the while loop runs the full 1440 before breaking out and you only end up with the difficulty for the chain tip.

src/rpc/blockchain.cpp
```diff
@@ -1363,26 +1367,30 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
...
+        if (pindex->IsProofOfStake() && nDiffPoS == 0) {
+            nDiffPoS = GetDifficulty(pindex);
+        } else if (pindex->IsProgProofOfWork() && nDiffProgPow == 0) {
+            nDiffProgPow = GetDifficulty(pindex);
+        } else if (pindex->IsRandomXProofOfWork() && nDiffRandomX == 0) {
+            nDiffRandomX = GetDifficulty(pindex);
+        } else if (pindex->IsSha256DProofOfWork() && nDiffSha256D == 0) {
+            nDiffSha256D = GetDifficulty(pindex);
+        } else if (pindex->IsX16RTProofOfWork() && nDiffX16rt == 0) {
+            nDiffX16rt = GetDifficulty(pindex);
```
CaveSpectre11: Flip these if conditions.  `if (we need it && it's one of the blocks)`.  Ideally we should probably make those check routines be inlined; but not really necessary for this one.  I would rather the efficiency of not checking the type of block unless you need that type of block.

CaveSpectre11:
We also need to solve that asm include issue

src/rpc/blockchain.cpp
codeofalltrades: Switched

codeofalltrades:
@CaveSpectre11 please take another look. I corrected the items you noted.

src/rpc/blockchain.cpp
```diff
@@ -1318,7 +1318,11 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
             "  ],\n"
             "  \"headers\": xxxxxx,            (numeric) the current number of headers we have validated\n"
             "  \"bestblockhash\": \"...\",       (string) the hash of the currently best block\n"
-            "  \"difficulty\": xxxxxx,         (numeric) the current difficulty\n"
```
mimirmim: This is a non-backwards compatible change. We can't remove fields as we don't know what someone might be using to parse the JSON output. Might cause their end to break.

codeofalltrades: This was a bug fix of the rpc help. The JSON returned keys for "difficulty_pow" and "difficulty_pos" have not changed.
`   obj.pushKV("difficulty_pow",        nDiffPoW);
    obj.pushKV("difficulty_pos",        nDiffPoS);`

[current master](https://github.com/Veil-Project/veil/blob/6a1f9f8426b58d98aa3e55aa13e3c6464f336303/src/rpc/blockchain.cpp#L1366)

CaveSpectre11: @mimirmim yes; Code is correct; help doc was incorrect.  See below:

image

so it's only additions, not changes of existing.

CaveSpectre11:
utACK https://github.com/Veil-Project/veil/pull/759/commits/8c3232720e14a6a5b648120aedf7b42d21a7360a

PR #766, opened 2020-03-31
Address segmentation faults in the Log message handlers on macOS.

sinetek:
### Problem ###
CaveSpectre and mimir (others?) have reported a crash on recent macOS builds.
The crash is a segmentation fault in a debug handler named DebugMessageHandler(),
at runtime, before the slash screen, or main window, is shown.

### Root Cause ###
The issue has been tracked down to incorrect usage of the debug.log message handler override.
This is a part of Qt that allows us to log to debug.log, using a bypassed message handler
in the form of a callback function.

The issue is that we should not access some of the “context” information, passed in the form of
a QmessageLogContext.

If we peruse the Qt documentation:

Note: By default, this information is recorded only in debug builds. You can overwrite this explicitly by defining QT_MESSAGELOGCONTEXT or QT_NO_MESSAGELOGCONTEXT.

### Solution ###
The correct way to fix this is to not make use of any of the potentially unavailable debugging information.

Furthermore, this Pull Request will pull in 4 additional commits, that are cherry-picks from Bitcoin upstream. The reason for this is to fix a build-time error on recent macOS versions,
due to code deprecation in the dock icon logic.

### Unit Testing Results ###

1)  The build is successful.
3)  Additionally, because this Pull Request touches the dock icon handlers, it is requested that we verify the dock icon is sane: separators, exit entry, and such.


sinetek:
@CaveSpectre11 
Please continue discussion here. My bad, it seems handling of PR's on Github is a bit strange, one can't delete an open pull request, is this okay?

src/qt/veil.cpp
```diff
@@ -146,16 +146,12 @@ static void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTrans
...
-   Q_UNUSED(context);
+    Q_UNUSED(context);
     if (type == QtDebugMsg) {
         LogPrint(BCLog::QT, "GUI: %s\n", msg.toStdString());
     } else {
         LogPrintf("GUI: %s\n", msg.toStdString());
     }
-#else // QT_MESSAGELOGCONTEXT needs to be added to the DEFS line of the Makefile
-    LogPrintf("GUI: %s:%d %s\n", context.file, context.line, msg.toStdString());
-#endif
```
CaveSpectre11: Let's comment out the relevant parts instead of deleting them; Add a comment also that says something like "can be used for debugging but may need changes for MacOS debugging"

sinetek: Sure, I don't really care.  
How about :  "Please don't enable this publicly. It may leak information, and we've been bit by a nasty crash where the context is not available in non-debug builds."

sinetek: I'm still unsure where that QT_MESSAGELOGCONTEXT is coming from, probing around.

sinetek: It's set by the Qt headers depending on how Qt was built.

CaveSpectre11: okay.  yea, just a message so it's clear why it's commented out, but there for anyone finding it by chasing through the debug logging of Qt.

Since it's dependent on the building style, yep.  Removing it is the right approach rather than changing the building style to remove it.  Perhaps we can re-visit later when we figure out exactly what's needed to clear up the crashes... e.g. `#ifndef (QT_MESSAGECONTEXT || DEBUG) `or whatever, and put it back in.  But for now this is good enough.

4x13:
ACK This does fix the segfault when natively compiling on OSX

src/qt/veil.cpp
```diff
@@ -146,15 +146,19 @@ static void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTrans
...
-#ifndef QT_MESSAGELOGCONTEXT
-   Q_UNUSED(context);
+#if 0
+
+    // Please don't enable this publicly. It may leak information, and we've been bit by a nasty crash where the context is not available in non-debug builds.
+
+LogPrintf("GUI: %s:%d %s\n", context.file, context.line, msg.toStdString());
+
+#else
+    Q_UNUSED(context);
```
CaveSpectre11, line 156: Just comment it out simply; so that it doesn't change the functionality of the #ifdef that _is_ required.

What I'm asking for is simple.

```
// #ifdef QT_MESSAGELOGCONTEXT
// Do not use this except in debugging environments as some OSs don't handle it properly
//QT_MESSAGELOGCONTEXT needs to be added to the DEFS line of the Makefile
//    LogPrintf("GUI: %s:%d %s\n", context.file, context.line, msg.toStdString());
// #else
```

sinetek:
@CaveSpectre11  Changes done.

CaveSpectre11:
- Includes 3 of 6 commits from - https://github.com/bitcoin/bitcoin/pull/14123 
    - https://github.com/bitcoin/bitcoin/pull/14123/commits/53bb6be3f8a50ee9e5c4d7e9155236152e7c4b7c
    - https://github.com/bitcoin/bitcoin/pull/14123/commits/2464925e7be832d4926b6204169bbbc1646c6368
-

CaveSpectre11:
Includes 3 of 6 commits from: https://github.com/bitcoin/bitcoin/pull/14123/commits
Includes https://github.com/bitcoin/bitcoin/pull/16720

Zannick (approved):
utACK 386e15e40b9259e4aa5b4ab8bb61af3c7f8f23d8

PR #767, opened 2020-03-31
Disable Dark Mode on recent versions of macOS.

sinetek:
### Problem ###
Recent macOS releases have introduced a so-called “Dark Mode” feature meant to give mac users an alternative theme for their operating system.
As of this time, the stylesheets we have in the Qt application do not work well with this change of colours.

In other words:
dmode wrong

### Root Cause ###
The root cause is a change in the Qt styling system on macOS, broadly speaking the problem was introduced by Apple Inc.

### Solution ###
The way I fix this is by just enforcing the known-good Aqua style.

In other words:
dmode right

### Unit Testing Results ###

1)  The build is successful.
3)  The main window is visible and usable.

Versions to test :  10.14, 10.13.   The former introduced the problem, check for regression on later.

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -171,7 +171,8 @@ QT_MOC_CPP = \
 
 BITCOIN_MM = \
   qt/macdockiconhandler.mm \
-  qt/macnotificationhandler.mm
+  qt/macnotificationhandler.mm \
+  qt/macdarkmode.mm
 
 QT_MOC = \
   qt/veil.moc \
@@ -211,6 +212,7 @@ BITCOIN_QT_H = \
   qt/intro.h \
   qt/macdockiconhandler.h \
   qt/macnotificationhandler.h \
+  qt/macdarkmode.h \
   qt/modaloverlay.h \
   qt/networkstyle.h \
   qt/notificator.h \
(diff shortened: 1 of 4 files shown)


sinetek:
Addressed a small compile issue on other platforms.
Ready for review now.

sinetek:
Why is the travis not pulling in new files?

src/Makefile.qt.include
```diff
@@ -171,7 +171,8 @@ QT_MOC_CPP = \
```
codeofalltrades, line 171: qt/macdarkmode.h needs to be added to the 
BITCOIN_QT_H
and 
QT_MOC_CPP
sections

sinetek: I'm not using any Qt classes, just regular QT_H is enough

sinetek:
Added the header file.

4x13:
NACK This won't compile natively on OSX, this PR ides with the following error. Once PR's 723, 727, 757, 758, 762 and 766 are merged.

Screenshot from 2020-04-02 22-38-20

sinetek:
Found a workaround.  Waiting on my 10.13 install to test.

CaveSpectre11:
utACK https://github.com/Veil-Project/veil/commit/a9bffafec3fa670ffaf3174ac5e6f96461189498

seanPhill:
Screen Shot 2020-07-05 at 11 08 25 pm
Colour scheme looks fine now, while Catalina Dark Mode is on.

PR #773, opened 2020-04-14
[RPC] Help user out if RPC return is Null.

sinetek:
### Problem ###
For the sake of lowering confusion with users, requests used in console that return 'null' should be accompanied with a short text description of what's been achieved. This will solve many cases of users believing something has gone wrong when entering a command in console and getting null as a return.

### Root Cause ###
Not all RPC calls return a value.

### Solution ###
In case of a Qt context.
If the RPC’s call return value is a Null object, then we append a string to help the user out.
Otherwise, the command-line RPC behaviour remains unmodified.

### Unit Testing Results ###
Run various RPC commands and see if the returned value is helpful. 

`clearmempool` returns Null, you can test with this command.
Other such commands (not exhaustive):
`savemempool`
`syncwithvalidationinterfacequeue`

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -382,6 +382,10 @@ bool RPCConsole::RPCParseCommandLine(interfaces::Node* node, std::string &strRes
                 strResult = lastResult.get_str();
             else
                 strResult = lastResult.write(2);
+
+            if (lastResult.isNull())
+                strResult.append(" (is not necessarily indicative of a fatal error)");
+
         case STATE_ARGUMENT:
         case STATE_EATING_SPACES:
             return true;


CaveSpectre11:
First concern was changing RPC behavior; but on first glance it looks like it's limited to the Qt console.  Definitely needs to have testing to determine the scope of changes.  Would be helpful to have a list of commands that will be affected by the change.

sinetek:
> First concern was changing RPC behavior; but on first glance it looks like it's limited to the Qt console. Definitely needs to have testing to determine the scope of changes. Would be helpful to have a list of commands that will be affected by the change.

Thought of that. Depends for what purpose you want it:  for testing you can run `clearmempool` or run a search for `return NullUniValue`. Otherwise no, the list would quickly become out of sync with the project.

CaveSpectre11:
> or run a search for `return NullUniValue`. 

The comment was to effectively provide a list for testing purposes, rather than having people have to dig it out of the code (which QA won't be doing)

sinetek:
> > or run a search for `return NullUniValue`.
> 
> The comment was to effectively provide a list for testing purposes, rather than having people have to dig it out of the code (which QA won't be doing)

done, added to instructions. thx

Rock-N-Troll:
The following are locations where 
`return NullUniValue;`
is the success return value at the end of the method.

some options are (as leveraged from rpczerocoin.cpp):

```
 //todo: better response
    return "done";
```

or
a descriptive message like:

`return _("Auto spend zerocoin stopped");`

**rpcwallet.cpp**
`setlabel`
`setnonautomintaddress`
`abandontransaction`
`backupwallet`
`keypoolrefill`
`walletpassphrase`
`walletpassphrasechange`
`walletlock`
`unloadwallet`
`sethdseed`

**rpcdump.cpp**
`importprivkey`
`importaddress`
`removeprunedfunds`
`importpubkey`
`importwallet`

**rpczerocoin.cpp**
`getautospendaddress`

CaveSpectre11:
> The following are locations where
> `return NullUniValue;` **Success Case**
> is the success return value at the end of the method.
> 

Thanks for the legwork here; It's one that's been sitting because we discussed wanting to do just that; have useful feedback to the user (both in CLI and QT), we've just not had time to do it right, so to speak.  Pretty much the model that was established with `setminingalgo` in #878

seanPhill:
I've added the label "Core App" ie. GUI, because of the need for this to apply to the GUI: Debug Console rather than CLI RPC outputs themselves.

PR #774, opened 2020-04-17
[GUI] Do not use Qt's debugging information in the log handler.

CaveSpectre11:
(Commit cherry picked from #766 for fast tracking)
### Problem ###
CaveSpectre and mimir (others?) have reported a crash on recent macOS builds.
The crash is a segmentation fault in a debug handler named DebugMessageHandler(),
at runtime, before the slash screen, or main window, is shown.

### Root Cause ###
The issue has been tracked down to incorrect usage of the debug.log message handler override.
This is a part of Qt that allows us to log to debug.log, using a bypassed message handler
in the form of a callback function.

The issue is that we should not access some of the “context” information, passed in the form of
a QmessageLogContext.

If we peruse the Qt documentation:

Note: By default, this information is recorded only in debug builds. You can overwrite this explicitly by defining QT_MESSAGELOGCONTEXT or QT_NO_MESSAGELOGCONTEXT.

### Solution ###
The correct way to fix this is to not make use of any of the potentially unavailable debugging information.

### Unit Testing Results ###

1)  The build is successful.

### DIFF
--- a/src/qt/veil.cpp
+++ b/src/qt/veil.cpp
@@ -146,16 +146,19 @@ static void initTranslations(QTranslator &qtTranslatorBase, QTranslator &qtTrans
 /* qDebug() message handler --> debug.log */
 void DebugMessageHandler(QtMsgType type, const QMessageLogContext& context, const QString &msg)
 {
-#ifndef QT_MESSAGELOGCONTEXT
-   Q_UNUSED(context);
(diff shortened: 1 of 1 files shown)


mimirmim:
Obviously can't build it right now but clear enough to get moving forward on this.

ACK (but fails).

sinetek:
ack

PR #779, opened 2020-05-04
[Tests] Fix some tests and use CAmount for SetValue Calls

akshaynexus:
This PR Fixes the mempool_tests,bech32_tests,bip32_tests and sets the value arg for SetValue calls to CTxOutBase to const CAmount& to match the header func.

There is a segfault on SetValue in blockencoding_tests.cpp that will need more investigation to be fixed.

### DIFF
--- a/src/policy/feerate.h
+++ b/src/policy/feerate.h
@@ -25,7 +25,7 @@ public:
     /** Fee rate of 0 satoshis per kB */
     CFeeRate() : nSatoshisPerK(0) { }
     template<typename I>
-    CFeeRate(const I _nSatoshisPerK): nSatoshisPerK(_nSatoshisPerK) {
+    explicit CFeeRate(const I _nSatoshisPerK): nSatoshisPerK(_nSatoshisPerK) {
         // We've previously had bugs creep in from silent double->int conversion...
         static_assert(std::is_integral<I>::value, "CFeeRate should be used without floats");
     }
--- a/src/test/blockencodings_tests.cpp
+++ b/src/test/blockencodings_tests.cpp
@@ -3,11 +3,11 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include <blockencodings.h>
-#include <consensus/merkle.h>
 #include <chainparams.h>
+#include <consensus/merkle.h>
 #include <pow.h>
-#include <random.h>
-
+#include <streams.h>
+#include <validation.h>
 #include <test/test_veil.h>
 #include <veil/budget.h>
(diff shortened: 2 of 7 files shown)


CaveSpectre11:
@akshaynexus are these specifically dependent on the v1.1 branch or can they be pushed to master?

akshaynexus:
Depends on if those tests were changed on master and dinners from 1.1.0

mimirmim:
Are these fixes from Bitcoin upstream? I swear I've seen almost exactly this when I was going to try to fix these once upon a time.

CaveSpectre11:
> Depends on if those tests were changed on master and differs from 1.1.0

Please do that research and push to master instead of v1.1 if they are not diverged.  The less diverging between master and v1.1, the better when merge day comes.

akshaynexus:
Ok,and yes some of the fixes are from bitcoin upstream.The bech32 test fix isnt cause veil has a higher threshold of max chars so i added more chars so that it is a invalid bech32 addr.

sinetek:
then put that in the commit message.

mimirmim:
Where possible, do mention / cherrypick from Bitcoin upstream.

src/test/blockencodings_tests.cpp
```diff
@@ -68,9 +68,8 @@ BOOST_AUTO_TEST_CASE(SimpleRoundTripTest)
 {
     CTxMemPool pool;
     TestMemPoolEntryHelper entry;
+    LOCK2(cs_main,pool.cs);
```
CaveSpectre11, line 71: I'm extremely hyper-sensitive to cs_main additions; do we know exactly what the lock is being used for in this context?

akshaynexus: I will recheck this change,maybe was a mistake when porting over the mempool test fixes

CaveSpectre11: might not be a porting mistake.  They like to sprinkle around cs_main like an american likes their salt.

mimirmim: It is just in tests, I'm personally not too worried about it.

src/policy/feerate.h
```diff
@@ -56,4 +56,4 @@ class CFeeRate
     }
 };
 
-#endif //  BITCOIN_POLICY_FEERATE_H
+#endif //  BITCOIN_POLICY_FEERATE_H
```
mimirmim, line 59: Needs new line at EOF

CaveSpectre11:
Won't be going into 1.1.0.  Need to point to master branch.

PR #780, opened 2020-05-07
init: remove boost from ThreadImport.

sinetek:
**Problem**
This is a refactor fix. We want to remove boost functions slowly.

**Root Cause**

**Solution**
We can work around the boost interruption point by just returning, and shutting down the program, instead. It is much cleaner than relying on boost to redirect control flow.

**Unit Testing Results**
Can be tested by calling -reindex or -loadblock and then pressing CTRL+C.

Should print something like:

> ...
> 2020-04-27T19:34:31Z [loadblk] Reindexing block file blk00005.dat...
> ^C2020-04-27T19:34:32Z [loadblk] Shutdown requested. Exit ThreadImport
> 2020-04-27T19:34:32Z [qt-init] Interrupting HTTP server
> ...
>

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -715,6 +715,10 @@ static void ThreadImport(std::vector<fs::path> vImportFiles)
                 break; // This error is logged in OpenBlockFile
             LogPrintf("Reindexing block file blk%05u.dat...\n", (unsigned int)nFile);
             LoadExternalBlockFile(chainparams, file, &pos);
+            if (ShutdownRequested()) {
+                LogPrintf("Shutdown requested. Exit %s\n", __func__);
+                return;
+            }
             nFile++;
         }
         pblocktree->WriteReindexing(false);
@@ -732,6 +736,10 @@ static void ThreadImport(std::vector<fs::path> vImportFiles)
             fs::path pathBootstrapOld = GetDataDir() / "bootstrap.dat.old";
             LogPrintf("Importing bootstrap.dat...\n");
             LoadExternalBlockFile(chainparams, file);
+            if (ShutdownRequested()) {
+                LogPrintf("Shutdown requested. Exit %s\n", __func__);
+                return;
+            }
             RenameOver(pathBootstrap, pathBootstrapOld);
         } else {
             LogPrintf("Warning: Could not open bootstrap file %s\n", pathBootstrap.string());
@@ -744,6 +752,10 @@ static void ThreadImport(std::vector<fs::path> vImportFiles)
         if (file) {
             LogPrintf("Importing blocks file %s...\n", path.string());
             LoadExternalBlockFile(chainparams, file);
+            if (ShutdownRequested()) {
+                LogPrintf("Shutdown requested. Exit %s\n", __func__);
+                return;
+            }
         } else {
             LogPrintf("Warning: Could not open blocks file %s\n", path.string());
         }
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -5678,7 +5678,7 @@ bool LoadGenesisBlock(const CChainParams& chainparams)
     return g_chainstate.LoadGenesisBlock(chainparams);
 }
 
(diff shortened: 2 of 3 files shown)


src/validation.cpp
```diff
@@ -5795,9 +5795,7 @@ bool LoadExternalBlockFile(const CChainParams& chainparams, FILE* fileIn, CDiskB
     } catch (const std::runtime_error& e) {
         AbortNode(std::string("System error: ") + e.what());
     }
-    if (nLoaded > 0)
-        LogPrintf("Loaded %i blocks from external file in %dms\n", nLoaded, GetTimeMillis() - nStart);
-    return nLoaded > 0;
+    LogPrintf("Loaded %i blocks from external file in %dms\n", nLoaded, GetTimeMillis() - nStart);
```
CaveSpectre11, line 5798: Leave the boolean return in; and "return false" on the ShutdownRequested().  While the calling routines don't check the success of calls to this routine currently; it is something that may get addressed at a later date (as well as better error handling within `LoadExternalBlockFile()`.  

Rest of the code looks good.  Just this part that I prefer is left as is; or perhaps refined so that
```
if (nLoaded > 0) {
    LogPrintf("...)
} else {
    LogPrintf("ERROR: ...
}

File: src/init.cpp [hunk 2]
```diff
@@ -715,6 +715,10 @@ static void ThreadImport(std::vector<fs::path> vImportFiles)
                 break; // This error is logged in OpenBlockFile
             LogPrintf("Reindexing block file blk%05u.dat...\n", (unsigned int)nFile);
             LoadExternalBlockFile(chainparams, file, &pos);
+            if (ShutdownRequested()) {
+                LogPrintf("Shutdown requested. Exit %s\n", __func__);
```
CaveSpectre11, line 719: I may prefer "Aborted " over "Exit", but not a show stopper.

src/validation.cpp
sinetek: nack, there are only 3 calls of the functions and none used the return value, and the function is not expected to fail per "the try/catch". I really think the commit is better as is without the confusing bool.

sinetek:
Travis early failed due to caching I believe..

src/validation.cpp
mimirmim: I agree with @CaveSpectre11, just leave it in for his reasons stated that it likely will get addressed at a later date with better error handling. Out of scope to remove it.

sinetek: I disagree. we'll go with democracy, undoing that bit.

codeofalltrades:
> Travis early failed due to caching I believe..
Yep, reran. All passing now.

sinetek:
Addressed the issues, ready for another review.

seanPhill:
ACK
With this PR #780 I've done numerous closes of the veil daemon (and of the Qt wallet) (with or without -reindex), with ctrl-c or otherwise and they all end without problem, with something like:
```
2020-09-15T00:00:38Z ThreadStaging exiting
2020-09-15T00:00:38Z tor: Thread interrupt
2020-09-15T00:00:38Z torcontrol thread exit
2020-09-15T00:00:38Z dnsseed thread exit
2020-09-15T00:00:38Z addcon thread exit
2020-09-15T00:00:38Z Shutdown: In progress...
2020-09-15T00:00:38Z UPNP_DeletePortMapping() returned: 0
2020-09-15T00:00:38Z upnp thread exit
2020-09-15T00:00:38Z msghand thread exit
2020-09-15T00:00:38Z net thread exit
2020-09-15T00:00:38Z opencon thread exit
2020-09-15T00:00:38Z Shutdown requested. Exit ThreadImport
2020-09-15T00:00:38Z ThreadStakeMiner() interrupted
2020-09-15T00:00:38Z ThreadStakeMiner exiting
2020-09-15T00:00:38Z scheduler thread interrupt
2020-09-15T00:00:38Z Shutdown: done
```

PR #782, opened 2020-05-07
[CI] Add basic GitHub Actions workflow

akshaynexus:
This PR Adds a basic Github Actions workflow.The workflow builds binaries for macOS,Linux and windows.


sinetek:
In no particular order.

What is this?
Why do we need this?
Who is going to maintain it?
Is it better than Travis?

akshaynexus:
- Github Actions is a alternative to Travis.
- GitHub actions allows for artifacts/ binaries to be uploaded on each run if the run succeeds.This is helpful for testers who may want to test the pr without waiting for a build to finish on their own pc.
- I will be able to maintain this.
- Depends,On travis its free to use for open source with no time limits,but on github actions there is a certain limit.
Screen Shot 2020-05-07 at 6 14 26 PM

PR #783, opened 2020-05-07
[CI] Add basic GitHub Actions workflow

akshaynexus:
This PR Adds a basic Github Actions workflow.The workflow builds binaries for macOS,Linux and windows.

Previous PR was closed due to the changes being on top of the 1.1.0 branch instead of master.


akshaynexus:
Will do

akshaynexus:
Squashed changes to 1 commit.

CaveSpectre11:
Pulling back from QA to Code Review.  Nothing to QA since we don't have Github Actions enabled in the Veil github, and the "code review" decision is if we merge this in as-is, as a stepping stone towards implementing Actions, or if we hold it until we figure out what is wrong with the man page build portion when building via actions.

@codeofalltrades what are your thoughts?

akshaynexus:
for qa you can check the branch the pr is from to see if builds are sucessful.Not necessarily needed to be enabled on veil repo to test it.

CaveSpectre11:
> for qa you can check the branch the pr is from to see if builds are successful. Not necessarily needed to be enabled on veil repo to test it.

Thanks for your input Akshay.  That's actually what I'm driving off of.  I'm inclined to merge as a stepping stone, but having QA judge on the results in your branch won't really be an accurate assessment since the builds do fail due to the man page build issue.  I'd rather not spend the time to debug that currently, but I'm also not inclined to have this PR sit around until we do have that time.   That's pretty much what the re-review is going to decide.  If we wait for this PR to be a complete alternate solution to travis, or we take it as a stepping stone towards that goal and get it working at a later date.

CaveSpectre11:
Closing this because trying to rebase it isn't going too well; so running further tests to make this functional and submitting as a separate PR.

Issue #784, opened 2020-05-09
Wallet bombs at block 1528 during reindex


sinetek:
While working on #781 we've encountered a bug that manifests when reindexing a chain.
Run the wallet like so:

`./src/qt/veil-qt -printtoconsole -reindex-chainstate`

... and eventually it will fail in such a way:

```
2020-05-09T17:37:10Z UpdateTip: new best=7b11c0cfeb93f9d08efdc81f6f50cc8f2822fd2b048c0d15a6aa81ff65566a43 height=1524 type=PoS version=0x20000000 tx=5735 date='2019-01-04T02:11:42Z' cache=0.8MiB(4571txo)
2020-05-09T17:37:10Z UpdateTip: new best=2d04f093e82021930d66680e4066c7b05b58b52684a333c61945d032b30a9df6 height=1525 type=PoS version=0x20000000 tx=5737 date='2019-01-04T02:11:51Z' cache=0.8MiB(4573txo)
2020-05-09T17:37:10Z UpdateTip: new best=e966faf49110e31b0258609d576f6bbde98f38cc1fb5f7630f552d86ad18d496 height=1526 type=PoS version=0x20000000 tx=5739 date='2019-01-04T02:11:44Z' cache=0.8MiB(4575txo)
2020-05-09T17:37:10Z UpdateTip: new best=6569c3676a97a7a13945e00d9672687f39e0535a5d104e6227a416b2174eda8a height=1527 type=PoS version=0x20000000 tx=5741 date='2019-01-04T02:12:05Z' cache=0.8MiB(4577txo)
2020-05-09T17:37:10Z UpdateTip: new best=c97704555f8d49cd66fd198162803004139cbca84724c90600c319ae1b2b34c4 height=1528 type=PoS version=0x20000000 tx=5744 date='2019-01-04T02:12:18Z' cache=0.8MiB(4579txo)
2020-05-09T17:37:10Z ConnectBlock: Duplicate anon-output 0293bdd4159b66df68f4e68129427a71a5b6e333d9231025e4103b61f97c7794f0, index 1, above last index 0.
2020-05-09T17:37:10Z Attempting to repair anon index.
2020-05-09T17:37:10Z ERROR: ConnectTip(): ConnectBlock 3494732d196280247c924bc7fa50372af3d01d2e2a57a0d9cc3b8ace407940dd failed
2020-05-09T17:37:10Z Failed to connect best block ( (code 0))
[...]

PR #785, opened 2020-05-15
[ReIndex] Skip RingCT TX computation during resynching

CaveSpectre11:
### Problem
**_Every other_** reindex fails when processing RingCT transactions.  (Issue #784 )

### Root Cause
When reindexing, the existing RingCT transactions still exist in the wallet; therefore when you compare a RingCT transaction against the list, it's found and generates an error.  The code then rolls back and clears the transaction record, so the next reindex is successful since the record doesn't exist anymore.

### Solution
As in other areas of the reindex process, some calculations are skipped since it's not a new block, but rather reindexing an existing block... so you already have some information on it.  This now skips the existence test of ringct transactions on a reindex.

### Testing
It is important that you run any test twice; as the first test may not have the records in them (from being deleted on a previous reindex).  Initial tests were done with `-reindex-chainstate -stopatheight=1600` as the first RingCT is in block 1529.  One pass will fail at block 1529, the other will successfully pass it.  The failing test appears as:
```
2020-05-09T17:37:10Z UpdateTip: new best=c97704555f8d49cd66fd198162803004139cbca84724c90600c319ae1b2b34c4 height=1528 type=PoS version=0x20000000 tx=5744 date='2019-01-04T02:12:18Z' cache=0.8MiB(4579txo)
2020-05-09T17:37:10Z Attempting to repair anon index.
2020-05-09T17:37:10Z ERROR: ConnectTip(): ConnectBlock 3494732d196280247c924bc7fa50372af3d01d2e2a57a0d9cc3b8ace407940dd failed
```

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2679,7 +2679,7 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                 CTxOutRingCT *txout = (CTxOutRingCT*)tx.vpout[k].get();
 
                 int64_t nTestExists;
-                if (!fVerifyingDB && pblocktree->ReadRCTOutputLink(txout->pk, nTestExists)) {
+                if (!fSkipComputation && !fVerifyingDB && pblocktree->ReadRCTOutputLink(txout->pk, nTestExists)) {
                     control.Wait();
 
                     if (nTestExists > pindex->pprev->nAnonOutputs) {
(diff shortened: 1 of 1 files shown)


sinetek:
Amazing work, @CaveSpectre11! This is refreshing, we have a PR or two pending on this.

sinetek:
Reviewed, tested. Works as advertised.
ACK

seanPhill:
tACK
I've run this PR785 twice on mainnet with `$ ./veild -reindex-chainstate -stopatheight=1600`
No crash. No problem.
Screen Sharing Picture 28 June 2020 at 9 59 52 pm AEST
Screen Sharing Picture 28 June 2020 at 9 57 16 pm AEST

PR #786, opened 2020-05-18
gui: Fix leak in CoinControlDialog::updateView

sinetek:
**problem**

This is admittedly a small issue, but nonetheless there's a leak in the CoinControl view update when the tree mode is not enabled. As you see the CCoinControlWidgetItem gets created when there's no need for it.

I think this right, please check that I didn't miss other logic there.

**how to solve**

Move the `new CCoinControlWidgetItem` a bit further down in the proper `if (treeMode)` area.

**how to test.**

You want to go into the Wallet ->  Send ->   Coin Control at the bottom.   Toggle knobs, play with the options, nothing should crash or behave strangely.

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -766,8 +766,7 @@ void CoinControlDialog::updateView(int nCoinType)
 
     if (nCoinType == OUTPUT_STANDARD && !CoinControlDialog::fSpendingZerocoin) {
         for (const auto &coins : model->wallet().listCoins()) {
-            CCoinControlWidgetItem *itemWalletAddress = new CCoinControlWidgetItem();
-            itemWalletAddress->setCheckState(COLUMN_CHECKBOX, Qt::Unchecked);
+            CCoinControlWidgetItem *itemWalletAddress{nullptr};
             QString sWalletAddress = QString::fromStdString(EncodeDestination(coins.first));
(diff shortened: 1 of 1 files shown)


CaveSpectre11:
I did take a look at this and I need to get back to it; build and do some testing.  I'm not sure I see how the existing code is a leak, but regardless of that; My main concern to investigate is:

`            itemWalletAddress->setCheckState(COLUMN_CHECKBOX, Qt::Unchecked);`
I want to make sure that truly isn't needed in the non tree mode.

I also want to understand the effect on the code lower down:
```
                if (treeMode) itemOutput = new CCoinControlWidgetItem(itemWalletAddress);
                else itemOutput = new CCoinControlWidgetItem(ui->treeWidget);
```

seanPhill:
I've been running this PR for a day or two, and tested coin control after accumulating a mature balance and I can't crash Coin Control, but I also could not crash it when I tested PR757 (now committed) "Fix Coin Control Crashes".

BUT I cannot SEE ANY RingCT utxos in Coin Control in this wallet!

VirtualBox_Linux Ubuntu 20 04 (from 19 04)_24_06_2020_21_02_15
VirtualBox_Linux Ubuntu 20 04 (from 19 04)_24_06_2020_21_01_43
VirtualBox_Linux Ubuntu 20 04 (from 19 04)_24_06_2020_21_01_24

seanPhill:
All go back to before 786 and see if any RingCT utxos show, and if not, I'll go back to the release and see.

PR #787, opened 2020-05-24
[GUI]: add a small indicator for the last block time in the left panel.

sinetek:
veil-lastblocktime

**Problem**
A small improvement to the usability: the user often needs to see what the block time is on the current node, and it is very annoying having to open the debug console to extract this information.

**Root Cause**
The value is not visible.

**Solution**
I have added two labels to the main form, and rigged up the signal to update the block time.

**Testing**
Should be straightforward.
Maybe show this around to see what people think.

### DIFF
--- a/src/qt/veil/balance.cpp
+++ b/src/qt/veil/balance.cpp
@@ -170,6 +170,8 @@ void Balance::onBtnBalanceClicked(int type){
 
 void Balance::setClientModel(ClientModel *model){
     this->clientModel = model;
+
+    connect(model, SIGNAL(numBlocksChanged(int,QDateTime,double,bool)), this, SLOT(setNumBlocks(int,QDateTime,double,bool)));
 }
 
 void Balance::setWalletModel(WalletModel *model){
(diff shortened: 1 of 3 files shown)


Veil-Grumpy:
Just a suggestion but what do you think about putting this info in the right bottom corner of the wallet?  

Sun May 24 19:47:19 2020 - <blocknr>

seanPhill:
When I first saw the Last Block Time position in the screenshot above it took me a while to find it, BUT now that I have it running and syncing on mainnet it actually catches the eye in that spot no problem!

seanPhill:
Discrepancy noticed! While certain peers were stalling the downloading of blocks it was taking a long time to sync (mainnet) and then I noticed that the new latest block time was reflecting the synced header block and not the downloaded block, and therefore conflicts with the time shown on the Information tab.
Screen Sharing Picture 14 July 2020 at 11 24 17 pm AEST

CaveSpectre11:
> This recurred and corrected itself at times afterward. I think this is not essential to fix, but only necessary to understand that when block downloads are being stalled by a peer the displayed latest block time might not always be correct.

Nice catch Sean.  I want to move it over to the right of the block height on the bottom bar; and since that block height shows the last block received, not the last header received, the discrepancy needs to be resolved (so the date is tied to that block height)

CaveSpectre11:
OBE #865 [which includes this commit]

PR #788, opened 2020-05-25
Prevent UB when computing abs value for num opcode serialize

sinetek:
**Problem, Cause**

* It seems that the original author of the snipped used glibc code:
  https://github.com/lattera/glibc/blob/master/stdlib/abs.c

However this is undefined behaviour under certain circumstances:
  https://stackoverflow.com/questions/17313579/is-there-a-safe-way-to-get-the-unsigned-absolute-value->

A relevant example can be found at :
  https://godbolt.org/z/yRwtCG

This was first reported in:
  https://github.com/bitcoin/bitcoin/issues/18046

**How to fix**

The solution is to add 1 instead of negating.

**How to test**

### DIFF
--- a/src/script/script.h
+++ b/src/script/script.h
@@ -328,7 +328,7 @@ public:
 
         std::vector<unsigned char> result;
         const bool neg = value < 0;
-        uint64_t absvalue = neg ? -value : value;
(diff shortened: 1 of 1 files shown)


mimirmim:
Casting as uint64, yeah makes sense enough given the context.

utACK https://github.com/Veil-Project/veil/commit/e19a58bcd6a7bb72fad3e706f087abc300a831a8

PR #790, opened 2020-05-25
Correct incorrect usage of string arithmetic.

sinetek:
**Problem**

It is not acceptable to add an int to a string. This will result in completely different behaviour than what is intended.

**Solution**

The simple way to fix this is std::to_string() invocation to create a string object and then the plus works as intended.

Furthermore, this fixes a bug where the ? Operator has a lower priority as +.

**Testing.**

~Check that the signrawtransaction rpc command’s help message is okay.~

Open to ideas on how to test Bulletproofs::splitIntoSets.

The basic idea is sound, demonstrated by the following snipped:

```c++
#include <stdexcept> 
  
int main() 
{ 
   int s = 10; 
    (A)  throw std::runtime_error("wrong number " + s); 
    (B)  throw std::runtime_error("wrong number " + std::to_string(s)); 

   return 0; 
}

```
-----
(A)
```bash
phil@crn ~/projects/tmp $ clang++ runerr.cpp  
runerr.cpp:6:46: warning: adding 'int' to a string does not append to the string [-Wstring-plus-int] 
   throw std::runtime_error("wrong number " + s); 
                            ~~~~~~~~~~~~~~~~^~~ 
runerr.cpp:6:46: note: use array indexing to silence this warning 
   throw std::runtime_error("wrong number " + s); 
                                            ^ 
                            &               [  ] 
1 warning generated. 
phil@crn ~/projects/tmp $ ./a.out  
terminate called after throwing an instance of 'std::runtime_error' 
 what():  er  
Aborted
```

-----
(B)
```c++
.phil@crn ~/projects/tmp $ ./a.out 
terminate called after throwing an instance of 'std::runtime_error'
  what():  wrong number 10
Aborted
```

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -1024,7 +1024,7 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
             "The third optional argument (may be null) is an array of base58-encoded private\n"
             "keys that, if given, will be the only keys used to sign the transaction.\n"
 #ifdef ENABLE_WALLET
-            + (!gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET)) ? HelpRequiringPassphrase(pwallet) + "\n" : ""
+            + HelpRequiringPassphrase(pwallet) + "\n"
 #endif
             "\nArguments:\n"
             "1. \"hexstring\"     (string, required) The transaction hex string\n"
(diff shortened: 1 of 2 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -1024,7 +1024,7 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
             "The third optional argument (may be null) is an array of base58-encoded private\n"
             "keys that, if given, will be the only keys used to sign the transaction.\n"
 #ifdef ENABLE_WALLET
-            + (!gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET)) ? HelpRequiringPassphrase(pwallet) + "\n" : ""
+            + HelpRequiringPassphrase(pwallet) + "\n"
```
mimirmim, line 1027: Isn't working, missing the end.

PR #798, opened 2020-06-26
[GUI] Fix RingCT Coin Control for minted change

CaveSpectre11:
### Problem
Coin Control does not display all RingCT TXOs.

### Root Cause
An errant check in the Coin Control code was removing RingCT transactions from the coin control list that were zerocoin mints.  This resulted in RingCT change from a minting transaction from being listed in coin control.

### Solution
Remove the zerocoinmint check in the coin control check.

### Testing
Test against all coin types (basecoin, stealth, ringct) after minting from each coin type [including using the 'true' flag to mint from basecoin].  After all is mature, compare total balance for each coin type in coin control (by using 'select all') against the balance from `getbalances` and ensure they match.

### DIFF
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -913,10 +913,6 @@ void CoinControlDialog::updateView(int nCoinType)
                 if (out.nType != nCoinType)
                     continue;
 
-                //Also need to filter out zerocoinmints
-                if (txRef->vpout[i]->IsZerocoinMint())
-                    continue;
-
                 //Don't display zero confirmed outputs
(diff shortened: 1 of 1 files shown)


seanPhill:
ACK.
This tested nicely! Coin Control's RingCT total equalled exactly getbalances' RingCT, and for a more demanding test I then spent and split every CT utxo via the GUI spend screen, which sent them all to RingCT. Even while there were a decent number of immature RingCT utxos the total RingCT in Coin Control and the total in getbalances was still exactly the same.

Six RingCT utxos in the first case and 26 RingCT utxos in the second case.

PR #799, opened 2020-06-26
[Trivial] Cleanup old bitcoin text

CaveSpectre11:
### Problem
Dump wallet output referenced Bitcoin

### Root Cause
leftover bitcoin reference

### Solution
Fix the reference and check the file for others (which there were)

### Testing
doublecheck:

- the help documentation for `importaddress`, `dumpwallet` and `importmulti`
- the contents of `dumpwallet`
- using invalid addresses in `importaddress` and `dumpprivkey`


seanPhill:
ACK.
This all tested fine.

PR #805, opened 2020-07-10
Update PoW Diff to use last blocktime

blondfrogs:
(description by CaveSpectre, take with a grain of salt)

### Problem
Difficulty calculations are based on the spacing between the last X blocks of the algorithm, and does not take into account if the next block for that algo is overdue.

### Solution
Refine logic to consider the spacing between the tip and the last algo block, if overdue, and adjust calculated spacing accordingly.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -563,7 +563,7 @@ public:
         consensus.BIP65Height = 581885; // 00000000007f6655f22f98e72ed80d8b06dc761d5da09df0fa1dc4be4f861eb6
         consensus.BIP66Height = 330776; // 000000002104c8c45e99a8853285a3b592602a3ccde2b832481da85e9e4ba182
         consensus.powLimit = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
-        consensus.powLimitRandomX = uint256S("000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
+        consensus.powLimitRandomX = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.powLimitProgPow = uint256S("0000000fffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.powLimitSha256 = uint256S("0000000fffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.nPowTargetSpacing = 120; // alternate PoW/PoS every one minute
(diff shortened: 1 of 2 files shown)


src/chainparams.cpp
```diff
@@ -563,7 +563,7 @@ class CDevNetParams : public CChainParams {
         consensus.BIP65Height = 581885; // 00000000007f6655f22f98e72ed80d8b06dc761d5da09df0fa1dc4be4f861eb6
         consensus.BIP66Height = 330776; // 000000002104c8c45e99a8853285a3b592602a3ccde2b832481da85e9e4ba182
         consensus.powLimit = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
-        consensus.powLimitRandomX = uint256S("000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
+        consensus.powLimitRandomX = uint256S("0000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.powLimitProgPow = uint256S("0000000fffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
         consensus.powLimitSha256 = uint256S("0000000fffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
```
CaveSpectre11, line 568: Given the current algorithm for adjusting the target:
```
    bnNew *= nActualTimespan;
    bnNew /= nTargetTimespan;
```
These, in combination with TargetSpacing define maximum `nDgwPastBlocks`.

I suggest one of 4 approaches (open to other ideas as well)
1) Note in CMainParams() that nDgwPastBlocks can't go over 127 (determined by 0xFFFF / 3 [the railed max] / 172 [the target spacing) = 127)
2) add an assert to check (max256 - powLimit) / 3 / targetspacing > nDgwPastBlocks
3) remove the overflow potential with bnNew *= (float)(actual/target)
4) remove the overflow potential with averaging the timespans (timespan/blocks)

Since it will work with the current settings; this is more of a nit

src/pow.cpp
```diff
@@ -217,20 +213,31 @@ unsigned int DarkGravityWave(const CBlockIndex* pindexLast, const Consensus::Par
...
     }
 
     int64_t nActualTimespan = pindexLastMatchingProof->GetBlockTime() - pindex->GetBlockTime();
     int64_t nTargetTimespan = params.nDgwPastBlocks * nPowSpacing;
 
+    // If we are calculating the diff for RandomX, Sha256, or ProgPow
+    // We want to take into account the latest block tip time into the calculation
+    int64_t nTipToLastMatchingProof = pindexLast->GetBlockTime() - pindexLastMatchingProof->GetBlockTime();
+    if (fNewPoW && nTipToLastMatchingProof > nPowSpacing) {
+        nActualTimespan += nTipToLastMatchingProof - nPowSpacing;
```
CaveSpectre11, line 238: took a minute but makes sense.  `-nPowSpacing` removes the need to add another nPowSpacing to the `nTargetTimespan`, and still covers if we're more than nPowSpacing overdue.

PR #806, opened 2020-07-14
[Build] Created Github Actions CI for Veil

CaveSpectre11:
**_With significant work by @akshaynexus in PR #783 .  Re-pushed due to rebasing challenges._**

### Problem
Travis limitations makes it's ability to serve as a Continuous Integration tool ineffective; needing frequent restarts (2-3 passes) to get through a full validation build via it's caching mechanism.

### Root Cause
Limitations of time cause us to have to exit to preserve the cache rather than timing out.  This makes it near impossible to get a clean build due to the length of time the build takes.

### Solution
Migrate to github actions, which have much higher limits and much better incorporated into github.  Initially this will build for the 64 bit platforms (x86 Linux, ARM Linux, Windows and OSX).  32 bit platforms will be added on a subsequent PR.

### Testing
As it's CI, it tests itself.  The completed images will need to be validated on their respective platforms, but as this is part of a migration effort; this PR will skip our traditional QA and be merged to begin use.  Validation of executables can be done in conjunction with it's creation of PR images.

### DIFF
--- a/.github/workflows/build-factory.yml
+++ b/.github/workflows/build-factory.yml
@@ -1,0 +1,207 @@
+# Copyright (c) 2018-2019 The Veil developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+name: Github Actions CI for Veil
+on: [push, pull_request]
+env:
+  SOURCE_ARTIFACT: source
+jobs:
+  create-source-distribution:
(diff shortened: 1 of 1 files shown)


CaveSpectre11:
A parallel change was pushed directly to master to test:

https://github.com/Veil-Project/veil/actions/runs/168878055

Those images can be tested for the final validation.

PR #807, opened 2020-07-14
[Build] Remove Travis-CI

CaveSpectre11:
_**This PR should not be merged until #806 has been fully reviewed **_

### Problem
Travis limitations makes it's ability to serve as a Continuous Integration tool ineffective; needing frequent restarts (2-3 passes) to get through a full validation build via it's caching mechanism.

### Root Cause
Limitations of time cause us to have to exit to preserve the cache rather than timing out.  This makes it near impossible to get a clean build due to the length of time the build takes.

### Solution
Remove Travis and replace with #806 github actions.

[This PR is also to test the automatic PR run of github actions]


CaveSpectre11:
> utACK
> 
> But just a note, Travis is still listed in the build checks below. Lint, etc., then skipped all our usual tests. I suppose we need to remove Travis in the project settings then?

@codeofalltrades found the settings to disable it for PRs.  I might do that reject this; so we still have it running on master until we can get that going on github actions.

PR #812, opened 2020-07-29
[PoW] Reduce overflow risk

CaveSpectre11:
### Problem
The `nActualTimeSpan` is at risk of overflowing if parameters are changed (hash limit, nDgwPastBlocks, etc).  While the Actual Time Span is railed to 3x of the Target, when you multiply the difficulty hash by the number of seconds between the first and last in the series, it can easily exceed the uint256 boundary if parameters aren't carefully calculated during changes in the future.

### Solution
Instead of multiplying the bnNew by the actual; which can be exceeded with parameter changes... we will divide the Actual by the Target, which have already been railed and result in a bnNew multiplier from 1/3 to 3; thus greatly reducing the chance of an inadvertent overflow


CaveSpectre11:
Rebased to bring in #811

PR #822, opened 2020-08-11
[Miner] Move miner started event message inside the actual mining code

codeofalltrades:
### Problem ###
The event log message "Veil Miner started" is written to the debug.log before we check in the wallet is PoW or PoS mining.
```
2020-08-07T03:15:49Z ThreadStagingBlockProcessing() start
2020-08-07T03:15:49Z ThreadStakeMiner() start
2020-08-07T03:15:49Z Veil Miner started
2020-08-07T03:15:49Z GUI: Platform customization: "other"
```

### Root Cause ###
Log message is at the first line of BitcoinMiner()

### Solution ###
Move the code into the generate bitcoin and generate stake loop.

### Unit Testing Results ###
1. Start the wallet. 
2. Turn on staking
3. Check the event log
```
2020-08-11T03:49:29Z ThreadStagingBlockProcessing() start
2020-08-11T03:49:29Z ThreadStakeMiner() start
2020-08-11T03:49:29Z Veil PoS miner started
2020-08-11T03:49:29Z GUI: Platform customization: "other"
```

1. Start the wallet. 
2. Go to the console, enter `generatecontinuous true 1`
3. Check the event log
```
2020-08-11T04:51:29Z ThreadStagingBlockProcessing() start
2020-08-11T04:51:29Z ThreadStakeMiner() start
2020-08-11T04:51:29Z Veil PoW miner started
2020-08-11T04:51:29Z GUI: Platform customization: "other"
```

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -842,8 +842,6 @@ static arith_uint256 nHashes = 0;
 static int32_t nTimeStart = 0;
 
 void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfStake = false, bool fProofOfFullNode = false) {
-    LogPrintf("Veil Miner started\n");
-
     unsigned int nExtraNonce = 0;
     static const int nInnerLoopCount = 0x010000;
     static int nStakeHashesLast = 0;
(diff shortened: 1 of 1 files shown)


CaveSpectre11:
I need to look a little deeper but I'm not sure this will do what you want it to.  I thought the while loop cycles every block, so that message will be spit out repetitively?  Since `ThreadBitcoinMiner()` is fired off from `GenerateBitcoins()`, and `GenerateBitcoins()` populates `fGenerateBitcoins` with `fGenerate`.... and doesn't start the threads if `!fGenerate` or `nThreads == 0`... if you're in ThreadBitcoinMiner and fProofOfStake == false; then you are going into that while loop until you stop mining (e.g. the message is fine outside the while loop).

Likewise, the ThreadStakeMiner() thread isn't being started unless your wallet is enabled (see init.cpp); so that's the actual worker thread.

So I would think it would be best outside the while loop (while the distinction between PoW and PoS is a sound change). I need to double check to make sure I'm recalling that code properly when I get to my system; as it looks like if you lock your wallet you'll start seeing that "Veil PoS miner started" every 5 seconds until you unlock)

PR #827, opened 2020-08-15
1.1 algo adjust

CaveSpectre11:
Note that this also contains the #826; I didn't bother to back out that change.
**_Important Note: This requires devnet to be restarted as it's going to change the difficulty calculations._**

### Problem
RandomX blocks are exceedingly difficult to mine.

### Root Cause
Dark Gravity Wave is designed to work with a single algorithm.  We started realizing this with #805 when a certain algorithm isn't mined, the difficulty stayed at the previous average.  This isn't a problem with Dark Gravity Wave since this would mean the chain is stuck; but with Veil we will keep motoring on with other algos.  So #805 added a way for the difficulty to drop if there's major changes in the mining power for a certain algo.

However the railing that Dark Gravity Wave has to prevent massive shifts in difficulty posed a new problem when there's significant shifts in mining power.  That algorithm will only reduce the difficulty by a factor of 3, regardless of how overdue the next block of that algo is.  Once it finds a block at that reduced difficulty, it will adjust it again; but if the adjusted difficulty is too hard for the current miners, it will only move more once a block is mined, and not continue towards minimum difficulty.

### Solution
Some major rework of the Dark Gravity Wave algorithm for our uses.   Firstly, another problem was realized when digging into the algorithm.  If there is a big gap where no blocks of a certain algorithm are being mined, that would lead to a situation where the algorithm can massively overshoot the necessary difficulty.   In simplified terms, assume the target time between blocks of a certain algo is 10 minutes; over 30 blocks this target time is 300 minutes.  Now assume it's not mined for a while, and the actual time between the mined blocks is still 300, but it's been 600 minutes since it was last mined, the 'actual time' is calculated to be 900 and it needs to reduce the difficulty.  So it calculates to be 1/3 of the previous difficulty [300/900].  Now a block is mined at that difficulty level.  The first block in the series falls away (900 - 10 = 890),  and the new block is used in the calculation [300/890].  So it adjusts the difficulty again.

In the above situation, the algorithm now won't adjust the difficulty until we're overdue again.  If the adjustment gets a block, we could be getting another block at the right time spacing, so we don't want to use the 30 block series in that case.  We will wait until we're actually overdue before we make any more adjustments.

Likewise when we're increasing the difficulty, we don't want to end up doing the inverse.  Lets say the separation has been 9 minutes instead of 10, so instead of 300 we have a separation of 270 so we increase the difficulty 300/270 and we hit the 10 minute spacing.  270 - 9 + 10 [dropping the oldest block and adding the newest block] is 300/271, so we jack the difficulty again.  Now the separation is 12.   271-9+12 = 273.  We jack it again, now the separation is 20.  271-9+20 = 282.  Now we keep jacking the difficulty until we get what we think is right, and in 30 blocks, we finally see that we way overshot and have to start reducing the difficulty.  So once we make an adjustment, we won't continue to make adjustments once the block is overdue.  We will leave it alone until the spacing catches up, or we see it's still too easy.

Onto the problem at hand.  Now instead of limiting the adjustment if it's gone a while since it's last been mined, we will continue to adjust until we hit the minimum difficulty.  This posed a risk to overflowing the target, and setting a target that's far more difficult then it's supposed to be, when you're trying to reduce the difficulty.   So an overflow check has been put into place, and if it does overflow, the difficulty is set to the minimum.  Testing of this algorithm actually revealed the true issue behind the observed problem.  multiplying the current difficulty times 3 _actually_ was overrunning the randomX minimum difficulty.  which could only handle a multiple of 2.  So the algorithm now will fine tune any adjustments under 2x; and over 2x it will fine tune to the multiple or the min difficulty if the multiple takes it over.

Other changes:
If we've changed to the new algorithm, it was still calculating the difficulty of the old algorithm.  That will get progressively longer since we have to walk back to the old algo blocks to calculate it.  Now if we're in the new algo, we'll not bother calculating the difficulty of the old algo.  Likewise getblockchaininfo will return -1 on algos that it's not calculating.

Similarly, if we're walking back and we cross over to the old algo range, we're not going to keep walking through the whole chain to determine that we don't have enough for the calculation.  In fact, with just one we'll have enough to make the calculation, since we know the time between the switch and the time since the one block was found, and the time from the last block to that one, and we can determine if it's pacing properly.  So we will use the number of blocks we found if there isn't enough.  This will often calculate out to min difficulty, but it will catch if min difficulty is way too low right at the starting point.

### Debug examples
[Debugging is currently available with debug=blockcreation, since we're going to need to reassess this all in a live environment after restarting the chain]

```
2020-08-15T21:10:04Z DarkGravityWave: nActualTimespan=7480, nTargetTimespan=7200 (PoS)
2020-08-15T21:10:04Z DarkGravityWave: PoS is behind but moving don't adjust anymore.
2020-08-15T21:10:04Z DarkGravityWave: nActualTimespan=790472, nTargetTimespan=36000 (RandomX)
2020-08-15T21:10:04Z DarkGravityWave: Multiplier 21 for RandomX overflowed.  Setting Min Difficulty.
2020-08-15T21:10:04Z DarkGravityWave: nActualTimespan=253289, nTargetTimespan=10320 (ProgPow)
2020-08-15T21:10:04Z DarkGravityWave: Adjusting ProgPow: old target: 0000000fffff0000000000000000000000000000000000000000000000000000
2020-08-15T21:10:04Z DarkGravityWave: Adjusting ProgPow: new target: 0000000fffffffffffffffffffffffffffffffffffffffffffffffffffffffff
2020-08-15T21:10:04Z DarkGravityWave: nActualTimespan=85526, nTargetTimespan=72000 (Sha256d)
2020-08-15T21:10:04Z DarkGravityWave: Adjusting Sha256d: old target: 00000006dd95b333333333333333333333333333333333333333333333333325
2020-08-15T21:10:04Z DarkGravityWave: Adjusting Sha256d: new target: 0000000827c43f6915ccc1262f0ef920b8f50825804bbb0151dfde80fa7e3f60
```
After syncing to the latest blocks
```
2020-08-15T21:22:01Z DarkGravityWave: nActualTimespan=6461, nTargetTimespan=7200 (PoS)
2020-08-15T21:22:01Z DarkGravityWave: Adjusting PoS: old target: 000000000000136352088888888888888888888888888888888888888888887c
2020-08-15T21:22:01Z DarkGravityWave: Adjusting PoS: new target: 0000000000001165e4080356652fb5f7f5af245541ea4e6e49e134430d93d5c8
2020-08-15T21:22:01Z SelectStakeCoins: FOUND 17 STAKABLE ZEROCOINS
2020-08-15T21:22:01Z BitcoinMiner: Tried 1020 stake hashes for block 26826 last=1597526551
2020-08-15T21:22:13Z New outbound peer connected: version: 70028, blocks=26825, peer=2
2020-08-15T21:22:14Z DarkGravityWave: nActualTimespan=6461, nTargetTimespan=7200 (PoS)
2020-08-15T21:22:14Z DarkGravityWave: Adjusting PoS: old target: 000000000000136352088888888888888888888888888888888888888888887c
2020-08-15T21:22:14Z DarkGravityWave: Adjusting PoS: new target: 0000000000001165e4080356652fb5f7f5af245541ea4e6e49e134430d93d5c8
2020-08-15T21:22:14Z DarkGravityWave: nActualTimespan=806137, nTargetTimespan=36000 (RandomX)
2020-08-15T21:22:14Z DarkGravityWave: Multiplier 22 for RandomX overflowed.  Setting Min Difficulty.
2020-08-15T21:22:14Z DarkGravityWave: nActualTimespan=268954, nTargetTimespan=10320 (ProgPow)
2020-08-15T21:22:14Z DarkGravityWave: Adjusting ProgPow: old target: 0000000fffff0000000000000000000000000000000000000000000000000000
2020-08-15T21:22:14Z DarkGravityWave: Adjusting ProgPow: new target: 0000000fffffffffffffffffffffffffffffffffffffffffffffffffffffffff
2020-08-15T21:22:14Z DarkGravityWave: nActualTimespan=88373, nTargetTimespan=72000 (Sha256d)
2020-08-15T21:22:14Z DarkGravityWave: Sha256d is behind but moving don't adjust anymore.
```


CaveSpectre11:
rebase with current 1.1.0 to address conflict

PR #833, opened 2020-08-21
[RPC] add times to getchainalgostats

CaveSpectre11:
### Problem
Get algo stats needed an indication of what time period the 1440 blocks were generated

### Solution
Add fields for start, finish and delta time.  start and finish is epoch time; delta is minutes.  Ideally it will be 1440.

```
spectre@octo:~/git-repos/veil/src$ veil-cli -regtest getchainalgostats
{
  "start": 1597795945,
  "finish": 1597977109,
  "period": 3019,
  "pos": 510,
  "progpow": 546,
  "randomx": 290,
  "sha256d": 94,
  "x16rt": 0
}
```
   "start" :   (numeric) Epoch time from the blcok 1440 blocks ago
   "finish" :  (numeric) Epoch time of the latest block
   "period" :  (numeric) Time, in minutes, the previous 1440 blocks spanned
```

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -244,6 +244,9 @@ static UniValue getchainalgostats(const JSONRPCRequest& request)
             "\nReturns the blocks found by each algo over that previous 1440 blocks.\n"
             "\nResult:\n"
                 "{ \n"
+                "   \"start\" :   (numeric) Epoch time from the blcok 1440 blocks ago \n"
+                "   \"finish\" :  (numeric) Epoch time of the latest block \n"
+                "   \"period\" :  (numeric) Time, in minutes, the previous 1440 blocks spanned\n"
                 "   \"pos\" :     (numeric) The number of PoS blocks found in the previous 1440 blocks \n"
                 "   \"progpow\":  (numeric) The number of ProgPow blocks found in the previous 1440 blocks \n"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/rpc/blockchain.cpp [hunk 1]
```diff
@@ -244,6 +244,9 @@ static UniValue getchainalgostats(const JSONRPCRequest& request)
             "\nReturns the blocks found by each algo over that previous 1440 blocks.\n"
             "\nResult:\n"
                 "{ \n"
+                "   \"start\" :   (numeric) Epoch time from the blcok 1440 blocks ago \n"
```
codeofalltrades, line 247: blcok should be block

src/rpc/blockchain.cpp
```diff
@@ -284,6 +287,10 @@ static UniValue getchainalgostats(const JSONRPCRequest& request)
     }
 
     UniValue obj(UniValue::VOBJ);
+
+    obj.pushKV("start", (uint64_t)pindex->nTime);
+    obj.pushKV("finish", (uint64_t)chainActive.Tip()->nTime);
+    obj.pushKV("period", (uint64_t)(chainActive.Tip()->nTime - pindex->nTime)/60);
```
codeofalltrades, line 293: NIT now I really want start and end block. It would make a graph I want to build for veil-stats much easier.

PR #840, opened 2020-08-27
[PoS][PoW] Refine PoS to manage overall chain spacing

CaveSpectre11:
**_This PR will require another restart of devnet._**

### Problem
When there isn't enough hash power to maintain desired spacing for the different PoW algorithms, the block spacing tends to extend longer than 60 seconds per block.

### Root Cause
All the algos, PoS included, manage their difficulty adjustments based on their designed spacing.  ProgPow targets one block every 172 seconds; RandomX targets one block every 600 seconds and Sha256 targets one block every 1200 seconds.  This all together accounts for one PoW block every 120 seconds.  Combining that with PoS one block every 120 seconds; you end up with a block every 60 seconds.

However, if any of the algorithms aren't meeting their designated block creation rate; there isn't any adjustments made to other algorithms.  So taking it to an extreme, with zero mining, the chain will adjust to target an average of one PoS block every 120 seconds, instead of every 60.

### Solution
Create a special case for ProofOfStake that doesn't look at it's own spacing, but rather looks at the overall spacing of the last 120 blocks (instead of the individual algorithms analysis of the last 60 blocks).  It then adjusts the PoS difficulty based on how close the last 120 blocks are to 60 seconds per block.  This essentially makes PoS the master spacing keeper for all the block spacing, and creates a situation where PoS will fill in the gaps if hashing power drops away suddenly.

### Additional changes
**_Difficulty Adjustment Capping_**
Even with a 8GB quadcore CPU, the SHA256 difficulty increases significantly from the minimum difficulty; not to mention the rapid increase in PoS difficulty.  This results in several hours of adjustments before the difficulty is raised to a point that the algorithm isn't running ahead.  Given the adjustments made in #827 in regards to waiting to see the result before making further adjustments, the limiting of the difficulty increases to just 3x does not appear to be necessary.  That limitation was removed so that the difficulty reacts to the current environment much quicker then previously.  This should prove to have a significant stabilization of chain timing upon switchover when the difficulties will be at their minimum, and could have taken several days with real hardware mining before the difficulty is raised to a point that the spacing is stabilized.

**_Regtest Stability_**
Regtest allowed light zerocoin protocol before enough blocks existed to support the hashing algorithm; thus causing a SegFault when the blocks needed can't be found.  A check was added to make sure that the block exists, as well as an increase in the switching point made to ensure Light zerocoin doesn't start until there's enough blocks to support it (100 + [1++10]*6) == 100 + 55*6 = 430 blocks.  So the switchover is set to 500.  This problem only existed with regtest.

Additionally regtest difficulties were changed to align closer to the other chains, to better test against the primary change in this PR.  The SHA256 difficulty was not adjusted higher due to another issue... the miner doesn't recognize it's solving for a genesis block that needs to be X16rt, and so it tries to mine it as one of the new protocols.  This "bug" won't exist with the fork on an existing chain, so it's only an issue with starting up a new chain; without enough hashing power to easily meet the SHA256 minimum difficulty.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -876,6 +876,24 @@ const CChainParams &Params() {
     return *globalChainParams;
 }
 
+int64_t CChainParams::GetDwgPastBlocks(int nPowType, bool fProofOfStake) const {
+    if (fProofOfStake)
+        return consensus.nDgwPastBlocks * 2; // count twice as many blocks
+    return consensus.nDgwPastBlocks;
+}
+
+int64_t CChainParams::GetTargetSpacing(int nPoWType, bool fProofOfStake) const {
+    if (nPoWType & CBlockHeader::PROGPOW_BLOCK)
+        return consensus.nProgPowTargetSpacing;
+    if (nPoWType & CBlockHeader::RANDOMX_BLOCK)
+        return consensus.nRandomXTargetSpacing;
+    if (nPoWType & CBlockHeader::SHA256D_BLOCK)
+        return consensus.nSha256DTargetSpacing;
+    if (fProofOfStake)
+        return consensus.nPowTargetSpacing/2; // Special case - actual block spacing
+    return consensus.nPowTargetSpacing;
+}
+
 std::unique_ptr<CChainParams> CreateChainParams(const std::string& chain)
 {
     if (chain == CBaseChainParams::MAIN)
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -196,7 +196,7 @@ public:
         consensus.BIP65Height = 388381; // 000000000000000004c2b624ed5d7756c508d90fd0da2c7c679febfa6c4735f0
         consensus.BIP66Height = 363725; // 00000000000000000379eaa19dce8c9b722d46ae6a57c2f1a988119488b50931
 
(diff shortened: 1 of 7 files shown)


CaveSpectre11:
> I am pretty sure just changing the the DGW nPastBlocks will break syncing :)

Yes; excellent catch.  The PoW algos are handled elsewhere, but the PoS does need to be tied to the fork time too.  On a quick glance, I think adding "and after fork" to the three "if proof of stake" (in the two function calls, and the different calculation method) it should function properly.   I'll look again and give it a try (and I just realized the wallet should function on mainnet, so that will be a good test that this [or other things] won't break before we put it up on testnet.

CaveSpectre11:
Reworked for backwards compatibility; and tested against mainnet (although not a full re-synch, only a few days synched).  The change in `VERSIONBITS_TOP_BITS` disabled all the soft forks (including V4 zerocoin) so that was the primary issue that hung me up.

@blondfrogs definitely take a look at see if there's an issue with switching the TOP_BITS back to 0x2000000 [which is what's needed for BIP9 as best as I could tell].

codeofalltrades (approved):
utACK 6b802f3d3fe3b2d24228f684a28c1331d4d55bb6

CaveSpectre11:
Running on mainnet for several days; plus running on devnet.   Merging.

Issue #659, opened 2019-08-27
Set unlimited wallet unlock from cli unlocktime to 0


Veil-Grumpy:
Instead of using 99999, it's much easier to set it to 0 for unlimited time.

From example:
walletpassphrase YOURWALLETPASSWORD true/false 999999

to example:
walletpassphrase YOURWALLETPASSWORD false 0

seanPhill:
Maybe a minimum time of 60 or 30 seconds? Gotta work fast to do something in less seconds than that. Zero seconds is pretty short. ;)

Veil-Grumpy:
I mean instead of having to type 9999999 just a 0 would do the same trick.

PR #844, opened 2020-09-05
[RPC] Let 0 be max walletpassphrase unlock

CaveSpectre11:
### Problem
Wallet unlock for 0 seconds is kinda useless

### Solution
Per suggestion from #659 -> have '0' set the max unlock


codeofalltrades (approved):
utACK b60e9ca6d25c6ce89209fb53d806a9c5cf88b465

PR #846, opened 2020-09-11
[PoW][RPC] Correct Sha256 Nonce incrementation

CaveSpectre11:
### Problem
Every pass through `generateBlocks` loop was trying the same hash 65535 times.

### Root Cause
The Sha256 hashing serialization algorithm uses nNonce64; but the `generate` RPC call was incrementing `nNonce`, so it was trying all with the same value of nNonce64.

### Solution
Change the code to increment nNonce64; Also check so we can detect if an nNonce64 loop expired without finding a solution [also effects progPow] so that it doesn't try to submit a bad block.


codeofalltrades (approved):
utACK 1f00d3aa1732c4117756ae1a19c854ed1a4df312

PR #849, opened 2020-09-27
[PoS] Fine tune PoS difficulty adjustments

CaveSpectre11:
### Problem
PoS blocks were falling behind their allotted slots due to increases in ProgPow hash power.

### Root Cause
A few problems existed.  The first was that the desired timespan for the PoS difficulty adjustments was based on the 120th PoS block back to the current tip; but the actual was calculated from the 120th PoS block back to the latest PoS block.  So if the latest block was not PoS, then the actual timespan in the calculation would be shorter then it actually was, and therefore the PoS difficulty wouldn't adjust where it should adjust to.

Secondly, the PoS difficulty was being based on a fixed spacing value, when it's actually dynamic.  When all PoW algos are running with enough hash power to maintain their spacing, then the PoS separation should be 120 seconds between blocks.  However, when algos hit their minimum difficulty because there's not enough power, PoS should take over, as the intent is that the chain can run without proof of work at all.  Without proof of work, then the spacing of PoS should be 60 seconds between blocks.

Lastly, the "overdue" calculation didn't take into account if the tip of the chain is the algo being checked; it was the time between the last of the algo and the tip.  If the tip is the last algo, then it can never be overdue.

### Solutions
First was to simply fix the actual timespan to be from the 120th PoS block back to the current tip; not the last PoS block.  This  now calculates the actual timespan and desired timespan over the same number of bocks.

The third issue is solved by keeping track of both the latest block of the algo, and one back from the latest.  This way if the chain tip is the algo being checked, then the overdue determination is made on the spacing between those last two blocks of the algo.

The second issue was the most complex.  This is solved by determining how many PoW blocks in the series that PoS difficulty is calculated on.  The expectation is 50% is PoS and 50% is PoW; thus making the desired PoS spacing to be 120 seconds.  The individual PoW algos difficulty is independently controlled by the spacing between it's own blocks.  The PoS spacing is determined by the overall spacing between all the blocks; targeting 60 between blocks.    This made it difficult to determine what the appropriate "expected" spacing for PoS is, as the expected PoS spacing is dynamic.   Since both the PoW spacing is independently controlled, and the overall spacing is controlled, the number of PoW blocks in the series used for calculating the PoS difficulty could be used to determine what the actual ratio is.

The logic now becomes determining the number of PoW blocks in the series, and the expected number of PoW blocks in the series.  That ratio is then used to adjust the expected PoS spacing.  60 + 60 * actual/expected.   For clarity of example, we will assume the number of PoS blocks counted in it's series is 90 (it's actually 120).  On a perfectly spaced chain, 45 PoS blocks will be contained in 90 blocks, with 45 PoW blocks.  In that situation, the PoS expected spacing is 60 + 60 * 45/45; or 120.  However if there's 50 PoS blocks, and 40 PoW blocks, then 60 + 60 * 40/45 = 60 + 53 = 113 seconds between PoS blocks.  Taking it to an extreme, if there is no mining; then 60 + 60 * 0/45 = 60 seconds per PoS block.

Real world examples:
```
2020-09-27T23:34:50Z DarkGravityWave: Blocks Measured=124, PoS Blocks=120, ActualPoWBlocks=4, ExpectedPoWBlocks=62, nPosSpacing=63 Overdue=5 (PoS)
2020-09-27T23:34:50Z DarkGravityWave: nActualTimespan=9838, nTargetTimespan=7380 Overdue=5 (PoS)
2020-09-27T23:34:50Z DarkGravityWave: Adjusting PoS: old target: 00031b99ccccccccccccccccccccccccccccccccccccccccccccccccccccccaf
2020-09-27T23:34:50Z DarkGravityWave: Adjusting PoS: new target: 00042495deda10984bb42f65eda10984bb42f65eda10984bb42f65eda1098493
```
In the example above; PoS is running behind [actual timespan of the series is 9838, and the target is 7380], so it should adjust the difficulty.    Of the last 124 blocks, 4 were PoW, but 62 were expected.  60 + 60 * 4/62 = 63.  The difference between the last two blocks [both PoS] was 68 seconds [not shown above, but it's 2020-09-27T23:32:08Z to 2020-09-27T23:33:16Z] ; so therefore it was 5 seconds overdue, so we apply the difficulty adjustment and increase the target mask needed for PoS [thus reducing the difficulty].

The next PoS block comes in at 2020-09-27T23:34:03Z.  This is 47 seconds from the previous block.

```
2020-09-27T23:34:53Z DarkGravityWave: Blocks Measured=124, PoS Blocks=120, ActualPoWBlocks=4, ExpectedPoWBlocks=62, nPosSpacing=63 Overdue=-16 (PoS)
2020-09-27T23:34:53Z DarkGravityWave: nActualTimespan=9844, nTargetTimespan=7380 Overdue=-16 (PoS)
2020-09-27T23:34:53Z DarkGravityWave: PoS is behind but moving don't adjust anymore.
```

The timespan is still behind, but we got a block in less then the expected 63 seconds [16 seconds early], so we're not going to adjust the difficulty anymore.

```
2020-09-27T23:35:44Z UpdateTip: new best=5ef8b0bfe25ff13e5960770eb024d92c39e4d603e56901ecb0467247d28f5b93 height=482 type=PoS version=0x38000001 tx=1071 date='2020-09-27T23:34:55Z' cache=0.0MiB(90txo)
2020-09-27T23:35:47Z DarkGravityWave: Blocks Measured=124, PoS Blocks=120, ActualPoWBlocks=4, ExpectedPoWBlocks=62, nPosSpacing=63 Overdue=-11 (PoS)
2020-09-27T23:35:47Z DarkGravityWave: nActualTimespan=9839, nTargetTimespan=7380 Overdue=-11 (PoS)
2020-09-27T23:35:47Z DarkGravityWave: PoS is behind but moving don't adjust anymore.
```
The next block comes in after 51 seconds.  We're still behind, but we're still moving faster than expected; so we again don't adjust the difficulty anymore.

This now 'waves' back and forth; adjusting when the blocks get behind, and leaving the adjustment alone when the blocks are ahead of schedule.

enable logging of 'blockcreation' to watch this occur in your debug log.

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -229,6 +229,7 @@ unsigned int DarkGravityWave(const CBlockIndex* pindexLast, const Consensus::Par
     const CBlockIndex *pindex = pindexLast;
     const CBlockIndex *pStart = pindexLast;
     const CBlockIndex* pindexLastMatchingProof = nullptr;
+    const CBlockIndex* pindexOneBack = nullptr;
     arith_uint256 bnPastTargetAvg = 0;
 
     unsigned int nCountBlocks = 0;
@@ -269,6 +270,9 @@ unsigned int DarkGravityWave(const CBlockIndex* pindexLast, const Consensus::Par
         } else if (!pindexLastMatchingProof) {
             // save the tip block for the proof
             pindexLastMatchingProof = pindex;
+        } else if (!pindexOneBack) {
+            // save the second one back in case the tip is the mined one
+            pindexOneBack = pindex;
         }
 
         arith_uint256 bnTarget = arith_uint256().SetCompact(pindex->nBits);
@@ -298,7 +302,7 @@ unsigned int DarkGravityWave(const CBlockIndex* pindexLast, const Consensus::Par
     if (fProofOfStake) {
         // Proof of Stake will judge the whole chain spacing, and adjust in case there isn't
         // enough mining to maintain optimum spacing
-        nActualTimespan = pindexLastMatchingProof->GetBlockTime() - pindex->GetBlockTime();
+        nActualTimespan = pStart->GetBlockTime() - pindex->GetBlockTime();
         nTargetTimespan =(pStart->nHeight - pindex->nHeight) * nPowSpacing;
     } else {
         nActualTimespan = pindexLastMatchingProof->GetBlockTime() - pindex->GetBlockTime();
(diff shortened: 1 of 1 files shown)


codeofalltrades (approved):
utACK c50ea71e0c64212af936624d980bcac555d01d30

PR #850, opened 2020-09-28
[DWG] Make sure OneBack exists

CaveSpectre11:
### Problem
segfault on first generated block after genesis block

### Root Cause
When there's only one block of a given algo, there is no "One Back", so the pIndexOneBack is null and thus calls to it's structure crash.

### Solution
If there's only one of any given algo, don't use oneback even if the tip is the current algo.


codeofalltrades (approved):
utACK dcdf296060054a6591dde57cf5691a7ec9258382

PR #854, opened 2020-10-03
[WIP][Build] Fix randomx build on 32b systems

CaveSpectre11:
### Problem
RandomX build issues on 32 bit systems

### Root Cause
64 bit code was pulled in, either arm or x86.  Re-work build to only build a64 for 64, and x86_64 as appropriate

### WIP
github action build of 32 bit linux x86 still fails due to dependency issues, but the others are working.  This may be merged as is and the last one [commented out] for later.

Note:  It's not been run yet with the commented out yml, so that might need a tweak


codeofalltrades (approved):
utACK
19eb939e381f55bf0f9a7b2f380e5469d8c5fd6b

PR #864, opened 2020-11-07
[RingCT Staking] Fix creation of stealth staking address on startup

CaveSpectre11:
### Problem
A Stealth Staking address isn't properly created when initially starting 1.2 on an existing chain.

### Root Cause
Only 1.2 generated wallets were tested; not transition from existing wallets on upgrade.

### Solution
Added `getstealthstakingaddress` RPC command to see what your created staking address is
Added logic to list unspent to check and display any coins in the staking address
Reworked whole creation and loading logic to handle transition from 1.1 wallet to 1.2 wallet, prior to this PR, it only worked if the wallet was generated with 1.2 and not encrypted.

### Testing
- Tested transition from 1.1 to 1.2
- Tested restart of 1.2 after transition
- Tested creation of 1.2 wallet
- Tested restart of 1.2 wallet after creation
- Tested transition of an encrypted wallet 1.1 to 1.2 [and restart]
- Tested transition of a 1.2 created wallet to encrypted wallet
- Tested `getstealthstakingaddress` through all
- **_Did not_** test rescanning report log output change

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -164,14 +164,21 @@ bool AnonWallet::Initialise(CExtKey* pExtMaster)
 
             //Load the stealth stake account
             idStakeAccount.SetNull();
+            idStakeAddress.SetNull();
             if (!wdb.ReadNamedExtKeyId("stealthstakeaccount", idStakeAccount)) {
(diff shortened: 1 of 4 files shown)


codeofalltrades (approved):
ACK 0898a3e4ba1719742646d15878bd9045dc78312e

image

PR #865, opened 2020-11-07
[GUI]: add a small indicator for the last block time on lower veil status bar

WetOne:
### Problem
A small improvement to the usability: the user often needs to see what the block time is on the current node, and it is very annoying having to open the debug console to extract this information.

### Root Cause
The value is not visible.

### Solution
Two labels have been added to the bottom status bar and connected the signal to update the block time.

### Bounty Payment Address
`sv1qqphsvuwhk29xcn2q9gdth9x73qpm8kcktmuxyd8szl50sgt2nt47egpqwnxr83hfj7s6fnec2jr0cv5yc7r6s7nvxhd9969qrgy4cgnznwewqqqauut5m`

Veil_Last_Block_In_Status_Bar2020-11-06 17-55-08

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -541,6 +541,7 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel)
         {
             walletFrame->setClientModel(_clientModel);
             balance->setClientModel(_clientModel);
+            veilStatusBar->setClientModel(_clientModel);
         }
 #endif // ENABLE_WALLET
         unitDisplayControl->setOptionsModel(_clientModel->getOptionsModel());
@@ -925,6 +926,7 @@ void BitcoinGUI::setNumBlocks(int count, const QDateTime& blockDate, double nVer
     }
 
     veilStatusBar->updateSyncIndicator(count);
+    veilStatusBar->setNumBlocks(blockDate);
 
     QString tooltip;
 
--- a/src/qt/veil/forms/veilstatusbar.ui
+++ b/src/qt/veil/forms/veilstatusbar.ui
@@ -91,7 +91,7 @@
        </widget>
       </item>
       <item>
-       <layout class="QHBoxLayout" name="horizontalLayout" stretch="0,0,0,0,0,0">
+       <layout class="QHBoxLayout" name="horizontalLayout" stretch="0,0,0,0,0,0,0,0,0">
         <property name="leftMargin">
          <number>9</number>
         </property>
@@ -188,12 +188,39 @@
          </widget>
         </item>
         <item>
-         <spacer name="horizontalSpacer_right">
+         <spacer name="horizontalSpacer_middle">
           <property name="orientation">
            <enum>Qt::Horizontal</enum>
           </property>
           <property name="sizeType">
            <enum>QSizePolicy::Expanding</enum>
+          </property>
+          <property name="sizeHint" stdset="0">
+           <size>
+            <width>20</width>
+            <height>20</height>
+           </size>
+          </property>
+         </spacer>
+        </item>
+        <item>
+         <widget class="QLabel" name="labelLastBlockTime_Head">
+          <property name="text">
+           <string> Last block time:</string>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QLabel" name="labelLastBlockTime_Content">
+          <property name="text">
+           <string>N/A</string>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <spacer name="horizontalSpacer_right">
+          <property name="orientation">
+           <enum>Qt::Horizontal</enum>
           </property>
           <property name="sizeHint" stdset="0">
            <size>
--- a/src/qt/veil/veilstatusbar.cpp
+++ b/src/qt/veil/veilstatusbar.cpp
@@ -6,6 +6,7 @@
 #include <qt/veil/forms/ui_veilstatusbar.h>
 
 #include <qt/bitcoingui.h>
+#include <qt/clientmodel.h>
 #include <qt/walletmodel.h>
 #include <qt/veil/qtutils.h>
 #include <iostream>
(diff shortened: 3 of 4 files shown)


CaveSpectre11:
@WetOne when submitting PRs, you're welcome to add a 
```
### Bounty Payment Address ##
`sv1`
```
To include your address for payment in the original PR.  Payment is made after code review, QA, and merging into the branch.

CaveSpectre11:
@WetOne how big of a deal would it be to flip the location; so "Height: " remains in the same place as it currently is, and the new stuff is in the space to the left.

image

Also, is the line something graphically intentional?

image

WetOne:
@CaveSpectre11 Moving the height should be easy - working it now.

I don't quite understand the line.  I did nothing that should have affected it.  possibly my display?

CaveSpectre11:
> I don't quite understand the line. I did nothing that should have affected it. possibly my display?

Definitely an artifact in the code; I see it on my screen too.

WetOne:
OK, I think I see what is going on.  Should have it ready tomorrow.

WetOne:
Looks like I got both.  How do I update into git?

Veil_Last_Block_In_Status_Bar_2020-11-07 19-37-39

CaveSpectre11:
> Looks like I got both. How do I update into git?
> 

Have the changes in your QtLastBlockTime branch; then commit the changes into that branch [so the branch will have 3 commits].  You would then "git rebase HEAD~2", which will bring up an editor with the two commits in it.  Go to the second and change `pick` to `fixup`, and exit the editor.  That will rebase the two changes into one.

When you push it, push it with `git push --force` which will know to replace the previous two commits [reflected in this PR] with the "new" two commits [the combined one and the original sinetek one].  It will automatically reflect in this PR [because the underlying branch in your fork repo changed].

WetOne:
"git rebase HEAD~2" resulted in "Current branch QtLastBlockTime is up to date."  

Did the forced push.

CaveSpectre11:
> "git rebase HEAD~2" resulted in "Current branch QtLastBlockTime is up to date."
> 
> Did the forced push.

My bad; should have been `git rebase -i HEAD~2`

WetOne:
### Bounty Payment Address
sv1qqphsvuwhk29xcn2q9gdth9x73qpm8kcktmuxyd8szl50sgt2nt47egpqwnxr83hfj7s6fnec2jr0cv5yc7r6s7nvxhd9969qrgy4cgnznwewqqqauut5m

CaveSpectre11 (dismissed):
ACK https://github.com/Veil-Project/veil/pull/865/commits/2beb3bbeb4e71377644a386e2048574016112bbb

codeofalltrades:
image
The staking slider is cut off on windows and linux. I didn't check macOS.
Also there is a new dash to the left of the staking slider.

image

seanPhill:
I'm getting a problem with the display of that data as you will see in the screenshots.
At first I couldn't make out what it was saying, but I realised later it was displaying "N/A" until it got data from peers.
Giving it more time in case the display text box would become completely visible, it didn't. It needs to be as clearly visible as it is when you click the middle of that bar (on "Synchronising with network ...") and display the Last Block Time that way.
pr865-01
pr865-02
pr865-03
pr865-04
pr865-05

seanPhill:
You can see also that it appears not to be using the same data source as the pop up syncing information, as that already has a displayable last block time while the one below only has "N/A".

Also, the popup window that works comes up by default and can be brought back up again by simply clicking on "Synchronising with network ...", so maybe this whole thing can be solved by just adding the text "--> last block time" after the ellipse.

seanPhill:
The above test is on MacOS Catalina 10.15.6 using the binaries downloaded from Github Actions.

CaveSpectre11:
> You can see also that it appears not to be using the same data source as the pop up syncing information, as that already has a displayable last block time while the one below only has "N/A".

Note that when it's showing "N/A", the pop up is downloading headers still, not blocks.   My guess is that it's only being populated by new blocks brought in, and no new blocks have come in yet since it's still doing headers?  Definitely something for WetOne to confirm [note that part of the code was from Sinetek's PR; WetOne was asked to move it to where it's supposed to be.

> 
> Also, the popup window that works comes up by default and can be brought back up again by simply clicking on "Synchronising with network ...", so maybe this whole thing can be solved by just adding the text "--> last block time" or something to that effect) after the ellipse.

Not sure we want to have it in different places when synchronizing vs. regular functional running of the code.  Remember this is supposed to be there all the time [so you don't have to dig it out from the "Debug Window->Advanced Options->Information.

But graphically I concur:
- Still has a strange line to the left of the Staking switch [the one that was to the right was cleared up]
- The top of the staking slider is clipped [Linux, MacOS].  The bottom of the date is clipped [MacOS].

@WetOne is there a new box around the whole status bar that's causing the clipping?

WetOne:
Looks like I fixed the clipping.  When I updated the bar there was a new layout which resulted the clipping.  This is what the update should look like.
Veil_Last_Block_In_Status_Bar_2020-11-13 14-49-54

WetOne:
Now, regarding the update to the Last Block Time in the status bar vs in the Modal Display:
CaveSpectre, you are correct that the N/A is only there until new blocks come in.

I see we have 2 options

1) The current implementation on the branch updates the status bar when signal numBlocksChanged is emitted (this is done from clientModel::BlockTipChanged).
2) If we want them to be the same we could change the implementation to be a function call rather than a SIGNAL-SLOT connection and call it from BitcoinGUI::setNumBlocks.

I prototyped the second.  There still seems to be a small difference between the popup and the status bar using this method and the N/A is still there until new blocks come in (really until setNumBlocks is called).   The small difference is probably due to some logic at the beginning of setNumBlocks.

The second option probably makes more sense as most of the updates to the status bar and modal overlay are both are done from BitcoinGUI::setNumBlocks.  Understand that I would probably remove most of Sinetek’s code as the SIGNAL-SLOT connection would not make sense with the second implementation.

@CaveSpectre11, @codeofalltrades, @seanPhill : your call as to which way you want this.

CaveSpectre11:
I would think the second implementation [as it makes more sense] would be desirable.  I'll have to defer to @seanPhill though, as he had the original expectation of them mirroring one another.

seanPhill:
I'm fine with whatever works, functionally and aesthetically.

CaveSpectre11 (approved):
ACK https://github.com/Veil-Project/veil/pull/865/commits/cc13532b53e3985cd096daa030770e7cf255e5a1

codeofalltrades (approved):
ACK cc13532b53e3985cd096daa030770e7cf255e5a1

seanPhill:
ACK
I have tested this one now (on MacOS Catalina), and the block date and time indicator looks fine now.
Screen Shot 2020-11-23 at 9 21 14 pm

CaveSpectre11:
10 hours @ 310 = 3100 Veil
fixed form use bonus = 100 Veil
3200 Veil paid

Thank you for your contribution to Veil!

PR #868, opened 2020-11-08
[RPC] add vout.n sequence number to decoderawtransaction

CaveSpectre11:
### Problem
When looking at raw transactions in the block chain, the vout sequence number is not shown

### Root Cause
Wasn't implemented

### Solution
Implement it

### Testing
Transactions now have a "vout.n" in vout sections, showing what the output sequence is [for use as input to raw transactions].
```
    {
      "type": "standard",
      "value": 345600.00000000,
      "valueSat": 34560000000000,
      "vout.n": 2,
      "scriptPubKey": {
        "asm": "OP_HASH160 4874bc3b94c4c85cf858670fef39d38fbe0d7b8b OP_EQUAL",
        "hex": "a9144874bc3b94c4c85cf858670fef39d38fbe0d7b8b87",
        "reqSigs": 1,
        "type": "scripthash",
        "addresses": [
          "38J8RGLetRUNEXycBMPg8oZqLt4bB9hCbt"
        ]
      }
```


codeofalltrades (approved):
ACK f20b416c7e85deb24e1e38e98eee87fb47dbeedc

PR #869, opened 2020-11-09
[PoW][RandomX] Display Hashspeed after processing

CaveSpectre11:
### Problem
RandomX Hashspeed was not being displayed in the logfile when using blockcreation debugging

### Root Cause
Technically it does; however only when you actually win a block or you get through the full 100,000 hashes without another block coming in.   The problem was, if you stopped computing because another block came in while you were computing, it would restart the hashing on the next block, before it reports the hash speed used.

### Solution
Move the hash debugging to before the check if the block was already found or not.  Also a couple other tweaks made; reduce the cs_nonce critical section to minimum needed, as well as remove the hardcoded inner loop count, as that inner loop count is also used for nonce spacing.  That way if the inner loop count is changed, the nonce separation between threads is consistent and doesn't leave gaps.

### Issue
Addresses Issue #866

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -1020,7 +1020,7 @@ void BitcoinRandomXMiner(std::shared_ptr<CReserveScript> coinbaseScript, int vm_
     LogPrintf("Veil RandomX Miner started\n");
 
     unsigned int nExtraNonce = 0;
-    static const int nInnerLoopCount = 100000;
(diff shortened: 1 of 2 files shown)


codeofalltrades (approved):
ACK 21aa6c459e7b06dafca3c468d7afc82dbbc4c1d2

PR #871, opened 2020-11-21
Optimize hashing of internal sha256d miner.

barrystyle:
Since the datahash is not likely to change much between blocks (other than
work refresh), we can calculate it once and reuse the value.

Average hashrate beforehand on 4core vps was 4mh, averages around 5.5-6mh with
blocks successfully solved on devnet.

### DIFF
--- a/src/primitives/block.cpp
+++ b/src/primitives/block.cpp
@@ -48,10 +48,20 @@ uint256 CBlockHeader::GetX16RTPoWHash(bool fSetVeilDataHashNull) const
 uint256 CBlockHeader::GetSha256DPoWHash() const
 {
     CSha256dDataInput input(*this);
-
     uint256 dataHash = SerializeHash(input);
-
     CSha256dInput sha256Final(*this, dataHash);
+    return SerializeHash(sha256Final);
+}
+
+uint256 CBlockHeader::GetSha256dMidstate() const
+{
+    CSha256dDataInput input(*this);
+    return SerializeHash(input);
+}
+
+uint256 CBlockHeader::GetSha256D(uint256& midState) const
+{
+    CSha256dInput sha256Final(*this, midState);
     return SerializeHash(sha256Final);
 }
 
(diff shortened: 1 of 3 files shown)


CaveSpectre11:
I need to run some tests to make sure there isn't changes to the hash during a run (doesn't look like it from code inspection).  While I'm doing that check:

`                uint256 localDataHash = pblock->GetDataHash();
`

This function needs to change, since it's specific to Sha256D and isn't a generic GetDataHash().

src/primitives/block.cpp
```diff
@@ -48,9 +48,19 @@ uint256 CBlockHeader::GetX16RTPoWHash(bool fSetVeilDataHashNull) const
...
     CSha256dDataInput input(*this);
-
     uint256 dataHash = SerializeHash(input);
+    CSha256dInput sha256Final(*this, dataHash);
+    return SerializeHash(sha256Final);
+}
+
+uint256 CBlockHeader::GetDataHash() const
+{
+    CSha256dDataInput input(*this);
```
CaveSpectre11, line 58: Since this is specific to Sha256; change the `GetDataHash()` to be reflective of that

barrystyle: `uint256 CBlockHeader::GetSha256dMidstate()` ?

CaveSpectre11: Sure; that works. Use your discretion; but just as long as it is identifiable as a sha256 function [I figured `GetSha256dDataHash()`, but whatever makes sense I'm good with.

codeofalltrades (approved):
ACK 9b07d64f8fa1e2ba42a828d8d02b335078a501e1
miner changes work. I ran it on the testnet for 24 hours.
sample block: http://testnet-explorer.codeofalltrades.com/block/c154ad36bffe1758e441fa2be2dfdb99bd6f30ec58b7742d167f711bb1e41590

src/primitives/block.cpp
barrystyle: amended in 5650b36

CaveSpectre11:
@barrystyle just `rebase -i HEAD~2' and fixup the second one so that it's only one commit, and we'll send it on to QA.

Also, add a stealth address.  We don't have a lot to offer, but we'll get you something for a bounty.

barrystyle:
Appreciated, but theres no need.. enjoy!

codeofalltrades (approved):
utACK b3603edbb9d03482fc44aff3d302ee4f14af8d44

CaveSpectre11 (approved):
https://github.com/Veil-Project/veil/pull/871/commits/b3603edbb9d03482fc44aff3d302ee4f14af8d44

Issue #692, opened 2019-11-22
GetAncestor pindexWalk->pprev assertion fail


mimirmim:
The assertion in `const CBlockIndex* CBlockIndex::GetAncestor(int)` at ` pindexWalk->pprev` fails it's assertion.

This is unfortunately unreliable to reproduce. Needs investigating into what had caused it.

This was done with v1.0.4.7, downloaded from GitHub Veil releases.

See backtrace:
```[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff3d5b700 (LWP 22025)]
[New Thread 0x7ffff1d9e700 (LWP 22026)]
[New Thread 0x7ffff0e51700 (LWP 22027)]
[New Thread 0x7fffe9ffd700 (LWP 22028)]
[New Thread 0x7fffe3e1e700 (LWP 22029)]
[New Thread 0x7fffe361d700 (LWP 22030)]
[New Thread 0x7fffe2e1c700 (LWP 22031)]
[New Thread 0x7fffe261b700 (LWP 22032)]
[New Thread 0x7fffe1e1a700 (LWP 22033)]
[New Thread 0x7fffe1619700 (LWP 22034)]
[New Thread 0x7fffe0e18700 (LWP 22035)]
veil-qt: chain.cpp:105: const CBlockIndex* CBlockIndex::GetAncestor(int) const: Assertion `pindexWalk->pprev' failed.
 
Thread 4 "QThread" received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff0e51700 (LWP 22027)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51  ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff5bc1801 in __GI_abort () at abort.c:79
#2  0x00007ffff5bb139a in __assert_fail_base (
    fmt=0x7ffff5d387d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n",
    assertion=assertion@entry=0x5555576f4170 "pindexWalk->pprev",
    file=file@entry=0x5555576741ab "chain.cpp", line=line@entry=105,
[...]

PR #872, opened 2020-11-22
[DB] Data error checks and frequency changes

CaveSpectre11:
### Problem
PindexWalk->pprev assert failures occur randomly on startup

### Root Cause
_**Note that root cause is not yet completely understood.  This is not a fix, but rather a mitigation that alleviates the risk of the problem occurring.**_

A corruption occurs occasionally in the writing of the block database, where an index no longer points to the correct block, but rather reads a different block off disk.  Usually the block read happens to be an orphan block.  It definitely is connected to writing of side chains and orphans, but the root cause is yet unknown.   This problem results in a valid block not being read from disk, so when it begins to connect the blocks together [for the staking algorithmic calculations], there is a block that doesn't have a valid previous [because it basically walks off the chain because there's a block missing.

### Mitigation
**_Note that this PR will not correct an already corrupt database; it will only minimize the occurrences.  Anyone currently with a corrupted database will have to recover their chain prior to being able to run this PR._**

Several changes were made.  Information is added to the debug log file to give an indication as to what the problem block was, when a problem block is detected on startup.  When the header information is disconnected from the block information, the block is no longer written to disk (This generally occurs with orphan blocks).  If a block is read when this occurs, it also is reported to the log file.

The frequency that blocks are written to disk has been changed to write a similar number of blocks at a time as Bitcoin.  Since veil's block creation is 10 times faster, the write timers have been changed to be 10% of what they were in Bitcoin.  This showed significant improvement to the occurrences of the corruption.  It's important to remind again that this is not a fix; there is still a lingering issue.  However over heavy test with two nodes, one of them running with the new write frequency and one without the write frequency; the one without the change found 63 instances of corrupted blocks in 1923 blocks.  The node with the change saw zero corruptions.

For that reason, this PR is being pushed to greatly reduce the occurrences that have become prevalent again as more people are in wallet mining and staking at the same time.  Issue #692 will remain open while this is continued to be worked over time, as attempts to correct the corruption are investigated, or root cause is continued to be found.  This PR however does reduce the urgency of the research.

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
+// Copyright (c) 2019-2020 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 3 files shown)


codeofalltrades (approved):
ACK 76abd68db4519adc97fb8000676c7ad514a579c2
I never received the error. I ran the wallet on intel amd and in a VM.

PR #873, opened 2020-11-22
[Trivial][RPC] Correct generatecontinuous help

CaveSpectre11:
### Problem
`generatecontinuous` help example didn't have the boolean [true false] in it.

### Solution
Fix the help


codeofalltrades (approved):
ACK 57d2963318fdd88ce85f74d782bfcf115810b4e4

Issue #253, opened 2019-01-10
Prune mode is incompatible with -txindex


pcartwright81:
Windows 10 wallet won't open and says "Prune mode is incompatible with -txindex"

![image](https://user-images.githubusercontent.com/2352477/50940259-d7f57f00-1445-11e9-8735-5e859972f390.png)

pcartwright81:
If anyone else has this issue fix the key in the registry under Computer\HKEY_CURRENT_USER\Software\Veil\Veil-Qt
You will see bPrune

ianGre:
You could take "txindex=1" out of your veil.conf, or 
turn off (in Options) Prune block storage. (Except that you can't, because it won't start.) or
edit your Registry with Regedit (as mentioned above). You might want to turn off Prune mode when you get it running.

Prune mode (keep only blocks interesting to your wallet) and
txindex (index all blocks) are certainly contradictory instructions.

pcartwright81:
I do not have a viel.conf and there does not seem to be an option in the GUI to turn off index all blocks.

ianGre:
Interesting! I'll give you some instructions to do the fix via the Windows Registry. Just a moment while I start up my Windows Virtual Machine. ...

ianGre:
I am surprised that pruning is turned on, as it was off by default with my installation. Also I did not find a registry entry for transaction indexing. So, the easiest way (to turn off pruning, assuming your wallet will not open, because of the error), is to open Notepad and paste the following lines into it, and save the file onto your Desktop as "Veil-pruneOff.reg" or whatever you like. And then right-click and select **Merge**, which should be the default action, if the filename ends in .reg and not.txt. 
```
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Software\Veil\Veil-Qt]
"bPrune"="false"
```

presstab:
txindex is turned on by default with the code. Interesting that prune is defaulting as on... it actually gets disabled in the code, as it is still under development.

seanPhill:
I haven't seen prune on by default in any of my installations.

seanPhill:
Thanks! ... Well, then the prune mode function had better turn off txindex when it turns on.

mimirmim:
Update on this issue as we have been investigating why all nodes are flagging as limited nodes, and not actually full nodes. This is related.

seanPhill:
[...]

PR #874, opened 2020-11-23
[Config] Ignore prune mode if txindex is set

CaveSpectre11:
### Problem
#253 - Prune mode incompatible with txindex.  
Having prune mode on can stick the user into being unable to start their wallet.  For Qt: they receive an error stating "Prune mode is Incompatible with -txindex"

### Root Cause
The error was a little heavy handed, and rejected the use of prune mode without txindex=0.  The part of this issue that makes it important to address is that you can inadvertently end up with bPrune set in your registry on windows; making it difficult to find and remove.  There is an additional issue that implies that uninstalling your wallet will not remove the registry entries; so getting bPrune=true in your registry would be quite challenging to clear out, and likely result in support tickets.

### Solution
**_Note that prune mode is unsupported and likely will result in corrupting your wallet.  This is not meant to claim that prune mode is functional._**

If Prune mode is set and txindex is not implicitly disabled [txindex=0], the wallet will not run.  This PR changes the prune mode from an error that shuts down the wallet, to a warning that ignores prune mode if txindex is enabled.   This warning will not show in a Windows pop-up, but rather in the log file and the terminal executing the daemon and/or qt.  It will also ignore the prune mode setting, changing it to '0' if it is set.

### Testing
combinations of txindex=[0|1], prune=[0|1] in your config file, command line, and bPrune=true in your registry [see issue #253 for registry details]

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1,7 +1,7 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2019 The Bitcoin Core developers
 // Copyright (c) 2015-2019 The PIVX developers
-// Copyright (c) 2018-2019 The Veil developers
+// Copyright (c) 2018-2020 The Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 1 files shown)


codeofalltrades (approved):
ACK 5d4965198d75d6758b06331175a58e4a9f43d286

Set the prune setting via the GUI.
tested with 
txindex=1, prune=1
txindex=0, prune=1
txindex=1, prune=0
txindex=0, prune=0

All the test passed.
OS Win10

Issue #340, opened 2019-01-28
getnetworkhashps seems to be unrelated to actual network hash.


seanPhill:
When checking getnetworkhashps the result seems to be nowhere near a believable value.

testnet
```
getnetworkhashps
17.08739622641509

getnetworkhashps
￼16.4808560197991
```
mainnet
```
getnetworkhashps
16.51547897196262

getnetworkhashps
16.7731179608773
```
That's crazy low, isn't it?
But the difficulty is high.

testnet
```
 "difficulty": 407361.8355413844,
```
mainnet
```
"difficulty": 1739806.144906988,
```

CaveSpectre11:
This actually isn't a bug.  It's a little misleading due to the fact that the difficulty checks happen to be done when the block was a PoS block rather than a PoW block.  The `getnetworkhashps` is an average per second over the last 120 blocks (without using parameters to change the time) and is based solely on the timestamp of the two blocks and the chainwork.  PoS uses a different mechanism for it's "work" target, so the chainwork is not going to accurately reflect the work.

What you end up with is a hash per second that doesn't really present any useful information, since it's a fruit salad of PoW apples vs. PoS oranges.  But mathematically, it's correct.

e.g.
PoS blocks 306022 to 306026 had difficulties over 44.6 mil; but the chain work went from 0x1964727a to 0x19648f67; a difference of 7405 over a period of 281 seconds.  7405/281 = 26.35

A more accurate reflection of the hashPS is calculated by looking at a stretch of PoW blocks.  e.g. 306015 to 306018.   difficulty at ~4k, and a hashPS of 11.71.

Comparatively; 169998 through 170000; which has a higher PoW difficulty, comes up with a hashPS of 13.54

So while it looks strange to see such low Hash PS with such high difficulties; what you're really seeing is a very disconnected average of PoS vs. PoW.

Rock-N-Troll:
So essentially the change you're suggesting is probably the best course of action for getting useful data out of this api call. I agree with that.

CaveSpectre11:
So you want the api call for getting the network hash per second to give you something more useful then the network hash per second?

Rock-N-Troll:
As you've mentioned, removing PoS values from the calculation would be a more accurate reflection of network hashPS

CaveSpectre11:
[...]

Issue #32, opened 2018-12-27
Difficulty wildly varying! Is reporting error or is it real?


ianGre:
The mining difficulty is varying wildly! Is this a reporting error or is it real?

getmininginfo -->
```
  "blocks": 7576,
  "currentblockweight": 4000,
  "currentblocktx": 0,
  "difficulty": 0.004190474607137011,
  "networkhashps": 568466053402.7554,```

```
"blocks": 7593,
  "currentblockweight": 5244,
  "currentblocktx": 1,
  "difficulty": 17916.17690491486,
  "networkhashps": 651041536223.8818,
  "pooledtx": 1,
  "chain": "test",```

```
  "blocks": 7595,
  "currentblockweight": 6488,
  "currentblocktx": 2,
  "difficulty": 0.004467099896173242,
  "networkhashps": 611342315891.5979,```

Is it normal that difficulty would adjust by an order of millions? (And back again?) Surely it should be the higher figure.

ianGre:
I have a theory. Please confirm.

Does the difficulty work out to a completely different number, depending on the algorithm for that block?

Secondly, are people getting banned by other peers after they have a problem getting the proper choice of algorithm? And then once you get one wrong, you never catch up?

presstab:
Proof of Stake and Proof of Work both have their own independent difficulty. More likely though is just a normal adjustment in difficulty. With so few on the network, it is easy to make diff change fast.

ianGre:
Thanks! Yes, I was just going to do a block by block comparison to see when it changes.

PR #876, opened 2020-12-05
[RPC] Reworked getnetworkhashps with tweaks to -mine and getmininginfo

CaveSpectre11:
[Problem]
#340 - `getnetworkhashps` seems to be unrelated to actual network hash
#32 - `getmininginfo` has vastly changing difficulties

[Root Cause]
As greatly discussed in #340, the `getnetworkhashps` never accurately reflected PoW vs. PoS; as if one of the blocks was a PoS block, and one was a PoW block, the average calculated would vary wildly.    Likewise, `getmininginfo` would only report on the last block, which gets even crazier with 4 different block types.

[Solution]
First was a need to get your current mining algorithm; this is a separate commit.  The original code was getting the mining algorithm only when generation was started; so the startup flag sat unused until that time.  This also had the side effect of not realizing you would be mining randomX, until you started mining, due to a typo in the `-mine` flag [e.g. sha256 instead of sha256d.  By changing the code to read the flag on startup, it opened the ability to sanity check it on startup; so users using an incorrect flag will now be told so.  This also opens the ability to switch mining algorithms without stopping the daemon [although that's for a future PR, as care is needed to make sure you're told to stop mining before switching algorithms].

```
% veild -mine=sha256
Error: Invalid mining algorithm: sha256
```

With that out of the way, `getnetworkhashps` was expanded to still function as would be expected, but instead of just randomly using whatever blocks it found; it now focuses on only the blocks related to the currently set algorithm.  e.g. if you want the average network hashes per second over the last 120 blocks [the default] it will be between the last 120 blocks of the specific algorithm you're configured for.  An algorithm parameter was added so that you can specify a different algorithm if you chose, including the historical xr16t, or if you care about PoS.  Note that this is actually calculating the change in difficulty over time, and not hashes.  That research was also out of scope for this PR, and if the logic needs to change it will be a separate PR.

The output to `getnetworkhashps` is also included in `getmininginfo`, which in testing revealed that `getmininginfo` displays the difficulty for the last block; and not the difficulty for the algorithm that you're actually mining.  So that logic was changed to report the difficulty for whatever algorithm you're set to mine.  An 'algorithm' field was also added to the output so it can be clear what algorithm the difficulty relates to:

```
$ veil-cli getmininginfo
{
  "blocks": 988984,
  "currentblockweight": 0,
  "currentblocktx": 0,
  "algorithm": "progpow",
  "difficulty": 174.107098381071,
  "networkhashps": 22.83160812508121,
  "pooledtx": 2,
  "chain": "main"
}
```

```
$ veil-cli getmininginfo
{
  "blocks": 988986,
  "currentblockweight": 0,
  "currentblocktx": 0,
  "algorithm": "sha256d",
  "difficulty": 19.31658925581963,
  "networkhashps": 23.3181708549265,
  "pooledtx": 1,
  "chain": "main"
}
```

Given the availability of all the difficulties with the `getblockchaininfo` command, a parameter to get mining info for a different algorithm was not added.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -58,38 +58,104 @@ unsigned int ParseConfirmTarget(const UniValue& value)
  * or from the last difficulty change if 'lookup' is nonpositive.
  * If 'height' is nonnegative, compute the estimate at the time when a given block was found.
  */
-static UniValue GetNetworkHashPS(int lookup, int height) {
+static UniValue GetNetworkHashPS(int lookup, int height, std::string type) {
     CBlockIndex *pb = chainActive.Tip();
 
     if (height >= 0 && height < chainActive.Height())
         pb = chainActive[height];
 
+    // if after fork, set pb to fork height unless specifying pos or x16rt
+
     if (pb == nullptr || !pb->nHeight)
         return 0;
 
     // If lookup is -1, then use blocks since last difficulty change.
-    if (lookup <= 0)
-        lookup = 1;
+    if (lookup <= 1)
+        lookup = 2;
 
-    // If lookup is larger than chain, then set it to chain length.
-    if (lookup > pb->nHeight)
-        lookup = pb->nHeight;
-
-    CBlockIndex *pb0 = pb;
-    int64_t minTime = pb0->GetBlockTime();
-    int64_t maxTime = minTime;
-    for (int i = 0; i < lookup; i++) {
-        pb0 = pb0->pprev;
-        int64_t time = pb0->GetBlockTime();
-        minTime = std::min(time, minTime);
-        maxTime = std::max(time, maxTime);
+    CBlockIndex *pbLast = pb;
+    CBlockIndex *pbFirst = nullptr;
+    bool found = false;
+    while (pb->nHeight > 0)
+    {
+        // find the last of the algo
+        if (0 == type.compare("PoS")) {
+            if (pb->IsProofOfStake()) found=true;
+        } else if (0 == type.compare("sha256d")) {
+            if (pb->GetBlockTime() < Params().PowUpdateTimestamp()) return 0;
+            if (pb->IsSha256DProofOfWork()) found=true;
+        } else if (0 == type.compare("randomx")) {
+            if (pb->GetBlockTime() < Params().PowUpdateTimestamp()) return 0;
+            if (pb->IsRandomXProofOfWork()) found=true;
+        } else if (0 == type.compare("progpow")) {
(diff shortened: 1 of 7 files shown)


src/rpc/mining.cpp
```diff
@@ -58,63 +58,144 @@ unsigned int ParseConfirmTarget(const UniValue& value)
...
+    if (lookup <= 1)
+        lookup = 2;
+
+    CBlockIndex *pbLast = pb;
+    CBlockIndex *pbFirst = nullptr;
+    bool found = false;
+    while (pb->nHeight > 0)
+    {
+        // find the last of the algo
+        if (0 == type.compare("PoS")) {
```
codeofalltrades, line 82: nit. This tripped me up testing. I was trying pos

src/rpc/mining.cpp
```diff
@@ -58,63 +58,144 @@ unsigned int ParseConfirmTarget(const UniValue& value)
...
 
-    LOCK(cs_main);
-    return GetNetworkHashPS(!request.params[0].isNull() ? request.params[0].get_int() : 120, !request.params[1].isNull() ? request.params[1].get_int() : -1);
+
+    if (!(request.params[2].isNull())) {
+        std::string algo = request.params[2].get_str();
+        // Check if it's a mining algorithm
+        if (!SetMiningAlgorithm(algo, false))
+            // check that it's not one of the other two
+            if (algo.compare("PoS") && algo.compare("xr16t"))
```
codeofalltrades, line 190: Should this be an or?  ||

codeofalltrades (commented):
0807f5243547323066ae5b8512e8f0bb18f32d0f
One question about logic.

src/rpc/mining.cpp
CaveSpectre11: compare() is confusing; it returns '0' if it matches.  so in C++, `if compare()` means if it doesn't match.  So here, we're checking that it "doesn't match PoS and doesn't match xr16t".  So rolling the whole flow together:

If SetMiningAlgorithm failed, it's either a typo, or it's PoS or it's xr16t.  So if it's not a mining algo, and it's not PoS and it's not xr16t, then we'll throw an error.

src/rpc/mining.cpp
CaveSpectre11: fair enough.  Changed to "pos" [untested] and pushed to #878.  [rather than pushing to all of them].  I left them as separate PRs so they're easier to test and review.

codeofalltrades:
ACK 0807f5243547323066ae5b8512e8f0bb18f32d0f

Issue #504, opened 2019-04-13
Feature Request: Toggle State Should Be Absolute. Toggle State Should Not Be A Suggesting An Action Is Being Done.


Rock-N-Troll:
Currently using a Toggle to perform an action in the GUI has a process go from State A to State B OR from State B back to State A.

However, the current visual toggle state appears to move to the next state before the next state is actually reached in the wallet's internals.
.
The wallet's state might be displayed as active/inactive when it actually is not what is displayed.

For this reason, there should be an intermediate state or pop-up that explains the user's action is being processed (Enabling...) (Disabling...) and the toggle should not move until the end state is actually reached by the wallet's internals.

**Current Implemenation:**
1.(May or May not actually be ENABLED)
![image](https://user-images.githubusercontent.com/34344520/56083468-71d91900-5df3-11e9-97e1-1fd16cb97c75.png)

2. (May or May not actually be DISABLED)
![image](https://user-images.githubusercontent.com/34344520/56083474-7c93ae00-5df3-11e9-933c-1b119a4d0ea7.png)

**Possible/Proposed Implementation:**
1.(Guaranteed to be ENABLED)
![image](https://user-images.githubusercontent.com/34344520/56083494-bf558600-5df3-11e9-807a-c7a4fd7ccce5.png)

2. Currently in the process of disabling and the GUI shows that the wallet is still precomputing
![image](https://user-images.githubusercontent.com/34344520/56083509-f5930580-5df3-11e9-8198-f9f4981802eb.png)

3. (Guaranteed to be DISABLED)
![image](https://user-images.githubusercontent.com/34344520/56083474-7c93ae00-5df3-11e9-933c-1b119a4d0ea7.png)

Whether or not this is the best way to show the user that their toggle click is being processed is not my place to decide. Maybe disabling the sliding indicator by greying it out will be a better way of showing an action is being performed and the toggle can't be clicked until the current toggle action is completed. Maybe a pop-up and graying out of the toggle switch is the best approach.

However, the current user experience should be better in line with what was described at the start of this Feature Request: Toggle State should be absolutely correct. Users shouldn't be required to go to the console to determine the state of their wallet's staking/precomputing/etc.

blondfrogs:
[...]

Issue #245, opened 2019-01-09
Provide separate state labels when staking is enabled and disabled


gabrielnergaard:
I would like to show a different label when staking is enabled or disabled:

![](https://clipz.in/9d8zx.png)

For example:

- Staking is active
- Staking disabled

Searching the codebase, however, it seems we currently don't support showing different labels depending on state. Could we please provide this?

PR #880, opened 2020-12-06
[GUI] Update Staking Slider to be reflective of status

WetOne:
### Problem
The staking slider does not always indicate the present state of staking.  There is a delay between when staking is enabled/disabled and when the actual enabling/disabling takes place.

### Root Cause
The Qt indicates the state that the wallet is in.  Some amount of time is required to ensure that the state change has taken affect.  Additionally, staking is not performed while syncing.  When the user initially opens the user interface, staking is displayed as active while syncing.  Staking is initially active but will not be performed until syncing is complete.

### Solution
* Text for the staking slider now shows the current state of actual staking (enabled, disable, disabled for sync, enabling ..., disabling...).
* The staking slider is disabled (cannot be changed) until syncing is complete.
* Once synchronization is complete, the staking slider will be changed to enabled with the state of "Enabling ...).  When finishing synchronization testing showed some thrashing between the states of "Enabling..." and "Staking Enabled" until a few blocks have gone by.  This is due to the way active staking is calculated.
* RPC getwalletinfo updated to show the actual state of staking and the wallet status
* Popup entering into staking disabled has been updated to indicate that some time is required prior to the state change taking effect.

### Issues Addressed
245: https://github.com/Veil-Project/veil/issues/245
504: https://github.com/Veil-Project/veil/issues/504

### Bounty Payment Address
sv1qqphsvuwhk29xcn2q9gdth9x73qpm8kcktmuxyd8szl50sgt2nt47egpqwnxr83hfj7s6fnec2jr0cv5yc7r6s7nvxhd9969qrgy4cgnznwewqqqauut5m

### DIFF
--- a/src/qt/veil/veilstatusbar.cpp
+++ b/src/qt/veil/veilstatusbar.cpp
@@ -10,6 +10,7 @@
 #include <qt/walletmodel.h>
 #include <qt/veil/qtutils.h>
 #include <iostream>
+#include <timedata.h>
 #ifdef ENABLE_WALLET
 #include <wallet/wallet.h> // For DEFAULT_DISABLE_WALLET
 #endif
@@ -51,6 +52,7 @@ void VeilStatusBar::updateSyncIndicator(int height){
 
 void VeilStatusBar::setSyncStatusVisible(bool fVisible) {
     ui->btnSync->setVisible(fVisible);
+    syncFlag = fVisible;
 }
 
 void VeilStatusBar::onBtnSyncClicked(){
@@ -59,6 +61,52 @@ void VeilStatusBar::onBtnSyncClicked(){
 
 #ifdef ENABLE_WALLET
 bool fBlockNextStakeCheckSignal = false;
+void VeilStatusBar::setStakingText() {
+
+    // Determine if staking is recently active. Note that this is not immediate effect. Staking could be disabled and it could take some time (activeStakingMaxTime) update state.
+    int64_t nTimeLastHashing = 0;
+    if (!mapHashedBlocks.empty()) {
+        auto pindexBest = chainActive.Tip();
+        if (mapHashedBlocks.count(pindexBest->GetBlockHash())) {
+            nTimeLastHashing = mapHashedBlocks.at(pindexBest->GetBlockHash());
+        } else if (mapHashedBlocks.count(pindexBest->pprev->GetBlockHash())) {
+            nTimeLastHashing = mapHashedBlocks.at(pindexBest->pprev->GetBlockHash());
+        }
+    }
+    bool fStakingActive = false;
+    if (nTimeLastHashing)
+        fStakingActive = GetAdjustedTime() + MAX_FUTURE_BLOCK_TIME - nTimeLastHashing < ACTIVE_STAKING_MAX_TIME;
+	
+    WalletModel::EncryptionStatus eStatus = this->walletModel->getEncryptionStatus();
+
+    if (syncFlag){
+	ui->checkStaking->setText("Staking Disabled while Syncing");
+    }else if (WalletModel::Locked == eStatus) {
+        ui->checkStaking->setText("Unlock wallet for Staking");
+    }else if (this->walletModel->isStakingEnabled()) {
+	if (fStakingActive) {
+            ui->checkStaking->setText("Staking Enabled");
+	}else{
+
+            interfaces::Wallet& wallet = walletModel->wallet();
+            interfaces::WalletBalances balances = wallet.getBalances();
+            int64_t zerocoin_balance = balances.zerocoin_balance;
+
+	    if (0.0 < zerocoin_balance) {
+                ui->checkStaking->setText("Enabling...");
+            }else{
+                ui->checkStaking->setText("You need some zerocoin");
+            }
+	}
+    }else{
+	if (fStakingActive) {
+	    ui->checkStaking->setText("Disabling...");
+	}else{
+	    ui->checkStaking->setText("Staking Disabled");
+	}
+    }
+}
+
 void VeilStatusBar::onCheckStakingClicked(bool res) {
     if (gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET))
         return;
@@ -77,16 +125,19 @@ void VeilStatusBar::onCheckStakingClicked(bool res) {
             openToastDialog(dialogMsg, mainWindow);
             fBlockNextStakeCheckSignal = true;
             ui->checkStaking->setChecked(false);
+            setStakingText();
             return;
         }else{
             this->walletModel->setStakingEnabled(true);
             mainWindow->updateWalletStatus();
             openToastDialog("Staking enabled", mainWindow);
+            setStakingText();
         }
     } else {
         this->walletModel->setStakingEnabled(false);
         mainWindow->updateWalletStatus();
-        openToastDialog("Staking disabled", mainWindow);
+        openToastDialog("Staking disabled - this may a few minutes", mainWindow);
+        setStakingText();
     }
 
 }
@@ -192,11 +243,15 @@ void VeilStatusBar::updateStakingCheckbox()
 
     if(walletModel) {
         WalletModel::EncryptionStatus lockState = walletModel->getEncryptionStatus();
-        bool stakingStatus = walletModel->isStakingEnabled() && lockState != WalletModel::Locked;
+
+	ui->checkStaking->setEnabled(!syncFlag);
+        setStakingText();
+
+	bool stakingStatus = walletModel->isStakingEnabled() && lockState != WalletModel::Locked && !syncFlag;
         if (ui->checkStaking->isChecked() != stakingStatus) {
             fBlockNextStakeCheckSignal = true;
             ui->checkStaking->setChecked(stakingStatus);
-            return;
+	    return;
         }
     }
 }
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -988,6 +988,8 @@ void BitcoinGUI::setNumBlocks(int count, const QDateTime& blockDate, double nVer
         tooltip += tr("Transactions after this will not yet be visible.");
     }
 
+    veilStatusBar->updateStakingCheckbox();
+
     // Don't word-wrap this (fixed-width) tooltip
     tooltip = QString("<nobr>") + tooltip + QString("</nobr>");
 
--- a/src/qt/veil/veilstatusbar.h
+++ b/src/qt/veil/veilstatusbar.h
@@ -52,8 +52,13 @@ private:
     WalletModel *walletModel = nullptr;
     ClientModel *clientModel = nullptr;
     UnlockPasswordDialog *unlockPasswordDialog = nullptr;
+#ifdef ENABLE_WALLET
+    void setStakingText();
+#endif
 
     bool preparingFlag = false;
+    bool syncFlag = true;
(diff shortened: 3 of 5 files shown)


src/qt/veil/veilstatusbar.cpp
```diff
@@ -59,6 +61,39 @@ void VeilStatusBar::onBtnSyncClicked(){
...
+            auto pindexBest = chainActive.Tip();
+            if (mapHashedBlocks.count(pindexBest->GetBlockHash())) {
+                nTimeLastHashing = mapHashedBlocks.at(pindexBest->GetBlockHash());
+            } else if (mapHashedBlocks.count(pindexBest->pprev->GetBlockHash())) {
+                nTimeLastHashing = mapHashedBlocks.at(pindexBest->pprev->GetBlockHash());
+            }
+        }
+        bool fStakingActive = false;
+        if (nTimeLastHashing)
+            fStakingActive = GetAdjustedTime() + MAX_FUTURE_BLOCK_TIME - nTimeLastHashing < 70;
```
CaveSpectre11, line 78: Define 70 as a static const, with a reflective name and description; then use it here and in the toast dialog further down

WetOne: I'm OK with using it within the time check.  

I suggest changing the toast to something along the lines of:  "This may take a few minutes" as
1) The display changing in the Qt is a function of when staking is active and the calls are made (there is no periodic task that performs this check, only when new blocks come in).  In my testing I've had a few minutes pass before the display is updated  though calls through the command line show active already.  This has to do with how the display is updated (when the command line is used, the calculation is done on the spot).
2) The 70 was copied from other code and I have no knowledge of what the actual high water mark is.

CaveSpectre11: The comment is merely to adjust away from using magic numbers.  Instead of "70" hardcoded in two places [the code and the message], make it 
`static const STAKING_DELAY = 70`

or something to that effect; so that it's defined; and if adjusted anytime in the future you don't have the message (stating 70 seconds) and the code diverging.

WetOne: Updated the message to be less definitive of the timeframe (as the user won't see the change real-time).  Added the constant and used it in the check.

CaveSpectre11: Did you take care of the "enabling..." thing while it's synching?

WetOne: Yup.  Works like it should now.  Have it a quick test too.

seanPhill:
Looking good so far. Status while syncing a testnet wallet on a ancient MacBook that was a month behind.
"Staking Disabled while Syncing"
Screen Shot 2020-12-10 at 8 13 41 PM

codeofalltrades (changes requested):
NACK 7556f9ba64d23a9113d89936a862265ea6b42f4a

src/qt/veil/veilstatusbar.cpp
```diff
@@ -59,6 +61,39 @@ void VeilStatusBar::onBtnSyncClicked(){
...
+    if (nTimeLastHashing)
+        fStakingActive = GetAdjustedTime() + MAX_FUTURE_BLOCK_TIME - nTimeLastHashing < ACTIVE_STAKING_MAX_TIME;
+	
+    if (syncFlag){
+	ui->checkStaking->setText("Staking Disabled while Syncing");
+    }else if (this->walletModel->isStakingEnabled()) {
+	if (fStakingActive) {
+            ui->checkStaking->setText("Staking Enabled");
+	}else{
+	    ui->checkStaking->setText("Enabling...");
```
codeofalltrades, line 86: I think we need to checked for a locked wallet here. If the wallet is locked, display "unlock wallet for staking"
image

More of NIT, when the wallet is unlocked and there are no zerocoins, the text stays at Enabling...
image

codeofalltrades (changes requested):

WetOne:
1) Updated to specify that the wallet needs to be unlocked to stake
2) when trying to stake with not zerocoins the text will notify the user that zerocoins are needed

seanPhill:
Got some more successful screenshots. Tried while already unlocked and with sufficient zerocoins. 
Screen Shot 2020-12-13 at 10 00 33 am
Screen Shot 2020-12-13 at 10 02 00 am
Screen Shot 2020-12-13 at 10 01 03 am
Screen Shot 2020-12-13 at 10 00 52 am

codeofalltrades (approved):
ACK 7c927b50c071036a72abad96c618206848dd92f4

CaveSpectre11:
@WetOne one obscure use case that still has a problem.  This came up due to a discussion with Hort regarding a problem he's had for a while.  So we set out to see if this fixes his problem.  First, the problem we found:
`staking=0` in your veil.conf file.  Start `veil-qt`.  It doesn't properly recognize that `staking=0` was set and it sits at "Enabling...", with 'staking_enabled: true and staking_active: false'.

After taking a look:
```
bool MnemonicWalletInit::Open() const
[...]
        if (gArgs.GetBoolArg("-exchangesandservicesmode", false))
            pwallet->SetStakingEnabled(false);
```

So a check on `-staking` needs to be put in that code as well to set staking disabled on startup when -staking=0.  A quick test with 'exchangesandservicesmode' confirms it comes up "staking disabled", so that should be the right place.

(if you're feeling really randy and want to play with veild and confirm that it works there too... look in `init.cpp` and you'll find:
```
        //Start staking thread last
        if (!gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET) && gArgs.GetBoolArg("-staking", true) && !gArgs.GetBoolArg("-exchangesandservicesmode", false))
            threadGroupStaking.create_thread(&ThreadStakeMiner);
```
which probably would work with a simple "else set staking enabled false".

seanPhill:
I like this implementation on the whole, but another potential annoyance is that if the wallet is locked and you click the slider where it says "Unlock wallet for staking" it then tells you to unlock the wallet, and when you click over on the padlock icon and type your password, unlocking it, then the slider activates staking. This might annoy some people, in that clicking "unlock wallet for staking" doesn't actually unlock the wallet, as you have to go to another control for that.

CaveSpectre11:
> where it says "Unlock wallet for staking" it then tells you to unlock the wallet, and when you click over on the padlock icon and type your password, unlocking it, then the slider activates staking.

I'll have to play around with it when I get a chance.  You're saying that it doesn't activate staking when you unlock for staking; only when you unlock the wallet as a whole?  Yea, that would be an issue.  It should start staking [assuming staking is enabled underneath] when it's unlocked for staking or when it's fully unlocked.

>this might annoy some people, in that clicking "unlock wallet for staking" doesn't actually unlock the wallet

Well, that should be the expected behavior; unlock for staking.... turns on staking.  unlock for staking shouldn't need a full unlock for spending to work.

seanPhill:
To test, I locked it again with the padlock icon and then unlocked for staking from the other slider in the Settings page.
The bottom slider unlocked as expected. (Currently saying "You need some zerocoin" on this wallet, only because I need to mint another 1000 denom in order to mature the others I already minted all at once.)

WetOne:
Quick update:  I modified line 75 to be 
```
if (gArgs.GetBoolArg("-exchangesandservicesmode", false) || gArgs.GetBoolArg("-staking",false))
```
and then ran "veil-qt -staking=0".  The GUI starts with "Unlock wallet for staking".  OK.  After unlocking the wallet it sits at "Enabling..." and doesn't transition to enabled.  Using the CLI shows that staking is enabled but staking doesn't go active.  I'm going to investigate this further when I get the chance later today.  I don't know what is keeping the enabled and the actual staking in two different states.

I did put instrumentation to verify that the staking flag when starting the GUI was being accepted.

Bottom line: either I put the one line change in the wrong spot (or the change is wrong, should probably default staking to true) or this is more than a one line change.

CaveSpectre11:
> **Bottom line**: either I put the one line change in the wrong spot (or the change is wrong, should probably default staking to true) or this is more than a one line change.

The two flags are inverted logic.  But that exchanges mode might be wrong?  I would think it should be "if exchangesandservicesmode, true || staking false.   But it could very well be that exchangesandservicesmode is "working" due to some code elsewhere; since it would imply that code would pass through and disable staking if exchangesandservices mode is not set.  So maybe that code path isn't exercised.

Maybe throw some debug logging in and see if that code is called at all; and if not, find where exchangesandservices mode disables staking.

WetOne:
I fixed the inverted logic and verified that SetStakingEnabled(false) being called in mnemonic wallet.  Then ran the qt with -staking=0, unlocked my wallet (staking stayed disabled), and hit the slider.  The text went to Enabling... but never got to staking enabled.  through the CLI staking enabled is set to true but staking active doesn't go to true.  I'm going to look through what is holding back the activity.

CaveSpectre11:
> ```
> if (gArgs.GetBoolArg("-exchangesandservicesmode", false) || gArgs.GetBoolArg("-staking",false))
> ```

Right place, wrong interpretation of `GetBoolArg()`.  After a bit of confusion... the parameter to GetBoolArg is not what you're checking against, it's what the 'default' is if it's not present.

so  `gArgs.GetBoolArg("-exchangesandservicesmode", false)` returns "false" if exchangesandservicesmode is not set.  Likewise `gArgs.GetBoolArg("-staking", false)` returns "false" if set to 0, or not set at all.

So the check should be
```
if (gArgs.GetBoolArg("-exchangesandservicesmode", false) // if exchanges and services mode is set
|| !gArgs.GetBoolArg("-staking",true)) // if staking is specified, and set to false (0)
```

WetOne:
@CaveSpectre11, @codeofalltrades 

Made the change in mnemonicalletinit.cpp to include the check for staking being disabled.  

There is still an issue when launching with staking disabled, unlocking, and then trying to enable staking.

CaveSpectre11:
> There is still an issue when launching with staking disabled, unlocking, and then trying to enable staking.

Just to help explain what we realized.  @codeofalltrades; so the problem is that, like so many other places, args are checked in the middle of the code, rather than on startup.  In this case what's happening rests right here:
```
//Start staking thread last
if (!gArgs.GetBoolArg("-disablewallet", DEFAULT_DISABLE_WALLET) && gArgs.GetBoolArg("-staking", true) && !gArgs.GetBoolArg("-exchangesandservicesmode", false))
    threadGroupStaking.create_thread(&ThreadStakeMiner);
```

If we come up with `staking=0`, the staking thread doesn't get started.  So when we later enable staking in the wallet, fStakingEnabled gets turned on, yet the thread isn't there lying dormant waiting to be staked:

So on a separate PR; that will all be fixed.  It's out of scope for this as starting with 'staking=0' has never been able to start staking without restarting the wallet.

That second PR will move the check for '-staking=0' to the init, but take it off the check to start the staking thread.  That way the staking thread will start regardless if staking is enabled or not.  That will allow the thread to lay dormant until staking is enabled.   The dormancy is here:

```
if (!pwallet || !pwallet->IsStakingEnabled() || (pwallet->IsLocked() && !pwallet->IsUnlockedForStakingOnly())) {
    MilliSleep(5000);
    continue;
}
```

This will also allow for a cli command to be able to turn staking on and off as well, and no longer require the wallet to be restarted.  You can set your desired default, and then enable or disable when you chose to.

codeofalltrades (approved):
ACK cad2120e49bff1188f2f55299b78d3272d6be363

CaveSpectre11 (approved):
ACK https://github.com/Veil-Project/veil/pull/880/commits/cad2120e49bff1188f2f55299b78d3272d6be363

CaveSpectre11:
Bounty Paid:  6200 + 100 for using PR form

Issue #875, opened 2020-11-27
[Feature Request][Mining] Provide API into current hashrates


CaveSpectre11:
Users would like a way to get the current hash rates from the running daemon.  Since it's calculated from within each thread, it's likely that it will need to be a mutex locked map that is updated when each thread calculates it's hash.  Each thread will need a unique identifier; and their entry in the map removed during thread teardown.

Two routes could be done; either place the latest calculated hash in the map; or place the time and count used to calculate the hash into the map.  The latter would be better to get an accurate total, rather than adding the hashes together.

cli can lock, copy the map, and unlock; then process.
generate json output that gives the total hashrate, and an optional 'true' for details on the indiividual hashrates.

for testing, the individual hashrates reported should match the last hashrate reported to the debug log from each thread.

PR #883, opened 2020-12-11
[RPC][Mining] Extract last Hashspeed calculation

CaveSpectre11:
### Problem
Users want to be able to see their current hashing speed

### Solution
Extract the last calculated hashspeed and provide it in `getmininginfo`.  To note, this only pulls the last calculation that was done, and is not a real-time pull.  It will be the average from the start of mining to the last calculation, from the time the first `generatecontinuous true` was run.  In other words, this will take into account any 'breaks' taken in mining.  e.g. if you have a script that turns off mining for a certain time; the hash speed calculations are still being run; and the break will be calculated into your overall hash rate.

Logic was added to `setminingalgo` to reset the statistical counters.  So after `setminingalgo` is issued, the statistics will restart on the next `generatecontinuous`.  This can be used to reset your counters even if you're not changing algorithms by setting your mining algo to the same algo:

```
$ ./veil-cli -regtest getmininginfo
{
[...]
  "hashspeed": 743034,
[...]
}
$ ./veil-cli -regtest setminingalgo sha256d
{
  "success": true,
  "message": "Mining algorithm changed from sha256d to sha256d"
}
$ ./veil-cli -regtest getmininginfo
{
[...]
  "hashspeed": 0,
}
```

Note that for simplicity sake, the calculation is "hashes per second", so that it can remain a numerical value and users can wrap the use however they wish if they want to convert it to kh/s or mh/s [or whatever they desire]

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -843,6 +843,20 @@ CCriticalSection cs_nonce;
 static int32_t nNonce_base = 0;
 static arith_uint256 nHashes = 0;
 static int32_t nTimeStart = 0;
+static int32_t nTimeDuration = 0;
+
+double GetHashSpeed() {
+    LOCK(cs_nonce);
+    if (!nTimeDuration) return 0;
+    return arith_uint256(nHashes/nTimeDuration).getdouble();
+}
+
+void ClearHashSpeed() {
+    LOCK(cs_nonce);
+    nHashes = 0;
+    nTimeStart = 0;
+    nTimeDuration = 0;
+}
 
 void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfStake = false, bool fProofOfFullNode = false) {
     LogPrintf("Veil Miner started\n");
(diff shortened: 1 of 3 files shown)


CaveSpectre11:
This addresses issue #875

codeofalltrades:
Looks goods
```
{
  "blocks": 1002401,
  "currentblockweight": 43832,
  "currentblocktx": 2,
  "algorithm": "sha256d",
  "difficulty": 18.71628533785223,
  "networkhashps": 23.47822805181118,
  "hashspeed": 351949,
  "pooledtx": 2,
  "chain": "main",
  "warnings": "This is a pre-release test build - use at your own risk - do not use for mining or merchant applications"
}
```

```
{
  "blocks": 1002409,
  "currentblockweight": 31364,
  "currentblocktx": 1,
  "algorithm": "randomx",
  "difficulty": 0.003405607281592146,
  "networkhashps": 23.28523685595233,
  "hashspeed": 43,
  "pooledtx": 1,
  "chain": "main",
  "warnings": "This is a pre-release test build - use at your own risk - do not use for mining or merchant applications"
}```

codeofalltrades (approved):
ACK ea18c5a269a5243675bee7eefd09e9408526c548

Issue #250, opened 2019-01-09
fea


gabrielnergaard:

CaveSpectre11:
Should be looked into at the same time as #136

Issue #136, opened 2019-01-01
Add Address numbering for address 002+


4x13:
Add a label that allows the user to see which address in line this is, like #002 or something. The address in the lower left (qr code) is the first generated address.

![request](https://user-images.githubusercontent.com/39984029/50575745-e0501900-0db8-11e9-9055-1791f71d01bb.png)

presstab:
@veildev I can help you get the info out of the backend, if you can do the display part.

presstab:
Bounty: 416

veilgets:
Developer: codeofalltrades

tirigor38:
OK

balarlardown:
good

codeofalltrades:
@veilgets Please remove me from this one.

veilgets:
reopened

veilgets:
Developer: alexeyneu

CaveSpectre11:
Bounty Reopened.  Bounty adjusted to 833.33

CaveSpectre11:
Should be investigated at the same time was #250

PR #888, opened 2021-01-02
[GUI] Update QR Code on address selection

WetOne:
### Problem
Selection of an address on the Address Book page does not update the QR codes on the receive page or the left hand side.

### Root Cause
Selection of an address on the Address Book page is isolated to the address book.

### Solution
Notify the other displays to allow for updates based on the selection of an address.  The update includes displaying address, QR code, name (if any).

### Issues Addressed
250: https://github.com/Veil-Project/veil/issues/250
136: https://github.com/Veil-Project/veil/issues/136

### DIFF
--- a/src/qt/veil/addresseswidget.cpp
+++ b/src/qt/veil/addresseswidget.cpp
@@ -30,6 +30,8 @@
 
 #include <iostream>
 
+#include <key_io.h>
+
 #define DECORATION_SIZE 54
 #define NUM_ITEMS 5
 
@@ -222,6 +224,7 @@ AddressesWidget::AddressesWidget(const PlatformStyle *platformStyle, WalletView 
     ui->lblSplit->setVisible(false);
     //connect(ui->btnMiningAddress,SIGNAL(clicked()),this,SLOT(onNewMinerAddressClicked()));
     //connect(ui->btnMiningAddress2,SIGNAL(clicked()),this,SLOT(onNewMinerAddressClicked()));
+
 }
 
 void AddressesWidget::setWalletModel(WalletModel *model)
@@ -240,6 +243,11 @@ void AddressesWidget::handleAddressClicked(const QModelIndex &index){
         listView = ui->listAddresses;
         type = AddressTableModel::Receive;
         updatedIndex = proxyModel->mapToSource(index);
+        auto address = this->model->index(updatedIndex.row(), AddressTableModel::Address, updatedIndex);
+        QString addressStr = this->model->data(address, Qt::DisplayRole).toString();
+        CTxDestination selectedAddress = DecodeDestination(addressStr.toStdString());
+
+        Q_EMIT rcvAddressSelected(&selectedAddress);
     }else{
         listView = ui->listContacts;
         type = AddressTableModel::Send;
@@ -371,6 +379,7 @@ void AddressesWidget::onNewAddressClicked(){
         openToastDialog(QString::fromStdString(toast + " Creation Failed"), mainWindow->getGUI());
     }
     // if it's the first one created, display it without having to reload
+// WE NEED TO DISPLAY THE NEW ADDRESS HERE AND IN THE BALANCE AND RECEIVE WIDGETS
     onForeground();
 }
 
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -37,6 +37,7 @@
 #include <chainparams.h>
 #include <interfaces/handler.h>
 #include <interfaces/node.h>
+#include <key_io.h>
 #include <ui_interface.h>
 #include <util.h>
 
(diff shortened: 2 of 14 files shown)


CaveSpectre11:
Investigating oddity with it pulling contact addresses.

src/qt/veil/addresseswidget.cpp
```diff
@@ -240,6 +243,16 @@ void AddressesWidget::handleAddressClicked(const QModelIndex &index){
         listView = ui->listAddresses;
         type = AddressTableModel::Receive;
         updatedIndex = proxyModel->mapToSource(index);
+        auto address = this->model->index(index.row(), AddressTableModel::Address, index);
```
CaveSpectre11, line 246: This needs to be the updatedIndex.

WetOne: Updated the line.  both index.row() and index (the last argument) have been updated to updatedIndex.

codeofalltrades:
In the console enter `getnewminingaddress`
Go to the contacts tab
Select the mining address.
Displayed address wrong
The copy button copies the incorrect address.
image

CaveSpectre11:
> In the console enter `getnewminingaddress`
> Go to the contacts tab
> Select the mining address.
> Displayed address wrong

I think it's decoding it into a destination address:
```
        LogPrintf("%s: index %d %d, updatedindex %d %d, addressStr = %s\n", __func__,
                  index.row(), index.column(), updatedIndex.row(), updatedIndex.column(), addressStr.toStdString().c_str());
        CTxDestination selectedAddress = DecodeDestination(addressStr.toStdString());
```
```
2021-01-03T03:32:52Z handleAddressClicked: index 6 0, updatedindex 8 0, addressStr = VKR4AXDC3mRq2zNfAhrMYgh1i74FbP6XwW
```

Probably have to dig into the bowels of the CTxDestination to determine if it should leave it as a raw address or decode it.

codeofalltrades:
copy via the tooltip copies the address that is displayed.
image
muQ3R2f8o2cf9S4gRGTxPJvXgU7zcP2a8z

WetOne:
@codeofalltrades, I think @CaveSpectre11 is correct - it is trying to decode the passed address into a destination address. The code was originally intended to only show a stealth receiving address (specifically, a stealth receiving address named "stealth", if you didn't have one, one is created for you).  The implementation was not intended to have anything other than a receiving address displayed.  What is your preference: to only have stealth receiving addresses displayed, or be able to have all non-contact addresses displayed (any address the My Addresses tab)?

codeofalltrades:
> or be able to have all non-contact addresses displayed (any address the My Addresses tab)

From a consistency perspective it should be all addresses. However with the move away from other coin types/addresses doing the work to support those types doesn't make sense resource wise. @CaveSpectre11 thoughts?

CaveSpectre11:
> > or be able to have all non-contact addresses displayed (any address the My Addresses tab)
> 
> From a consistency perspective it should be all addresses. However with the move away from other coin types/addresses doing the work to support those types doesn't make sense resource wise. @CaveSpectre11 thoughts?

I'm not sure how much work it would be to recognize the coin type.  It probably is possible, I'm sure there's a routine.  If there's a routine to check it and it's a couple lines of code (if !miningAddr, Decode destination), then it might be worthwhile since we don't have mining to stealth yet.

But if it's not simple, probably let it go.  But test it and see if you can send to that decoded address.  That would change the priority.  If they can't send to the decoded address, then that's a definite problem and we'll need to detect it.

If we can send to the decoded address; I'll let you make the severity call.

codeofalltrades:
> If we can send to the decoded address

looks like a no on the decoded address.
image

CaveSpectre11:
> > If we can send to the decoded address
> 
> looks like a no on the decoded address.

I'm wondering if it's a problem in decode destination; where the leading character isn't recognized; or otherwise decoding incorrectly for a mining address; which could also be related to the stuck mining spend problem.

WetOne:
@CaveSpectre11 , @codeofalltrades : This is almost definitely the encode/decode.  Do we know what needs to be changed other than the mining address?  The update for the mining address looks simple enough.  I would like to make the change once.

WetOne:
With a little more digging I found that the problem appears in the lines where EncodeDestination(arg1, arg2) is called.  In my implementation arg2 is always set to true.  This needs to be changed to be "if !(receive_miner == purpose)" in both balance.cpp and receive address widget. 

The change will definitely fix the issue with receiving miner addresses.  I don't see any indication in the code that more needs to be done for other addresses.  Let me know if you want me to just push the one change or if another update needs to be made.

codeofalltrades:
My vote is push the one change.

CaveSpectre11:
> My vote is push the one change.

Agreed.  Let's get it in here and make sure everything looks the way that's needed.

WetOne:
@CaveSpectre11, @codeofalltrades 
Updated.  Testing with mining addresses worked.

CaveSpectre11:
ACK https://github.com/Veil-Project/veil/pull/888/commits/7cf0778c04f68911aa075fa8b51a3d86b266bf68

CaveSpectre11 (approved):
ACK https://github.com/Veil-Project/veil/pull/888/commits/7cf0778c04f68911aa075fa8b51a3d86b266bf68

codeofalltrades (approved):
ACK 7cf0778c04f68911aa075fa8b51a3d86b266bf68

PR #893, opened 2021-01-20
[Wallet] Reduce shutdown delays when mining

CaveSpectre11:
# Problem
The wallet sometimes still takes a while before it actually shuts down when mining; this is most prevalent when the mining has been shutdown before the wallet is shut down.

# Root cause
The wallet will have to abort the mining generation loops when shutdown is active

# Solution
Add checks in the mining loops to abort when necessary


codeofalltrades (approved):
ACK f71b152ac726ca8279232e00252873efb554bab8

PR #895, opened 2021-01-21
Change default automint denom to 1000

codeofalltrades:
Changing the default automint denomination to 1000.


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/895/commits/1399cc4f883e4f21423788e4cce6a4cfabece2ef

Issue #281, opened 2019-01-12
Transactions show "Basecoin received with" either an unknown address or anonymous 1.0.0.9


veilMaDOS:
Transactions list shows "Basecoin received with" either a random bv address or anonymous, when I have been exclusively receiving basecoin mining rewards to a single bv address that I have generated.

**Expected behavior:**
Transactions should be showing "Basecoin received with suprnova", "suprnova" being the label I attached to a specific bv address that I generated to receive mining rewards.

Transaction list in v1.0.0.6
![labelsunknownaddresses_6-1](https://user-images.githubusercontent.com/44738391/51075904-e7d2c480-16cc-11e9-90e3-949812296bb3.PNG)

Transaction Details of the very bottom two transaction in v1.0.0.6. Notice how Sent to bv addresses are the same in both.
![labelsunknownaddresses_6-2](https://user-images.githubusercontent.com/44738391/51075908-ef926900-16cc-11e9-839b-ece55f3e9380.PNG)
![labelsunknownaddresses_6-3](https://user-images.githubusercontent.com/44738391/51075910-f28d5980-16cc-11e9-831d-a4d701de5809.PNG)

Address Book in v1.0.0.6. Notice how bv address with "suprnova" label is the same as the Sent to address of the two transactions above.
![labelsunknownaddresses_6-4](https://user-images.githubusercontent.com/44738391/51075914-f91bd100-16cc-11e9-8ef5-93eb9e796638.PNG)

**Actual behavior:**
It instead shows received with "Basecoin received with <random bv address>" (addresses listed are not in my Address Book or Contacts), or "Basecoin received with anonymous".

Transaction list in v1.0.0.9
![labelsunknownaddresses_9-1](https://user-images.githubusercontent.com/44738391/51075920-05079300-16cd-11e9-8cf0-098b1fb310ee.PNG)

Transaction Details of the very bottom two transactions in v1.0.0.9. For "Received with anonymous" transaction, Send to is blank. The other transaction shows a bv address that is unknown to me.
![labelsunknownaddresses_9-2](https://user-images.githubusercontent.com/44738391/51075941-397b4f00-16cd-11e9-91b8-ef56871b0309.PNG)
![labelsunknownaddresses_9-3](https://user-images.githubusercontent.com/44738391/51075943-3f713000-16cd-11e9-8067-253a5c79f506.PNG)

[...]

PR #898, opened 2021-03-06
[GUI] Display correct address when transaction has multiple addresses…

WetOne:
### Problem
Basecoin transactions with multiple addresses do not display correctly within the Qt display.

### Root Cause
When creating the transaction the wallet performs an extra push of an address.  This is a non-factor for stealth transactions as stealth transactions use a lookup table for the used address.  The basecoin address gets pushed in.  The extra push performed in the code results in a misalignment which appears when decoding the transaction.

### Solution
Removed the extra push.

### Testing
Tested for basecoin transactions which directly addresses issue 281.  This may also address issue 667 as well.

### Related Issues
https://github.com/Veil-Project/veil/issues/281

### DIFF
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -102,7 +102,6 @@ WalletTx MakeWalletTx(CWallet& wallet, const CWalletTx& wtx)
         auto txout = wtx.tx->vpout[i];
         auto ismine = wallet.IsMine(txout.get());
         result.txout_is_mine.emplace_back(ismine);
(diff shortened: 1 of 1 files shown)


CaveSpectre11 (approved):
ACK: https://github.com/Veil-Project/veil/pull/898/commits/44e00953a8eea4168ae667c6512929f1b0046ea0

codeofalltrades (approved):
ACK 44e00953a8eea4168ae667c6512929f1b0046ea0

PR #900, opened 2021-03-11
[GUI] Update wallet overview with coin and wallet information

codeofalltrades:
## Problem
New users don't have enough information to get the wallet synced.

## Solution
Provide the information.
image

### DIFF
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -94,7 +94,7 @@
            <number>0</number>
           </property>
           <item>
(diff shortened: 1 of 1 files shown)


CaveSpectre11 (approved):
ACK https://github.com/Veil-Project/veil/pull/900/commits/bf7cc796e96217f48c6cf2d1f7fb33c5522dae2a

Two things that would be nice to have but definitely not necessary for merging:
- It would be cool if "Encrypt Wallet" was a hotlink that popped up the Encrypt Wallet popup
- Likewise, if the reference to the debug console also was a hotlink

Issue #897, opened 2021-01-30
Total Balance question mark is difficult to use


WetOne:
The checkmark for total balance in the balance section of the UI does not properly respond to clicks.  Once it appears, it does not properly go away.

Referenced in discussion:
https://discord.com/channels/481997816982405130/566636115990151169/804968820811890688

sinetek:
Yes I have hit that issue as well.

seanPhill:
In cleaning up these issues I have confirmed that the current master does not have this problem, and the behaviour is as described in the commit merged #902 on March 17th, 2021.

PR #902, opened 2021-03-13
[GUI] Make total balance menu easier to display

WetOne:
### Problem
Clicking on the question mark icon for the Total Balance does not (consistently) result in the tooltip appearing.

### Root Cause
The tooltip is drawn too close to where the cursor is.  

### Solution
Move the origin of the tooltip to the right.  Note that the whole of the item (text and all) is the clickable to make the menu appear.  The issue only occurred when clicking the icon which was on the right end of the button.

### Issues Addressed
https://github.com/Veil-Project/veil/issues/897


codeofalltrades (approved):
ACK 4e08ba3ffd3ad7d5e4abb2b42b924043d7dc4d60

Issue #903, opened 2021-03-21
[GUI] Send Page Address Book Issues


CaveSpectre11:
Some issues to look into:
**Send->Address Book**
- There are 3 columns, the 3rd column appears to be a second copy of the address.  Need to look into the code to see if there's an indication of what that column is supposed to be
- Columns should have titles
- Right clicking on a Row brings up a menu, but the mouseover makes the selection white on white.

Rock-N-Troll:
1. It looks like address dot is really the shortened version of the address with the ellipsis in the middle (as discovered through copy paste) 
sv1xxxxxxxzg3qltya...h37xxxxxxxxxxxxzzd

2. The columns do have titles but I think the display on resize or platforms is sometimes not displaying properly:
Label, Address, Address dot

3.I suspect this is also related to Display issue

Rock-N-Troll:
addresstabelmodel.cpp:
 
```
 case Address_dot:
            return rec->address.left(18) + "..." + rec->address.right(18);
```

PR #905, opened 2021-03-25
[GUI] Add missing resize Address_dot column

Rock-N-Troll:


Rock-N-Troll:
#903 reference

Small Change: 
- Added missing column resize
- Also added the Address_dot column to the csv function although I don't believe it is working

codeofalltrades (approved):
ACK cfa7e278ead0c1b60d38e6dcfe5feacf39c20931
Tested on Win10 only.

PR #906, opened 2021-03-26
[GUI] Move menu to cursor location instead of far right of the rectangle

Rock-N-Troll:
**Before:**
image

**After:**
image

Previous code moved the selection menu to the far top right of the screen.
Instead, the  selection menu should intuitively be right next to where you click.

### DIFF
--- a/src/qt/veil/addressesmenu.cpp
+++ b/src/qt/veil/addressesmenu.cpp
@@ -19,9 +19,14 @@ AddressesMenu::AddressesMenu(const QString _type, const QModelIndex &_index, QWi
     mainWindow(_mainWindow),
     index(_index),
     type(_type),
-    model(_model)
+    model(_model),
+	timeoutTimer(0)
+
 {
     ui->setupUi(this);
+
+    timeoutTimer = new QTimer(this);
+    connect(timeoutTimer, SIGNAL(timeout()), this, SLOT(hide()));
 
     connect(ui->btnCopy,SIGNAL(clicked()),this,SLOT(onBtnCopyAddressClicked()));
     connect(ui->btnDelete,SIGNAL(clicked()),this,SLOT(onBtnDeleteAddressClicked()));
@@ -81,7 +86,15 @@ void AddressesMenu::setInitData(const QModelIndex &_index, AddressTableModel *_m
 }
 
 void AddressesMenu::showEvent(QShowEvent *event){
-    QTimer::singleShot(3500, this, SLOT(hide()));
+	timeoutTimer->start(3500);
+}
+
+void AddressesMenu::enterEvent(QEvent *event){
+	timeoutTimer->stop();
+}
+
+void AddressesMenu::leaveEvent(QEvent *event){
+	hide();
 }
 
 AddressesMenu::~AddressesMenu() {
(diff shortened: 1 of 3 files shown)


Rock-N-Troll:
I did another test and although the selection window moves to the cursor, it doesn't want to stay open long enough to move the cursor to one of the selections....will need to revisit

Rock-N-Troll:
The timeout really has no effect on my system because the click action immediately activates the mouse enter event, but still good to include. 

The original singleshot timer approach didn't reset the timer each time you click. It will always hide the widget after 3.5 seconds. This approach will always reset the timer even though the mouse over event disables the timer and is triggered immediately on click.

CaveSpectre11 (approved):
I'm not 100% sure the timer is working; but I like the behavior I'm seeing.

ACK https://github.com/Veil-Project/veil/pull/906/commits/22829f38c7ba75f04da8f61380fca0336c2374c3

Rock-N-Troll:
The timer is basically a "just in case" the mouse isn't somehow over it on click now that the mouse enter event immediately occurs which also stops it.

codeofalltrades:
If I go to the far right and click the menu is hidden behind the right nav. 
Maybe z-index on the pop up would fix it?
image

Rock-N-Troll:
adjusted (but it was a pain to guess and test this one):

image

codeofalltrades:
Ready for code review and QA?

WetOne:
If one has enough address to fill the screen, clicking the bottom address cuts the menu at the bottom of the screen (I only see "copy" not "copy" and "edit").  After, my window scrolls so that the clicked address is no longer the bottom (but is still highlighted).

Rock-N-Troll:
Would you happen to have a screenshot of the issue for better understanding?

WetOne:
image

CaveSpectre11:
> If one has enough address to fill the screen, clicking the bottom address cuts the menu at the bottom of the screen (I only see "copy" not "copy" and "edit"). After, my window scrolls so that the clicked address is no longer the bottom (but is still highlighted).

Nice catch @WetOne .   I've seen in the past with other applications that when there's not enough room to bring up the menu where it normally brings it up, it basically flips the anchor point.  E.g. With the menu coming up with the pointer to the top left, if it was too close to the right side, it would make the anchor point top right, too close to the bottom, anchor bottom left.  Too close to the bottom right corner, anchor bottom right.

Rock-N-Troll:
> > If one has enough address to fill the screen, clicking the bottom address cuts the menu at the bottom of the screen (I only see "copy" not "copy" and "edit"). After, my window scrolls so that the clicked address is no longer the bottom (but is still highlighted).
> 
> Nice catch @WetOne . I've seen in the past with other applications that when there's not enough room to bring up the menu where it normally brings it up, it basically flips the anchor point. E.g. With the menu coming up with the pointer to the top left, if it was too close to the right side, it would make the anchor point top right, too close to the bottom, anchor bottom left. Too close to the bottom right corner, anchor bottom right.

Let me see if I can copy that behavior. Good suggestions both.

Rock-N-Troll:
@WetOne fix applied. Please retest when you get a chance. Thank you

WetOne:
ack 51bd07ed719898ca643a4be013181200345dec5b

codeofalltrades (approved):
ACK 51bd07ed719898ca643a4be013181200345dec5b

WetOne (approved):

PR #907, opened 2021-03-26
[GUI] Fix: Address Book menu hover background is white and text is white

Rock-N-Troll:
image

Current (Screenshot 1 -> Screenshot 2): hovering a selection shows a blank selection due to the background and the text turning white

New (Screenshot 1 -> Screenshot 3): hovering a selection shows a grey background correctly. 
Aligns with grey #bababab used throughout the project.

### DIFF
--- a/src/qt/addressbookpage.cpp
+++ b/src/qt/addressbookpage.cpp
@@ -127,6 +127,7 @@ AddressBookPage::AddressBookPage(const PlatformStyle *platformStyle, Mode _mode,
     if(tab == SendingTab)
         contextMenu->addAction(deleteAction);
     contextMenu->addSeparator();
+    //contextMenu->setStyleSheet("QMenu::item:selected {background-color: #bababa;}");
 
     // Connect signals for context menu actions
     connect(copyAddressAction, SIGNAL(triggered()), this, SLOT(onCopyAddressClicked()));
--- a/src/qt/res/css/main.css
+++ b/src/qt/res/css/main.css
@@ -69,6 +69,10 @@ QLineEdit{
     padding:0px;
 }
 
+QMenu:item:selected {
+    color: #000000;
+    background-color: #6f9bf5;
+}
 
 QToolBar > QToolButton {
 


codeofalltrades (approved):
ACK 676242c5fe8852f99b8734b62a2a0ea4099b6725
Win10

WetOne:
All the other menus change the text color rather than change the background color.  For consistency this should change the foreground color of the text on hover.  I admit that (depending on your screen) that the other menus can be difficult to read (the thin font grey with white background).  We should change this to black text for the non-hovered text (as current) and use a more bold color (blue) for the hovered text.  This change should be rolled out over time to the other menus.

@CaveSpectre11, @codeofalltrades  - thoughts?

CaveSpectre11:
> We should change this to black text for the non-hovered text (as current) and use a more bold color (blue) for the hovered text. This change should be rolled out over time to the other menus.

I agree that it should be consistent, and I also agree that the light gray to black that exists on the My Addresses page isn't necessarily the right approach.  I also notice that the select color in My Addresses and in the Transactions list is different then the color scheme of the Send page address book.

I think we should work towards bringing them all together, this PR can be used as a starting point and then other PRs for the other portions.  There's two possible schemes that I see that would use existing layout elements and work for the mouseover; and that being the scheme the Transactions page, or the My Addresses page works.  Either:
1) the mouseover feel on those pages; black on white becomes dark gray on light gray
2) The selection feel on those pages; dark gray on light blue.

I'm not sure which one would look better, but I'm leaning towards the second option

WetOne:
Of the two you mention, dark grey on light blue is better.  Let me suggest a third: black text on light blue.  The contrast would help the text stand out more.

CaveSpectre11:
> Of the two you mention, dark grey on light blue is better. Let me suggest a third: black text on light blue. The contrast would help the text stand out more.

I'm good with that; of course the existing My addresses and and Transaction page scheme would need to change slightly (to have the black text on light blue instead of gray on light blue) to maintain consistency.

CaveSpectre11 (dismissed):
Technically sound; just a color scheme change for this (and can be kept with this even though there will be more places that will need this color scheme change).

Have the foreground stay black and have the background change to the light blue that is shown when you select an address on transaction on their relevant pages.

WetOne:
I was thinking about this.  Qt allows for a styleSheet for these menus.  I haven't looked but would be willing to bet that the "My Addresses" and "Transactions" pages are using the same style sheet.  If we change just the style sheet, one change would change all the screens that need to be changed - 1 file and maintain consistency.  Worst case, we need to create a stylesheet for these and then have these screens use the stylesheet.

Rock-N-Troll:
image

The blue I set corresponds with the color used elsewhere in the wallet (like hovering to find # of confirmations)

image

The hover in the addressbook is a lighter blue than this, but the value is easily changed now that the css has been established as was agreed upon in the issue comment thread above.

CaveSpectre11:
remember to rebase/squash your commits to a single atomic commit for a single change,

CaveSpectre11 (approved):
ACK https://github.com/Veil-Project/veil/pull/907/commits/1dba6c4ea1189e5e6a63c99965b763610faf9ecd

codeofalltrades:
ack 1dba6c4ea1189e5e6a63c99965b763610faf9ecd

PR #908, opened 2021-03-27
[Wallet] If nautomintdenom is set, prioritize its use

Rock-N-Troll:
nPreferredDenom == DEFAULT_AUTOMINT_DENOM is actually false because:
nPreferredDenom is 10
DEFAULT_AUTOMINT_DENOM is 1000

This code change just wraps the 2 nPreferredDenom inits based on if nautomindenom is set or not in the config.

The real existing issue is in veil.cpp which occurs after init.cpp and overwrites the logic in veil.cpp for nautomintdenom anyway when it pulls from QT getSettings()

Easiest way to test is to compare this code vs. existing 
use a fresh wallet or re-build/remake from code and config file nautomintdenom value is not used on initial startup

### Release Note
[GUI] Users that have set their desired zerocoin automint denomination via the GUI and in the veil.conf file will now find the configuration file setting is used over the GUI setting.  If the veil configuration file has an automint denom setting, the GUI controls are now disabled and the user informed.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1757,9 +1757,10 @@ bool AppInitMain()
 
 
     // Automint denom
-    if(nPreferredDenom == DEFAULT_AUTOMINT_DENOM) {
+    if(gArgs.IsArgSet("-nautomintdenom")){
         nPreferredDenom = gArgs.GetArg("-nautomintdenom", DEFAULT_AUTOMINT_DENOM);
         if(nPreferredDenom != 10 && nPreferredDenom != 100 && nPreferredDenom != 1000 && nPreferredDenom != 10000){
+        	LogPrintf("Invalid -nautomintdenom value set: %d. Defaulting to : %d\n", nPreferredDenom, DEFAULT_AUTOMINT_DENOM);
             nPreferredDenom = DEFAULT_AUTOMINT_DENOM;
         }
     }
--- a/src/qt/veil.cpp
+++ b/src/qt/veil.cpp
@@ -649,12 +649,15 @@ int main(int argc, char *argv[])
         return EXIT_FAILURE;
     }
 
-    // Preferences check
-    // If the UI settings have a different value than the server args then use the UI settings.
-    QSettings* settings = getSettings();
-    int tempPref = settings->value("nAutomintDenom").toInt();
-    if(tempPref != nPreferredDenom && tempPref != 0){
-        nPreferredDenom = tempPref;
+    // Automint denom
+    // If nautomintdenom is set, use it
+    // Else use saved UI settings
+    if(!gArgs.IsArgSet("-nautomintdenom")){
+    	QSettings* settings = getSettings();
+    	int tempPref = settings->value("nAutomintDenom").toInt();
+		if(tempPref != nPreferredDenom && tempPref != 0){
+			nPreferredDenom = tempPref;
+		}
     }
 
     // Now that the QApplication is setup and we have parsed our parameters, we can set the platform style
--- a/src/qt/veil/settings/settingsminting.cpp
+++ b/src/qt/veil/settings/settingsminting.cpp
@@ -53,6 +53,15 @@ SettingsMinting::SettingsMinting(QWidget *parent, WalletView *mainWindow, Wallet
         ui->btnSendMint->setEnabled(true);
     }
 
+    //retrieved saved settings if nautomintdenom is not in use
+    QSettings* settings = getSettings();
+    if (!gArgs.IsArgSet("-nautomintdenom") && settings->contains("nAutomintDenom")){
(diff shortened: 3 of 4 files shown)


Rock-N-Troll:
image

CaveSpectre11:
Generally we like to see multiple commits rebased down to one single commit (git rebase -i HEAD~6 and then change the ones that you want to combine to 'fixup'), but in this case I would be okay with 2 commits; one for the prioritization, and one for the disabling of the UI.

src/qt/veil.cpp
```diff
@@ -649,12 +649,15 @@ int main(int argc, char *argv[])
         return EXIT_FAILURE;
     }
 
-    // Preferences check
-    // If the UI settings have a different value than the server args then use the UI settings.
+    // Automint denom
+    // If nautomintdenom is set, use it
+    // Else use saved UI settings
```
CaveSpectre11, line 654: This change is a behavior change (changing the preference from the UI setting to the veil.conf setting) so definitely needs a statement in the release notes; as those that have both will see a change in behavior.

src/qt/veil.cpp
```diff
@@ -649,12 +649,15 @@ int main(int argc, char *argv[])
...
     QSettings* settings = getSettings();
```
CaveSpectre11, line 655: Since settings isn't used anywhere else, this should be moved into the if not IsArgSet context, so it doesn't waste the cycles if it's not going to be used.

src/qt/veil/settings/settingsminting.cpp
```diff
@@ -177,35 +187,35 @@ void SettingsMinting::mintzerocoins()
 }
 
 void SettingsMinting::onCheck10Clicked(bool res) {
-    if(res && nPreferredDenom != 10){
+    if(res && !gArgs.IsArgSet("-nautomintdenom") && nPreferredDenom != 10){
```
CaveSpectre11, line 190: This (and others) shouldn't be necessary since the radio buttons would be disabled if `nautomintdenom` is set.  No sense wasting the cycles for code that wouldn't be reached if it argument was set.

src/init.cpp
```diff
@@ -1757,7 +1757,7 @@ bool AppInitMain()
 
 
     // Automint denom
-    if(nPreferredDenom == DEFAULT_AUTOMINT_DENOM) {
+    if(gArgs.IsArgSet("-nautomintdenom")){
         nPreferredDenom = gArgs.GetArg("-nautomintdenom", DEFAULT_AUTOMINT_DENOM);
         if(nPreferredDenom != 10 && nPreferredDenom != 100 && nPreferredDenom != 1000 && nPreferredDenom != 10000){
             nPreferredDenom = DEFAULT_AUTOMINT_DENOM;
```
CaveSpectre11, line 1763: There should be a log message (at least) informing that the nautomintdenom is invalid and using the default instead (similar to other init messages that are logged).  Have to tell the user somehow that they need to go fix it.  Since the UI setting isn't the preference anymore, this now becomes more important; as there would be no way to understand that their nautomintdenom=20000 is actually using 1000 [for example]

CaveSpectre11 (commented):
Still need to do some testing; but overall looks good with a few minor things.

src/init.cpp
Rock-N-Troll: Sure.  This was pre-existing without any log message but I will certainly make that change for clarity

src/qt/veil/settings/settingsminting.cpp
Rock-N-Troll: Agreed. I'll remove

src/qt/veil.cpp
Rock-N-Troll: Yes I was thinking the same thing but left it as is for simplicity's sake. Will update

CaveSpectre11 (approved):
https://github.com/Veil-Project/veil/pull/908/commits/3ad26d360ca82be8206f25912b1912de76c44c21

codeofalltrades:
It seems like this is partially working. The disabled setting screen when nautomintdenom is set works.
If I remove nautomintdenom, restart the wallet, set the size, then restart the setting is not retained.

Rock-N-Troll:
> It seems like this is partially working. The disabled setting screen when nautomintdenom is set works.
> If I remove nautomintdenom, restart the wallet, set the size, then restart the setting is not retained.

I think that is an existing issue, but let me see what I can do

Rock-N-Troll:
The latest commit will also properly apply the saved checkbox settings whereas previously and existing code wasn't working.

codeofalltrades (approved):
ACK 313772e3c1183a59c68e1feb282c3e2b25339375

Issue #889, opened 2021-01-02
[Staking] Add a check for mature zerocoins to avoid "endless" Enabling... on GUI


dzacca:
If zerocoins are available but not mature enough and you try to enable staking, the gui shows "Enabling..." without any kind of working about the need of mature zerocoins.

A message informing the user that zerocoins need to me mature (and what maturity they need to have) before starting staking.

GUI just shows "Enabling..."

100% reproducible. You just need to have some non-mature zerocoins and enable staking

Veil Core version v1.1.1.1-4ca87868f (64-bit) on Mint Linux 20. Prebuilt

OS: Linux Mint 20 x86_64 
Host: TM1709 Mi Laptop 15.6 
Kernel: 5.4.0-58-generic  
Resolution: 1920x1080 
DE: Cinnamon 
WM: Mutter (Muffin) 
WM Theme: Mint-Y-Dark-Blue (Mint-Y-Dark)
Theme: Mint-Y-Dark-Blue [GTK2/3] 
Icons: Mint-Y-Dark-Blue [GTK2/3] 
Terminal: gnome-terminal 
CPU: Intel i5-8250U (8) @ 3.400GHz 
GPU: NVIDIA GeForce MX110 
GPU: Intel UHD Graphics 620 
Memory: 10925MiB / 15919MiB

PR #912, opened 2021-03-28
[WIP][Staking][GUI] Check for at least 1 mature zerocoin instead of no immature zerocoin

Rock-N-Troll:
Reference Issue #889 

-Change status bar logic to display `Staking Enabled` correctly:
     **Previous:** display `Enabling...` if there are 1 or more immature zerocoin in the wallet.
     **New:** display `Enabling...` if there are exactly 0 mature zerocoin in the wallet.
-Replace `Enabling...` text with `Waiting for zerocoin to mature...` text

### DIFF
--- a/src/qt/veil/veilstatusbar.cpp
+++ b/src/qt/veil/veilstatusbar.cpp
@@ -74,8 +74,11 @@ void VeilStatusBar::setStakingText() {
         }
     }
     bool fStakingActive = false;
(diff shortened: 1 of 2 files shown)


Rock-N-Troll:
Previous logic was checking to essentially make sure all zerocoin are mature.
This code also addresses rpcwallet -> `getwalletinfo` for `staking_active` field which follows the same logic

PR #914, opened 2021-04-01
[Utils] Use gmtime_s with _WIN32

Zannick:
### Problem ###
Following the build instructions for WSL with Ubuntu 20.04 resulted in this error:

```
utiltime.cpp: In function ‘std::string FormatISO8601DateTime(int64_t)’:
utiltime.cpp:85:5: error: ‘gmtime_r’ was not declared in this scope; did you mean ‘gmtime_s’?
   85 |     gmtime_r(&time_val, &ts);
      |     ^~~~~~~~
      |     gmtime_s 
```

Manually changing `gmtime_r` to `gmtime_s` results in a further warning because the arguments are reversed, a sign that we're building with the Windows version that the code uses when `_MSC_VER` is set.

### Root Cause ###
WSL with Ubuntu 20.04 is not building with msvc but with mingw.

### Solution ###
Use the Windows version of this code when either `_MSC_VER` or `__MINGW64__` is set.

(**Note**: I also needed to change the build command `make HOST=x86_64-w64-mingw32` to `CC_FOR_BUILD=x86_64-linux-gnu-gcc make HOST=x86_64-w64-mingw32` in order to build gmp, but I was not sure if I should update the docs.)

### Bounty PR ###
None

### Bounty Payment Address ##

### Unit Testing Results ###
Tested on win64 mingw WSL/Ubuntu 20.04 by building successfully and running unittests successfully compared to master.

### DIFF
--- a/src/utiltime.cpp
+++ b/src/utiltime.cpp
@@ -79,7 +79,7 @@ void MilliSleep(int64_t n)
 std::string FormatISO8601DateTime(int64_t nTime) {
     struct tm ts;
     time_t time_val = nTime;
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(_WIN32)
     gmtime_s(&ts, &time_val);
 #else
     gmtime_r(&time_val, &ts);
@@ -90,7 +90,7 @@ std::string FormatISO8601DateTime(int64_t nTime) {
 std::string FormatISO8601Date(int64_t nTime) {
     struct tm ts;
     time_t time_val = nTime;
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(_WIN32)
     gmtime_s(&ts, &time_val);
 #else
     gmtime_r(&time_val, &ts);
@@ -101,7 +101,7 @@ std::string FormatISO8601Date(int64_t nTime) {
 std::string FormatISO8601Time(int64_t nTime) {
     struct tm ts;
     time_t time_val = nTime;
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(_WIN32)
     gmtime_s(&ts, &time_val);
 #else
     gmtime_r(&time_val, &ts);


Rock-N-Troll:
@codeofalltrades had mentioned to make this change to support windows as well. I was encountering this issue and lazily just defaulted to gmtime_s, but this is a much better approach.

CaveSpectre11 (approved):
https://github.com/Veil-Project/veil/pull/914/commits/be9d3ec07b2a565aef917c37f5cf94ff2af69734

Ran a test with changing Github Actions to use Ubuntu 20.  This does solve the Win64 issue with ubuntu 20; but the same problem exists for Win32; So it looks like it needs one more conditional on that # if.

See here:
https://github.com/CaveSpectre11/veil/runs/2274102054?check_suite_focus=true

Zannick:
According to `touch /tmp/test.c; x86_64-w64-mingw32-gcc -E -dM /tmp/test.c | grep -i "mingw\|win"` the compiler defines plenty of symbols like `_WIN32`, `__WIN32`, `__WIN32__`, `__MINGW32__` and the 64 variants of those. Not sure which is really the most appropriate and I can't test Win32 specifically except by uploading here. I put `_WIN32` to try to be more general (in case cygwin ever gets used?), but I can switch it to `__MINGW32__`.

CaveSpectre11 (approved):
ACK https://github.com/Veil-Project/veil/pull/914/commits/aa7e26eb2e990bf2d223322eee3e8be1a0530025

https://github.com/CaveSpectre11/veil/actions/runs/736392142 (Ubuntu 20 actions)

codeofalltrades (approved):
ACK aa7e26eb2e990bf2d223322eee3e8be1a0530025

WetOne:
Ack aa7e26eb2e990bf2d223322eee3e8be1a0530025

Issue #862, opened 2020-11-02
Mining SHA256 via wallet becomes stuck - stopping and starting the miner fixes the problem temporarily


codeofalltrades:
**Describe the issue** 
Mining SHA256 via wallet becomes stuck though i never got left behind or banned. Feels like when the 5 in a row limit is reached while on sha, wallet swithes all resources trying to build the 6th but it gets rejected, creating a loop.

**What behavior did you expect? **
Mining SHA256 should not stop syncing the chain.

Mining is stuck at block until you stop the miner `generatecontinuous false`
and then restart it `generatecontinuous true 1`

**How reliably can you reproduce the issue, what are the steps to do so?** 

**What version of the Veil client are you using, where did you get it (website, self-compiled, etc)?** 
Qt wallet

**What type of machine are you observing the error on (OS/CPU and disk type)?** 
Windows 10 - Ryzen 3900x
AMD r7 3700x 

**Any extra information that might be useful in the debugging process.** 
 I have tried with both staking enabled and disabled

**Log file**
2020-11-02T20:08:01Z UpdateTip: new best=ff8881348a6283a2ec09740be6a30088bada4587490c0c4a7bc5d889f8c5063f height=942845 type=PoS version=0x30000000 tx=2155963 date='2020-11-02T20:07:24Z' cache=0.2MiB(1262txo)
2020-11-02T20:08:03Z ERROR: CreateNewBlock: stale tip.
2020-11-02T20:08:08Z ERROR: CreateNewBlock: stale tip.
2020-11-02T20:08:24Z UpdateTip: new best=fd18796a4c88dd8c24f119f73c020fa2c7053b5db7beae516395c868620d6af1 height=942846 type=PoS

codeofalltrades:
```020-11-06T06:42:42Z PruneStaleBlockIndexes: Checking for stale indexes. Block index size=947772; Chain height=947760
2020-11-06T06:42:43Z PruneStaleBlockIndexes: Checking for stale indexes. Block index size=947775; Chain height=947763
2020-11-06T07:01:15Z ERROR: CreateNewBlock: stale tip.
2020-11-06T07:50:10Z PruneStaleBlockIndexes: Checking for stale indexes. Block index size=947857; Chain height=947833
2020-11-06T07:50:10Z PruneStaleBlockIndexes: Checking for stale indexes. Block index size=947857; Chain height=947834```
Log file data at the block my test go stuck on.

seanPhill:
[...]

PR #915, opened 2021-04-01
[Mining][SHA256D] Reduce lock contention with SHA256D

Zannick:
### Problem ###
While mining SHA256D, the wallet periodically gets stuck, failing to stay caught up with the chain. (At the same time, the mining significantly slows down.)

### Root Cause ###
Lock contention; SHA256D is very fast, completing its loop in subsecond time on a modern cpu, meaning every thread goes through lock-heavy sections of the block template creation code multiple times per second. In comparison, RandomX on the same machine takes minutes to complete its loop, usually being interrupted by the creation of a new block instead of reaching the maximal number of attempts.

### Solution ###
Increase the number of loop iterations for SHA256D. To avoid doing useless work, also periodically check for a new block to end the outer iteration early.

* Adds an additional loop inside `BitcoinMiner` for SHA256D only, which reduces (ideally, minimizes) the number of times per chain tip that the mining thread needs to grab `cs_main`.
*  Periodically checks `chainActive.Height()` to see if a block has already been found, in order to exit early, similar to RandomX.

### Bounty PR ###
#862

### Bounty Payment Address ##

### Unit Testing Results ###
Manually ran all tests in `src/test/` via a script and compared to master. No differences.  
Ran on testnet.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -973,6 +973,13 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
             IncrementExtraNonce(pblock, chainActive.Height(), nExtraNonce);
 
             int nTries = 0;
+            bool success = false;
+            // As SHA runs much faster than the other algorithms, run more of them in this section
+            // to avoid lock contention from creating block templates. Use a separate counter
+            // nMidLoopCount to keep track of this, so that we can check every nInnerLoopCount hashes
+            // if the block is done.
+            int nMidTries = 0;
+            static const int nMidLoopCount = 0x1000;
             if (pblock->IsProgPow() && pblock->nTime >= Params().PowUpdateTimestamp()) {
                 uint256 mix_hash;
                 while (nTries < nInnerLoopCount &&
@@ -983,14 +990,25 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
                     ++pblock->nNonce64;
                 }
                 pblock->mixHash = mix_hash;
+                success = nTries != nInnerLoopCount;
             } else if (pblock->IsSha256D() && pblock->nTime >= Params().PowUpdateTimestamp()) {
                 uint256 midStateHash = pblock->GetSha256dMidstate();
-                while (nTries < nInnerLoopCount &&
-                       !CheckProofOfWork(pblock->GetSha256D(midStateHash), pblock->nBits,
-                                         Params().GetConsensus(), CBlockHeader::SHA256D_BLOCK)) {
-                    boost::this_thread::interruption_point();
-                    ++nTries;
-                    ++pblock->nNonce64;
+                // Exit loop when nMidLoopCount loops are done, or when a new block is found.
+                // Either way, success will be false.
+                while (nMidTries < nMidLoopCount && chainActive.Height() < pblock->nHeight) {
+                    while (nTries < nInnerLoopCount &&
+                           !CheckProofOfWork(pblock->GetSha256D(midStateHash), pblock->nBits,
+                                             Params().GetConsensus(), CBlockHeader::SHA256D_BLOCK)) {
+                        boost::this_thread::interruption_point();
+                        ++nTries;
+                        ++pblock->nNonce64;
+                    }
+                    if (nTries != nInnerLoopCount) {
+                        success = true;
+                        break;
+                    }
+                    ++nMidTries;
+                    nTries = 0;
                 }
             } else if (pblock->nTime < Params().PowUpdateTimestamp()) {
                 while (nTries < nInnerLoopCount &&
@@ -999,6 +1017,7 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
                     ++nTries;
                     ++pblock->nNonce;
                 }
+                success = nTries != nInnerLoopCount;
             } else {
                 LogPrintf("%s: Unknown hashing algorithm found!\n", __func__);
                 return;
@@ -1007,14 +1026,14 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
             double nHashSpeed = 0;
             {
                 LOCK(cs_nonce);
-                nHashes += nTries;
+                nHashes += nTries + (nMidTries * nInnerLoopCount);
                 nTimeDuration = GetTime() - nTimeStart;
                 if (!nTimeDuration) nTimeDuration = 1;
                 nHashSpeed = arith_uint256(nHashes/1000/nTimeDuration).getdouble();
             }
 
             LogPrint(BCLog::MINING, "%s: PoW Hashspeed %d kh/s\n", __func__,  nHashSpeed);
-            if (nTries == nInnerLoopCount) {
+            if (!success) {
                 continue;
             }
         }
@@ -1024,6 +1043,7 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
             LogPrint(BCLog::MINING, "%s: Failed to process new block\n", __func__);
             continue;
         }
+        LogPrint(BCLog::MINING, "%s: Found block\n", __func__);
 
         if (!fProofOfStake)
             coinbaseScript->KeepScript();


src/miner.cpp
```diff
@@ -964,26 +965,33 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
...
             pblock->nNonce64 = rand();
+            // Avoid holding cs_main during IncrementExtraNonce which doesn't require it.
+            unsigned int nHeight;
             {
                 LOCK(cs_main);
-                nExtraNonce = nNonceLocal;
-                IncrementExtraNonce(pblock, chainActive.Tip(), nExtraNonce);
+                nHeight = chainActive.Height();
             }
+            IncrementExtraNonce(pblock, nHeight, nExtraNonce);
```
CaveSpectre11, line 986: In this context, the cs_main lock is not needed, however the cs_main lock is actually needed for IncrementExtraNonce (or better yet a different lock)

src/miner.cpp
```diff
@@ -1011,28 +1031,29 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
                     ++pblock->nNonce;
                 }
             } else {
-                LogPrintf("%s: Unknown hashing algorithm found!\n");
+                LogPrintf("%s: Unknown hashing algorithm found!\n", __func__);
```
CaveSpectre11, line 1034: nice catch!

src/miner.cpp
```diff
@@ -996,12 +1004,24 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
...
-                    ++nTries;
-                    ++pblock->nNonce64;
+                bool success = false;
+                while (nMidTries < nMidLoopCount) {
+                    while (nTries < nInnerLoopCount &&
+                           !(success = CheckProofOfWork(pblock->GetSha256D(midStateHash), pblock->nBits,
+                                   Params().GetConsensus(), CBlockHeader::SHA256D_BLOCK))) {
+                        boost::this_thread::interruption_point();
+                        ++nTries;
+                        ++pblock->nNonce64;
```
CaveSpectre11, line 1014: Since this essentially runs it much longer, I do have some concerns over this change; but I'm not 100% in disagreement with it.  I need to give it another look on the next pass.

src/miner.cpp
```diff
@@ -1011,28 +1031,29 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
...
                 if (!nTimeDuration) nTimeDuration = 1;
                 nHashSpeed = arith_uint256(nHashes/1000/nTimeDuration).getdouble();
             }
-            LogPrint(BCLog::BLOCKCREATION, "%s: PoW Hashspeed %d kh/s\n", __func__, nHashSpeed);
 
-            if (nTries == nInnerLoopCount) {
+            LogPrint(BCLog::MINING, "%s: PoW Hashspeed %d kh/s\n", __func__,  nHashSpeed);
+            if (nMidTries == nMidLoopCount || fBlockFoundAlready) {
+                fBlockFoundAlready = false;
                 continue;
```
CaveSpectre11, line 1050: This is a generic section for all mining algos, not just Sha256.  So while it will work for that, it will have problems with the others since they don't have `fBlockFoundAlready` or an incrementing `nMidTries`.  In other words, ProgPoW would miss this check since `nTries == nInnerLoopCount` but `fBlockFoundAlready` is false, and `nMidTries` is 0; so it would assume it found a block every pass.

This makes for one NACK

src/miner.cpp
```diff
@@ -814,7 +814,7 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
     }
 }
 
-void IncrementExtraNonce(CBlock* pblock, const CBlockIndex* pindexPrev, unsigned int& nExtraNonce)
+void IncrementExtraNonce(CBlock* pblock, unsigned int nHeight, unsigned int& nExtraNonce)
 {
     // Update nExtraNonce
     static uint256 hashPrevBlock;
```
CaveSpectre11, line 820: As I hinted at (although this probably is going to show up first in the review comments), this I believe is where the cs_main is protecting.  It doesn't necessarily need to be cs_main and we could probably use a different lock to isolate to just this static, but here's the issue.

`IncrementExtraNonce` can be called from multiple mining threads, yet there's a static variable in there to store the latest hashPrevBlock and reset nExtraNonce if the block has changed.  So that introduces a race condition if two threads are coming in here at the same time, as one may change that static when the previous thread is looking at it.

Should this be locked by cs_main?  Probably not.  Should IncrementExtraNonce be locked?  Yes.  (Statics are the #1 warning sign of a need to lock)

CaveSpectre11 (changes requested):
Good stuff, Thank you Zannick.  One thing to keep in mind is that atomic changes are best for PRs.  Your 3 commits are distinctly different and therefore wouldn't be appropriate to be squashed; however if there's a problem with one of them, then it holds up the other ones.

Unfortunately this must be NACKed; I will explain inline. https://github.com/Veil-Project/veil/pull/915/commits/4f4528cbfd907fee0524afe15aa8f02529127ba4

src/miner.cpp
Zannick: Ah, I hadn't thought of the static.
https://github.com/Veil-Project/veil/blob/e8c4d290c647b8e3363de49c440587fe091e963b/src/miner.cpp#L815-L820
It looks like this is the entire usage. This could be replaced with a `std::atomic<T>::exchange`, with `nExtraNonce` being reset if the previous value was not `pblock->hashPrevBlock`. Does that seem sufficient in place of a full lock, if there's nothing else?

src/miner.cpp
Zannick: To be clear, though, running the loop longer is the main way I've dealt with the lock contention from setting up the block template and nonce.

Zannick:
Thanks, I'll get on the changes and split the required commits into separate PRs. Wasn't really sure how many PRs to make for my first PR. =)

src/miner.cpp
CaveSpectre11: The original response was much different, until I read the change description again.  I just want to double check everything that passes through when it fully exits and returns.

src/miner.cpp
CaveSpectre11: Let me ramp myself up on std::atomic(), but I'll definitely consider it [The bulk of my multi-threading development came with C, and I never really re-learned solutions with C++]

Zannick: Moved to https://github.com/Veil-Project/veil/pull/921

src/miner.cpp
Zannick: This was straightforward to fix by adding another case to the conditional, since `nTries == nInnerLoopCount` is never true for SHA with this change.

CaveSpectre11:
@Zannick It looks like there's still some stuff left in here that's not in the other two PRs.  Do you want to migrate this one, or clean up the commit and submit as a 3rd PR when it's ready?  (I'm leaning towards the latter).

Zannick:
My intent was to leave this WIP until the logging PR is in, and then clean up, but if you prefer I can close now and resubmit later.

CaveSpectre11:
> My intent was to leave this WIP until the logging PR is in, and then clean up, but if you prefer I can close now and resubmit later.

No worries; just going through them all and making sure.

codeofalltrades:
This PR can be closed correct?

Zannick:
I'll revive this after #921 is merged.

src/miner.cpp
Zannick: Fixed more simply by casing only on "success".

Zannick:
Ready for re-review.

This fixes #862 in preventing regular, frequent occurrences of the wallet getting stuck. Unfortunately I did see a very rare stuckness in a similar branch (with slightly differently changes), which had an interesting symptom of taking minutes for a mining thread to be interrupted once I ask it to stop mining. If I see it again, I'll turn all debugging on before I hit exit, see if I can see what's going on/which threads are blocked, or try other methods of finding lock contention.

codeofalltrades (approved):
ACK 2270c9499c912536df95fd300a52b8c5e58dffad
Tested SHA mining on Win10 and Linux 18.04

CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/915/commits/2270c9499c912536df95fd300a52b8c5e58dffad

PR #920, opened 2021-04-04
[Logging][Mining] Add BCLog::MINING category.

Zannick:
### Problem ###
Mining code uses the `BLOCKCREATION` logging category, which includes logging about difficulty targets, staking, zerocoins, etc, making it harder to investigate mining-specific issues.

### Root Cause ###
Using the `BLOCKCREATION` category for lots of different information.

### Solution ###
Adds a new logging category `MINING` to aid debugging mining specifically
without including block template creation logs that may be noisier.

Converts most logs in BitcoinMiner and RandomXMiner to use this category
instead of `BLOCKCREATION`.

Fixes one unlikely logging statement to include the function name.

### Bounty Payment Address ###

### Unit Testing Results ###
Tested by running and sending the command `logging [\"mining\"]` to enable.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -1011,7 +1011,7 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
                     ++pblock->nNonce;
                 }
             } else {
-                LogPrintf("%s: Unknown hashing algorithm found!\n");
+                LogPrintf("%s: Unknown hashing algorithm found!\n", __func__);
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


src/miner.cpp
```diff
@@ -932,7 +932,7 @@ void BitcoinMiner(std::shared_ptr<CReserveScript> coinbaseScript, bool fProofOfS
                 auto it = mapStakeHashCounter.find(nHeight);
                 if (it != mapStakeHashCounter.end() && it->second != nStakeHashesLast) {
                     nStakeHashesLast = it->second;
-                    LogPrint(BCLog::BLOCKCREATION, "%s: Tried %d stake hashes for block %d last=%d\n", __func__, nStakeHashesLast, nHeight+1, mapHashedBlocks.at(hashBestBlock));
+                    LogPrint(BCLog::MINING, "%s: Tried %d stake hashes for block %d last=%d\n", __func__, nStakeHashesLast, nHeight+1, mapHashedBlocks.at(hashBestBlock));
```
CaveSpectre11, line 935: This one is specific to Proof of Stake rather than Proof of Work, and relevant to watching staking rather than mining.  So this one needs to stay at BlockCreation (and perhaps will be split off from Blockcreation in a separate PR to have staking separated).  But for now, it should stay in the BLOCKCREATION group.

CaveSpectre11 (changes requested):
One small correction and then it should be good.

src/miner.cpp
Zannick: Done.

CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/920/commits/8dfd59ab5c4f2b4319f7fb7e71ddfb47acf843c3

codeofalltrades (approved):
ACK 8dfd59ab5c4f2b4319f7fb7e71ddfb47acf843c3
Tested on Win10

PR #921, opened 2021-04-04
[Mining] IncrementExtraNonce without cs_main

Zannick:
### Problem ###
While mining SHA256D, the wallet periodically gets stuck, failing to stay caught up with the chain. (At the same time, the mining significantly slows down.)

### Root Cause ###
Lock contention; SHA256D is very fast, completing its loop in subsecond time on a modern cpu, meaning every thread goes through lock-heavy sections of the block template creation code multiple times per second. In comparison, RandomX on the same machine takes minutes to complete its loop, usually being interrupted by the creation of a new block instead of reaching the maximal number of attempts.

### Solution (partial) ###
Reduce lock contention on `cs_main`. The nonce setup prior to running the hashes includes grabbing `cs_main` to perform `IncrementExtraNonce`. It takes this lock to safely get the chain tip and then to safely compare the current thread's `prevHeaderHash` against a static value, resetting the nonce to 0 and updating the static value if it doesn't match.

However, it only uses the chain tip for its height, which we can get without a lock.

Second, the nonce being reset to 0 is thread-local; other threads are not affected by the first thread to reach this part in a new block. Also, the nonce base is constantly increasing in a separate section, so there does not really appear to be a requirement that nonces start from 0 on a new block; the lock on `nNonce_base` already assures us that we get a unique number for each thread. As such, I believe we can safely remove the nonce reset here. (I think we can also remove the extra increment, but I've left it in for now.)

Cleanup: Also removes one extraneous copy of the nonce into a thread local variable
(there were previously two, and I suspect the extra was already being optimized away).

### Bounty PR ###
#862

### Bounty Payment Address ##

### Unit Testing Results ###
`test_veil.exe` appears to crash on my system, so manually ran all the tests via a script and compared to master. No differences.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -814,19 +814,13 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
     }
 }
 
-void IncrementExtraNonce(CBlock* pblock, const CBlockIndex* pindexPrev, unsigned int& nExtraNonce)
+void IncrementExtraNonce(CBlock* pblock, unsigned int nHeight, unsigned int& nExtraNonce)
 {
     // Update nExtraNonce
-    static uint256 hashPrevBlock;
-    if (hashPrevBlock != pblock->hashPrevBlock)
-    {
-        nExtraNonce = 0;
-        hashPrevBlock = pblock->hashPrevBlock;
-    }
     ++nExtraNonce;
-    unsigned int nHeight = pindexPrev->nHeight+1; // Height first in coinbase required for block.version=2
     CMutableTransaction txCoinbase(*pblock->vtx[0]);
-    txCoinbase.vin[0].scriptSig = (CScript() << nHeight << CScriptNum(nExtraNonce)) + COINBASE_FLAGS;
+    // Height first in coinbase required for block.version=2
+    txCoinbase.vin[0].scriptSig = (CScript() << (nHeight+1) << CScriptNum(nExtraNonce)) + COINBASE_FLAGS;
     assert(txCoinbase.vin[0].scriptSig.size() <= 100);
 
     pblock->vtx[0] = MakeTransactionRef(std::move(txCoinbase));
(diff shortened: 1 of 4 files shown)


CaveSpectre11 (approved):
With the explanation it does make sense, and I don't see anything structurally wrong with it.  I'm going to utACK https://github.com/Veil-Project/veil/pull/921/commits/a3168e74a1ef886d220e771ea9974b8ca582fb1a

But I definitely want @codeofalltrades to give it a run for it's money, and it also needs some QA time as well.

seanPhill:
I am running #921 on mainnet and just checked it (after a couple of days of running) and found it lagging by several minutes. All the connected peers were showing sync further behind. I had been running 5 out of 6 cores on SHA, while using the computer for Twitter and Discord chat. I stopped mining, and immediately it caught up several blocks and then the peers tab showed the peers were caught up. Evidently the peers' sync data shown was old.

Zannick:
That (the wallet falling behind while mining SHA) sounds like a symptom of #862, which this doesn't fully address. Potentially "several minutes behind" is an improvement over my experience of being full-on stuck. Interesting to note that it might be the peers' sync data that is getting stuck on the other end, I may look into that separately.

So I think that's not caused by this PR, but if the same thing happens with RandomX mining, please do let me know.

seanPhill:
Okay. I'll switch to RandomX on mainnet and testnet. I've never done very well on these computers with RandomX, but with SHA difficulty so much higher than before (with so many new in-wallet miners) I didn't get any mainnet SHA in these two days anyway. (My iMac and Mac Pro are both from late-2013. i7 and XEON E5 respectively.)

seanPhill:
I just accidentally quit the mainnet wallet while it was mining RandomX and I got 

> $ ./veil-qt
> Assertion failed: (!setBlockIndexCandidates.empty()), function PruneBlockIndexCandidates, file validation.cpp, line 3467.
> Abort trap: 6

It's been a while since I've had that. I thought something else had the foreground when I quit.

Zannick:
Interesting. I don't see a way this might be related to the changes here, but I'll take a look at the issue regardless.

codeofalltrades (approved):
ACK a3168e74a1ef886d220e771ea9974b8ca582fb1a

codeofalltrades:
Running for 4 days
Machine 1: 1 core mining on a 2 core machine 
Machine 2: 2 cores mining on a 6 core machine; Change today to 4 of 6 cores mining.

codeofalltrades:
Running for 7 days
Machine 1: 1 core mining on a 2 core machine
Machine 2: 4 of 6 cores mining to much, got stuck. 3 of 6 was fine.
QA Complete for me.

Zannick:
4 would probably work with RandomX (which is affected by this change even though the main intent behind it is an improvement to SHA).

seanPhill:
Complete and fine for me. My testnet RandomX ran fine with 4 threads on a four core (old) i7.
SHA was fine on my usual 4 of 6 cores, but I didn't run it as long.
Had a power outage this morning. Busily mining RandomX on testnet was fine. Mainnet wallet was fine after two restarts, but had not been mining at the time of the power outage, as it was just resyncing from being off for a while.

Issue #867, opened 2020-11-08
[Mining] Provide a way to mine to a specific wallet address


dragonmike:
the parameter -miningaddress (tested in command line and as conf file entry) followed by a bv- or V- address does not force the wallet to mine randomX towards given address. The wallet still mines to a randomly generated mining address.

 the wallet should have mined towards the given address

 the wallet generated a new mining address and mined towards it, ignoring the set parameter

 tested several times, easy to reproduce

 v1.1.0 from github

 Windows 10, Ryzen 9 3900, Sandisk SSD

 n/a

 n/a

CaveSpectre11:
To follow with a note on this:

`miningaddress` is used by progpow's `getblocktemplate` and is confusing as it is not originally intended to be a target for all mining payouts.  As that doesn't override the feature request intended here [for a use case that has a central management wallet with the private keys for a users distributed mining wallets], the feature is not without merit.  So the title will be renamed to be more in line with the feature.

Zannick:
*Should* `miningaddress` be the address the wallet uses for in-wallet mining, or should we have separate options, e.g., `walletminingaddress` and `rpcminingaddress`?

CaveSpectre11:
> _Should_ `miningaddress` be the address the wallet uses for in-wallet mining, or should we have separate options, e.g., `walletminingaddress` and `rpcminingaddress`?

I'm not a miner, so I'll leave that to those that are to answer the following question:  "What is the use case for two separate addresses?"

Zannick:
The use case I was picturing was doing some local cpu mining plus the rpc mining, but I seem to have forgotten that you can't mine two different algorithms with one wallet. So, if it's acceptable to expand the usage of miningaddress for in-wallet mining, this ends up looking pretty simple.

CaveSpectre11:
Bounty assigned: @Zannick

Zannick:
Included in versions >= 1.2

PR #923, opened 2021-04-07
[Mining] Use -miningaddress for in-wallet mining.

Zannick:
### Problem ###
`-miningaddress` is used for rpc progpow gpu mining, but ignored for in-wallet mining such as for RandomX and SHA256D.

### Root Cause ###
It's just not looking at the config value when it sets up the block script.

### Solution ###
When the wallet attempts to retrieve a mining address, prefer the configured value for `-miningaddress`, falling back to the previous address generation from the wallet keys.

Disallows stealth addresses as per current behavior validating `-miningaddress` at startup and in rpc mining.

Performs an additional safety check that the wallet actually owns the given address, falling back to a random wallet address otherwise (preserving current behavior).

### Bounty PR ###
#867

### Bounty Payment Address ##

### Testing Results ###
Tested on win10 x64 mining RandomX on testnet: both with basecoin and "mining" addresses.
Tested on win10 veild, veil-qt: Using a testnet basecoin address that was not mine. Successfully produced an error and exited.

RandomX mining results

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1933,6 +1933,17 @@ bool AppInitMain()
         // Link thread groups
         LinkAutoSpendThreadGroup(&threadGroupAutoSpend);
 
+        // Validate wallet mining address.
+        std::string sAddress = gArgs.GetArg("-miningaddress", "");
+        if (!sAddress.empty()) {
+            CTxDestination dest = DecodeDestination(sAddress);
+
+            auto pt = GetMainWallet();
+            if (pt && pt->IsMine(dest) != ISMINE_SPENDABLE) {
+                return InitError(strprintf(_("Invalid -miningaddress: '%s' not owned by wallet"), sAddress));
+            }
+        }
+
         // Start wallet CPU mining if the -gen=<n> parameter is given
         int nThreads = gArgs.GetArg("-gen", 0);
         if (nThreads) {
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -4702,7 +4702,7 @@ void CWallet::MarkReserveKeysAsUsed(int64_t keypool_id)
     }
 }
 
-void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
+void CWallet::GetScriptForMiningReserveKey(std::shared_ptr<CReserveScript> &script)
 {
     std::shared_ptr<CReserveKey> rKey = std::make_shared<CReserveKey>(this);
     CPubKey pubkey;
@@ -4711,6 +4711,19 @@ void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
 
     script = rKey;
     script->reserveScript = CScript() << ToByteVector(pubkey) << OP_CHECKSIG;
+}
+
+void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
+{
+    std::string sAddress = gArgs.GetArg("-miningaddress", "");
+    if (!sAddress.empty()) {
+        CTxDestination dest = DecodeDestination(sAddress);
+
+        script = std::make_shared<CReserveScript>();
+        script->reserveScript = GetScriptForDestination(dest);
+        return;
+    }
+    return GetScriptForMiningReserveKey(script);
 }
 
 void CWallet::LockCoin(const COutPoint& output)
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1221,6 +1221,7 @@ public:
 
     const std::string& GetLabelName(const CScript& scriptPubKey) const;
 
+    void GetScriptForMiningReserveKey(std::shared_ptr<CReserveScript> &script);
     void GetScriptForMining(std::shared_ptr<CReserveScript> &script);
 
     unsigned int GetKeyPoolSize() EXCLUSIVE_LOCKS_REQUIRED(cs_wallet)


codeofalltrades (approved):
ACK bcdd0f5524e201caee1ea1eec01569a41ecd5ba9

CaveSpectre11 (commented):
I'm having a hell of a time downloading a binary to test this, so I'll just ask a few questions:

1) If you specify `-miningaddress` with no value, it will just use a generic one and not give you an error?
2) If the destination is invalid, same thing?
3) When do you see these errors?

I think I might rather see the validity tests for -miningaddress in veild.cpp and qt/veil.cpp, so that they get feedback when they try to start the daemon or qt; and have it not continue if they don't have a valid basecoin address that is theirs.  Having the differentiation between basecoin and stealth is good, so we have a hook when we do allow stealth; but if it doesn't fail on startup until they correct their use of -miningaddress, it should.

If I can get the thing to download, I'll get some of my own answers to the above questions.

src/wallet/wallet.cpp
```diff
@@ -4713,6 +4713,31 @@ void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
...
+            if (IsMine(dest) != ISMINE_SPENDABLE) {
+                WalletLogPrintf("-miningaddress %s not under my control, falling back to random address", sAddress);
+                return GetScriptForMiningReserveKey(script);
+            }
+            // Disallow Stealth Addresses for now
+            CBitcoinAddress address(sAddress);
+            if (address.IsValidStealthAddress()) {
+                WalletLogPrintf("-miningaddress must be a basecoin address");
+                return GetScriptForMiningReserveKey(script);
+            }
```
CaveSpectre11, line 4732: This is unnecessary since this check is being done in veild.cpp and qt/veil.cpp

src/wallet/wallet.cpp
```diff
@@ -4713,6 +4713,31 @@ void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
...
+{
+    std::string sAddress = gArgs.GetArg("-miningaddress", "");
+    if (!sAddress.empty()) {
+        CTxDestination dest = DecodeDestination(sAddress);
+
+        if (IsValidDestination(dest)) {
+            if (IsMine(dest) != ISMINE_SPENDABLE) {
+                WalletLogPrintf("-miningaddress %s not under my control, falling back to random address", sAddress);
+                return GetScriptForMiningReserveKey(script);
+            }
```
CaveSpectre11, line 4726: These are good checks, but they should be done where miningaddress is being checked on startup.

src/wallet/wallet.cpp
```diff
@@ -4713,6 +4713,31 @@ void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
     script->reserveScript = CScript() << ToByteVector(pubkey) << OP_CHECKSIG;
 }
 
+void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
```
CaveSpectre11, line 4716: See other comments, this should only load it if it exists; the validity checks should be on startup so they know right when they execute veild or veil-qt that they have a mistake in their -miningaddress, and not have to wait until later (or in the case of veild, not even notice that they didn't) get what they wanted.  I.e. find where it's checked in veild.cpp and qt/veil.cpp and add the additional checks in there.  We shouldn't fall over to do something they didn't intend when they used the flag incorrectly.

CaveSpectre11 (changes requested):
See changes needed.

src/wallet/wallet.cpp
Zannick: It's also being done in rpc/mining.cpp--I presume as a backstop in case the arg was changed between startup and the rpc to start mining. If it's changed, and manages to get by here, the script will be empty, and I think that will make it fail when it tries to commit a block (I tried mining to stealth on testnet at some point and iirc that's what happened). Perhaps that is user error, since we should not be mutating miningaddress in-process? The difference is that we can't return an RPC error here, and I felt it was better to fall back to existing behavior (mining to an unexpected address) rather than surprising behavior (mining fails at the last possible moment).

src/wallet/wallet.cpp
Zannick: Done.

src/wallet/wallet.cpp
Zannick: Had to move it to init.cpp so I could be sure the wallet was loaded, but it works for both veild and veil-qt (with a delay to load the wallet, of course). Done.

src/wallet/wallet.cpp
Zannick: Cleared these checks out.

Zannick (commented):
1. Supplying an empty miningaddress is the same as providing none, since the `getArg` lookups provide a default of `""`.
2. As you found, yeah, the errors are on startup. See my comments below..

src/wallet/wallet.cpp
CaveSpectre11: Ahh yes, because of the IsMine().  Let me think about this a little bit more.  I put in a good address that I didn't own, and I didn't see an error anywhere.  I'm building this PR so I can do some extra debugging later.

src/init.cpp
```diff
@@ -1933,6 +1933,17 @@ bool AppInitMain()
...
         LinkAutoSpendThreadGroup(&threadGroupAutoSpend);
 
+        // Validate wallet mining address.
+        std::string sAddress = gArgs.GetArg("-miningaddress", "");
+        if (!sAddress.empty()) {
+            CTxDestination dest = DecodeDestination(sAddress);
+
+            auto pt = GetMainWallet();
+            if (pt && pt->IsMine(dest) != ISMINE_SPENDABLE) {
+                return InitError(strprintf(_("Invalid -miningaddress: '%s' not owned by wallet"), sAddress));
```
CaveSpectre11, line 1943: This has an issue; as it hangs up the wallet and doesn't exit:
```
2021-04-20T23:24:24Z ThreadStakeMiner: starting
2021-04-20T23:24:24Z Veil Miner started
2021-04-20T23:24:24Z Error: Invalid -miningaddress: 'bv1q7cmcjcazyxger9y9llnqpuhj33pmrn9hl0ny24' not owned by wallet
2021-04-20T23:24:24Z tor: Thread interrupt
2021-04-20T23:24:24Z torcontrol thread exit
2021-04-20T23:24:24Z opencon thread exit
2021-04-20T23:24:24Z dnsseed thread exit
2021-04-20T23:24:24Z Shutdown: In progress...
2021-04-20T23:24:24Z addcon thread exit
2021-04-20T23:24:24Z msghand thread exit
2021-04-20T23:24:24Z net thread exit
2021-04-20T23:24:24Z Imported mempool transactions from disk: 3 succeeded, 0 failed, 0 expired, 0 already there
2021-04-20T23:24:37Z ThreadStakeMiner: interrupted
2021-04-20T23:24:37Z Veil Miner started
```
```
./veil-cli stop
error: Could not connect to the server 127.0.0.1:58812

Make sure the veild server is running and that you are connecting to the correct RPC port.
```

Zannick: This appears to be an existing issue. I can reproduce it here:
https://github.com/Veil-Project/veil/blob/02cc068b03abe4aeac551762f24ba621819ad733/src/init.cpp#L1972-L1974

Let me file a separate issue...

Zannick: I filed PR #935 with a fix.

codeofalltrades (approved):
ACK b2eb97778cbd7390e16dd38e6308d73962b2277e
Tested on Win 10

CaveSpectre11 (approved):
ACK b2eb97778cbd7390e16dd38e6308d73962b2277e

But #935 must be merged first.

PR #925, opened 2021-04-10
[WIP] Wallet Dark Mode

Rock-N-Troll:
Dark Mode Work now being handled in pull request #933 

**WIP:**
Feedback Welcome!

Enable darkmode using `darkmode=true` in walletconfig. 
Default wallet styling otherwise.

- Currently refactoring so I don't break any pre-existing normal wallet styling
- Currently refactoring and restyling dark mode. Looking for suggestions
- Opening pull request to support github actions to build for anyone curious or wanting to check the latest build with test wallet (once builds are working)

All colors and styling subject to change.

image

image

image

image

image

image

**Development To-Do:**
- [x] Missing line separators in settings widget page
- [x] Make outer window dark
- [ ] Make black text gray
- [x] Apply dark mode to Transaction List Filter 
- [x] Apply dark mode to Scroll Bar
- [ ] Apply dark mode to debug window
- [ ] Apply dark mode to send screen buttons on bottom left
- [ ] Move configuration from config file to settings pop-up as a checkbox

**Final To-Do:**
- [ ] Make changes/confirm normal styling is working
- [ ] Make changes/Confirm dark mode styling is working
- [ ] Get builds working (especially you Mac)


seanPhill:
I've tested this on MacOS Catalina with Dark Mode turned on in the OS, but without the darkmode=true in the config file (yet).
Most things are as normal, but the settings page was behaving badly, changing to dark mode line by line, it seems, as I moused over it. Then when I selected Preferences, the Settings page cleaned up (back to light mode) and remained that way until selecting another tab and returning. (I haven't seen this before, but I will double check on the same computer with a released version.)

Rock-N-Troll:
> I've tested this on MacOS Catalina with Dark Mode turned on in the OS, but without the darkmode=true in the config file (yet).
> Most things are as normal, but the settings page was behaving badly, dark-greyed over and then line by line with gaps in between as I moused over it. Then when I selected Preferences, the Settings page cleaned up (back to full light mode) and remained that way until selecting another tab and returning. (I haven't seen this before, but I will double check on the same computer with a released version.) (Strangely, the colours are inverted in the shaded area of that screenshot compared to how it actually looked on the other Mac. It was light on that one line and dark grey over the rest.)
> pr925-settingsPage


yes that is one place that the styling was changed that I will need to address. Unfortunately it will be a lot of trial and error to get it working as I am trying to break out the embedded styling from the .ui files into the .css files for both normal and dark mode.

I'm hoping to get dark-mode working somewhat (still would like some help on design) and then go back and meticulously make sure the wallet styling is the same as before, at least on my system which is windows. I might have to do some refactoring as well as some of the styling is broken or duplicated many times but different making it difficult to determine which styling is "correct".

Thank you for testing and the feedback :)

seanPhill:
Looking even better than your earlier screenshots when I have started with `darkmode=1`. I guess you've already enhanced the legibility in darkmode.

Rock-N-Troll:
I'm going to refactor the in-line css code that exists in the .ui files into main.css so we can all maintain the stylings in 1 place, especially for easily handling dark mode (and reduce unnecessary repeated stylings)

Issue #934, opened 2021-04-21
[Core] Some veild errors during init result in hangs


Zannick:
```
2021-04-20T23:24:24Z ThreadStakeMiner: starting
2021-04-20T23:24:24Z Veil Miner started
2021-04-20T23:24:24Z Error: Invalid -miningaddress: 'bv1q7cmcjcazyxger9y9llnqpuhj33pmrn9hl0ny24' not owned by wallet
2021-04-20T23:24:24Z tor: Thread interrupt
2021-04-20T23:24:24Z torcontrol thread exit
2021-04-20T23:24:24Z opencon thread exit
2021-04-20T23:24:24Z dnsseed thread exit
2021-04-20T23:24:24Z Shutdown: In progress...
2021-04-20T23:24:24Z addcon thread exit
2021-04-20T23:24:24Z msghand thread exit
2021-04-20T23:24:24Z net thread exit
2021-04-20T23:24:24Z Imported mempool transactions from disk: 3 succeeded, 0 failed, 0 expired, 0 already there
2021-04-20T23:24:37Z ThreadStakeMiner: interrupted
2021-04-20T23:24:37Z Veil Miner started
```
```
./veil-cli stop
error: Could not connect to the server 127.0.0.1:58812

Make sure the veild server is running and that you are connecting to the correct RPC port.
```

_Originally posted by @CaveSpectre11 in https://github.com/Veil-Project/veil/pull/923#discussion_r617101114_

@Zannick:
> This appears to be an existing issue. I can reproduce it here:
> https://github.com/Veil-Project/veil/blob/02cc068b03abe4aeac551762f24ba621819ad733/src/init.cpp#L1972-L1974
> 
> Let me file a separate issue...

PR #935, opened 2021-04-21
[Core] Properly shut down veild if init fails.

Zannick:
### Problem ###
veild doesn't shut down properly if it fails during init.

### Root Cause ###
Background threads were hanging after being interrupted, because they were looking to `ShutdownRequested()`, which was only being set by `StartShutdown()` on SIGTERM.

### Solution ###
Solution is to call `StartShutdown()` when veild fails init and goes to interrupt running threads.

### PR ###
#934 

### Testing ###
Set `autospend=1` and `autospendaddress=<basecoin address>` (or make a similar error that should cause it to close with an error on startup) in your veil.conf and run veild with `-printtoconsole`.

### DIFF
--- a/src/veild.cpp
+++ b/src/veild.cpp
@@ -199,6 +199,7 @@ static bool AppInit(int argc, char* argv[])
 
     if (!fRet)
     {
+        StartShutdown();
         Interrupt();
     } else {
         WaitForShutdown();


codeofalltrades (approved):
ACK a66022a5dfe4af4abd903ce5de5fcbb3f7e85c44
Tested on Win 10

CaveSpectre11 (approved):
ACK a66022a5dfe4af4abd903ce5de5fcbb3f7e85c44

Must be merged before #923

PR #937, opened 2021-04-25
[Trivial] Add config include for building DEBUG_LOCKCONTENTION

Zannick:
config/veil-config.h includes `HAVE_THREAD_LOCAL`, without which `DEBUG_LOCKCONTENTION` will fail to build due to the `static_assert`.

This allows building with `./configure CPPFLAGS="-DDEBUG_LOCKCONTENTION"` without manually adding `-DHAVE_THREAD_LOCAL` (when thread_local exists, of course).

### Tested ###
Built via `./configure --with-incompatible-bdb CPPFLAGS="-DDEBUG_LOCKCONTENTION"` on Fedora, `CONFIG_SITE=$PWD/depends/x64_64-w64-mingw32/share/config.site ./configure --prefix=/ CPPFLAGS="-DDEBUG_LOCKCONTENTION"` on WSL Ubuntu 20.04

### DIFF
--- a/src/sync.cpp
+++ b/src/sync.cpp
@@ -4,6 +4,10 @@
 
 #include <sync.h>
 
+#ifdef HAVE_CONFIG_H
+#include <config/veil-config.h>
+#endif
+
 #include <logging.h>
 #include <utilstrencodings.h>
 


CaveSpectre11:
This  might work.  I'll need to dig up the weirdness that happened with OSX when I made the option available.

CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/937/commits/f86a61837e2b1cb562dfdfe20b2c80f422c2aac2

I was confusing a couple different issues; the issue I'm thinking of was in the debug logging on MAC, the issue I had with DEBUG_LOCKCONTENTION is it sometimes flags things as potential deadlocks that aren't; which is why I stopped running with it on except when I was specifically looking for something.

codeofalltrades (approved):
ACK f86a61837e2b1cb562dfdfe20b2c80f422c2aac2

PR #939, opened 2021-05-02
[RPC][Trivial] Tweak exportzerocoins example to be sensible

CaveSpectre11:
### Problem
exportzerocoins help gives an example with a denomination of 5; which doesn't make any sense.

```
Examples:
> veil-cli exportzerocoins false 5
> curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "exportzerocoins", "params": [false 5] }' -H 'content-type: text/plain;' http://127.0.0.1:58812/
```

### Solution
Tweak it to show an example of 10

```
Examples:
> veil-cli exportzerocoins false 10
> curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "exportzerocoins", "params": [false 10] }' -H 'content-type: text/plain;' http://127.0.0.1:58812/
```

### DIFF
--- a/src/wallet/rpczerocoin.cpp
+++ b/src/wallet/rpczerocoin.cpp
@@ -1,5 +1,5 @@
 // Copyright (c) 2015-2019 The PIVX developers
-// Copyright (c) 2019-2020 The Veil developers
+// Copyright (c) 2019-2021 The Veil developers
 // Distributed under the MIT software license, see the accompanying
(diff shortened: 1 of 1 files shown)


codeofalltrades (approved):
utACK 3cf9ddf10417ae97d4158f2efade9d025967c73c

PR #944, opened 2021-05-16
[Consensus][RingCT] Add check for Duplicate Key Images

CaveSpectre11:
### Problem
Duplicate Key Images are not checked across transactions in the same block when validating a block

### Solution
Keep a list of Key Images while checking the TXs in a block; reject the block if there is a duplicate key index used


codeofalltrades (approved):
ACK 53bb35e39691ad7d760e142a42cf629215042ddf

PR #945, opened 2021-05-16
[Consensus][RingCT] Remove additional mempool check

CaveSpectre11:
### Problem
In the v1.2.0 code, an experimental addition of commented out code was errantly included in the commit, which created a situation where a node will reject a block if there is a duplicate key image being used in a transaction that has been broadcast to the mempool.  That transaction in mempool would not be included into a block, but the code was causing the block with the duplicate key image to be rejected and the node submitting the block to be banned.

### Solution
Take the code section out.

### Additionally
- This is built on top of v1.2.0, (PR #944) and contains the code for v1.2.0 also.  
- A scenario where no key images were found in the transaction could result in the node crashing with an assert.  That assert has been removed.
- RPC calls `gettransaction` and `getrawtransaction` have been modified to include key_image information for RingCT transactions.


codeofalltrades (approved):
ACK 53bb35e39691ad7d760e142a42cf629215042ddf

PR #946, opened 2021-05-17
[GUI][Zerocoin] transaction list - handle receive from zerocoin transactions

codeofalltrades:
### Problem ###
The wallet transaction list was updated to show the correct address #898 
Zerocoins sent to a bv1 address were causing the force_return error. The address format is bech32 however the decode method  was trying to decode the address as bech32 = false. 

### Root Cause ###
The address format is bech32 however the decode method  was trying to decode the address as bech32 = false. 

### Solution ###
Decode the address using the correct format.

### Unit Testing Results ###
1. Create a new wallet
2. Get a basecoin address
3. Spend Zerocoins from a different wallet to the basecoin address
4. Verfiy the tx(s) are confirmed in the new wallet
5. Restart the wallet(veil-qt) with -reindex-chainstate
6. Verify the force_return error does not occur

### DIFF
--- a/src/qt/transactionrecord.cpp
+++ b/src/qt/transactionrecord.cpp
@@ -1,6 +1,6 @@
 // Copyright (c) 2011-2019 The Bitcoin Core developers
 // Copyright (c) 2011-2019 The Particl developers
-// Copyright (c) 2018-2019 Veil developers
+// Copyright (c) 2018-2021 Veil developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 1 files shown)


src/qt/transactionrecord.cpp
```diff
@@ -411,7 +411,11 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const interface
             if (wtx.txout_address_is_mine[i]) {
                 // Received by Bitcoin Address
                 sub.type = TransactionRecord::RecvWithAddress;
-                sub.address = EncodeDestination(wtx.txout_address[i]);
+                bool fBech32 = false;    
+                if (boost::get<CScriptID>(&wtx.txout_address[i])){
+                    fBech32 = true;
+                }
```
Zannick, line 417: Can this be condensed into a single line instead of the if? say, something like
`bool fBech32 = (bool)boost::get<CScriptID>(...);` or `boost::get<CScriptID>(...) != ...` ?

CaveSpectre11: That's true:
`bool fBech32 = boost::get<CScriptID>(&wtx.txout_address[i])` would do the trick; but functionally it's the same and and it's not against style; so I'll say that's a NIT

Another NIT: space between the end of the if and the `{`

But given problems with this problem, I'm not concerned with holding it up.

CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/946/commits/352d78e4c37ea60b6ea542c07142c1bb33c1db45

WetOne:
utACK 352d78e4c37ea60b6ea542c07142c1bb33c1db45

WetOne (approved):

PR #948, opened 2021-05-18
[Utils] Templatify AssertLockHeld to placate clang warning

Zannick:
By changing the type of the lock from void* (which has no capability) to the actual type (which does).

(Templates more or less have to be defined in a header, included in a header, or specializations explicitly declared in a header, so the actual implementation here is to call the original function that took a void*.)

### Test Results

Built and run veild/veil-cli with clang. Built veild/veil-cli with clang with `-DDEBUG_LOCKORDER`.

### DIFF
--- a/src/sync.cpp
+++ b/src/sync.cpp
@@ -152,7 +152,7 @@ std::string LocksHeld()
     return result;
 }
 
-void AssertLockHeldInternal(const char* pszName, const char* pszFile, int nLine, void* cs)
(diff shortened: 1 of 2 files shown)


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/948/commits/185c71fab114521dd03cf9735e737e27ee11559c

codeofalltrades (approved):
ACK 185c71fab114521dd03cf9735e737e27ee11559c

PR #949, opened 2021-05-18
[Wallet] Lock cs_wallet in additional places.

Zannick:
### Problem
tsan identifies a data race within WalletBatch, commonly between KeepKey and AddToWallet.

### Solution
Declare CWallet::KeepKey as requiring `cs_wallet`, and take it before all calls as prescribed by clang. (This moved the tsan data race to the nearby fields of CReserveKey, which was mitigated by including their access within the lock.)
Add another cs_wallet use per clang warning: AddKeyPubKey requires holding mutex 'pwalletParent->cs_wallet' exclusively.

### Tested
Configured with --with-sanitizers=thread, built with clang, and run on regtest. tsan warnings for this area no longer appear.

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -4527,6 +4527,7 @@ bool AnonWallet::AddStealthDestinationMeta(const CKeyID& idStealth, const CKeyID
 
 bool AnonWallet::AddKeyToParent(const CKey& keySharedSecret)
 {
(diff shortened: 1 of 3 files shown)


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/949/commits/c9cdb225158fe1c115af6dd60c127d33abdc5f6a

codeofalltrades (approved):
ACK c9cdb225158fe1c115af6dd60c127d33abdc5f6a

PR #950, opened 2021-05-18
[Consensus] Add locking on the chain index vector.

Zannick:
### Problem
tsan identified a data race on CChain, particularly concurrent access to the underlying vector's `operator[]`, `size`, and `resize` member functions.

### Solution
Create a new lock `cs_vchain` and mark `vchain` as guarded by it, and ensure that the lock is taken prior to any use.
For the `operator==` member function, grab both the locks in a prescribed order to avoid a deadlock.

### Tested
Configured with --with-sanitizers=tsan, built with clang, run on regtest with 2 SHA mining threads enabled.

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -16,6 +16,7 @@
  * CChain implementation
  */
 void CChain::SetTip(CBlockIndex *pindex) {
+    LOCK(cs_vchain);
     if (pindex == nullptr) {
         vChain.clear();
         return;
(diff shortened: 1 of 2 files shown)


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/950/commits/3cf8acfefeb5f9b924660548367317462829e107

WetOne:
When does cs_vchain get unlocked?

Zannick:
When the lock object is destroyed, when it goes out of scope.

WetOne (approved):

WetOne:
utACK 3cf8acfefeb5f9b924660548367317462829e107

codeofalltrades (approved):
ACK 3cf8acfefeb5f9b924660548367317462829e107

PR #952, opened 2021-05-18
[Wallet][Consensus] Resolve misc clang lock annotation warnings.

Zannick:
### Problem
Building with clang results in lots of warnings about lock annotations not being obeyed.

### Solution
Follow instructions on warnings:
- Replace double lock of cs_main with an assertion.
- Reorganize code to release cs_main before calling blockheaderToJSON.
- Lock cs_main before calling any of: GetBlocksToMaturity, GetAddressBalances, IsSpent, TestBlockValidity.

### Tested
Configured with --with-sanitizers=thread, built with clang, run on regtest.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -960,29 +960,35 @@ static UniValue getblockheader(const JSONRPCRequest& request)
             + HelpExampleRpc("getblockheader", "\"00000000c937983704a73af28acdec37b049d214adbda81d7e2a3dd146f6ed09\"")
         );
 
-    LOCK(cs_main);
-
-    std::string strHash = request.params[0].get_str();
-    uint256 hash(uint256S(strHash));
-
-    bool fVerbose = true;
-    if (!request.params[1].isNull())
-        fVerbose = request.params[1].get_bool();
-
-    const CBlockIndex* pblockindex = LookupBlockIndex(hash);
-    if (!pblockindex) {
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block not found");
-    }
-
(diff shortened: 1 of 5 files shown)


src/rpc/blockchain.cpp
```diff
@@ -960,29 +960,35 @@ static UniValue getblockheader(const JSONRPCRequest& request)
             + HelpExampleRpc("getblockheader", "\"00000000c937983704a73af28acdec37b049d214adbda81d7e2a3dd146f6ed09\"")
         );
 
-    LOCK(cs_main);
+    const CBlockIndex* pChainTip;
+    const CBlockIndex* pblockindex;
+    {
+        LOCK(cs_main);
```
CaveSpectre11, line 966: This can definitely move further down to after validating the parameters; but it definitely needs to be taken before LookupBlockIndex() since it has an AssertLockHeld (but of course see commentary in the mapBlockIndex possibly being it's own lock, in which case we wouldn't need cs_main until it's protecting the chain tip).

codeofalltrades (approved):
ACK 8d58da5af339125d1751f8e7f4eaa3ff21784cf0

codeofalltrades:
@CaveSpectre11 is your review a NIT or need to change?

src/rpc/blockchain.cpp
Zannick: Yeah, we can tighten this up separately, or with the mapBlockIndex PR (update pending), but this is the minimal change needed to accomplish the unlocking.

CaveSpectre11: acceptable

CaveSpectre11 (approved):
utACK ab801ee

PR #954, opened 2021-05-22
[Logging] Improve hash mismatch warning log message.

Zannick:
Previously, it was logging "Block 2: ..." when 2 was the status, which is fairly misleading. Now, output both the height and the status and make clear which is which.

### Tested
Mildly, during my other regtest runs, and startup for testnet.


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/954/commits/8626e8f782de3995a69bfc82f064980e5f11c031

codeofalltrades (approved):
ACK 8626e8f782de3995a69bfc82f064980e5f11c031

PR #955, opened 2021-05-23
arith: fix some constructor issues.

sinetek:
Here is a patch regarding a few issues with the way arith_uint256 is used.
The constructor is called directly to set the initial value, and there's a typo in the name of a class.


Zannick:
utACK 8eb0add
(I guess we're not compiling with `USE_NUM_OPENSSL` or we might have caught the misnamed function.)

Zannick (approved):

CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/955/commits/8eb0add48b095ab99915c0a2f3ffb04171c7b209

PR #956, opened 2021-05-23
Avoid non-trivial global constants in SHA-NI code

Zannick:
This is a cherry-pick of https://github.com/bitcoin/bitcoin/pull/18553 fixing https://github.com/bitcoin/bitcoin/issues/18456 which suddenly cropped up on my linux 18.04 machine.

### Problem
Certain linux configurations may encounter a `SIGILL` Illegal Instruction that causes a locally-built veild to crash immediately. Using gdb to catch the stack trace will give a reference to global variable initialization in the sha256_shani.cpp file. Root cause appears to be a global constant is not produced correctly by the compiler as a constant.

### Solution
Cherry-pick bitcoin's solution to the same problem.

### Testing
Verified that veild (regtest) is now able to run on the machine where this issue occurred.

### DIFF
--- a/src/crypto/sha256_shani.cpp
+++ b/src/crypto/sha256_shani.cpp
@@ -16,9 +16,9 @@
 
 namespace {
 
-const __m128i MASK = _mm_set_epi64x(0x0c0d0e0f08090a0bULL, 0x0405060700010203ULL);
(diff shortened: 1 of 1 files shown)


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/956/commits/9c98451169623ee09ec2365e95816abb213a90e3

codeofalltrades (approved):
ACK 9c98451169623ee09ec2365e95816abb213a90e3

Issue #930, opened 2021-04-13
Assertion failed. Wallet.dat corrupted. Quit mainnet macOS wallet while busy RandomX.


seanPhill:
I just accidentally quit the mainnet wallet while it was mining RandomX and (on two restarts) I got (this both times)

> $ ./veil-qt
> Assertion failed: (!setBlockIndexCandidates.empty()), function PruneBlockIndexCandidates, file validation.cpp, line 3467.
> Abort trap: 6

It's been a while since I've had that. I thought something else had the foreground when I quit. (I'll check where that problem belongs. Thought it had been fixed.) (Note: snapshot, and reindex-chainstate did not fix this. Wallet.dat was corrupted.)
The repeated error after the chainstate reindex attempt was 
> $ ./veil-qt -reindex-chainstate
> Assertion failed: (false), function forced_return, file /home/runner/work/veil/veil/source/depends/x86_64-apple-darwin14/share/../include/boost/variant/detail/forced_return.hpp, line 47.

The related past issues were #405 and #371. 405 had been closed, presumed fixed in release v1.0.3.0, and those instances had been fixed with  `-reindex-chainstate`. The 371 version of this problem had happened on Linux, had failed to be fixed with a snapshot, and -reindex-chainstate had also failed. The only solution had been to abandon the wallet.dat.

_Originally posted by @seanPhill in https://github.com/Veil-Project/veil/pull/921#issuecomment-818981830_

seanPhill:
This was an empty mining wallet. Rewards had always been spent out to staking wallets. So I have not tried to restore from seed words, only set up a new, clean wallet.

Zannick:
Were there regular backups of the wallet that were usable?

Zannick:
https://github.com/Veil-Project/veil/blob/e8c4d290c647b8e3363de49c440587fe091e963b/src/validation.cpp#L3463-L3468

If I understand the context right, this is trimming away candidate chains that have less work than the current best tip, where the "less work" comparison is actually effectively comparing a tuple (work, sequence id, pointer value) [noting that pointer value is only supposed to be relevant when reading indexes from disk, where sequence id would be zero]. It's only possible for this assert to trip if the last element in the set has less work (or equal work and a lower sequence id) than the tip, which also must not be in the set.

The set is always modified while holding `cs_main` (it should be annotated thus but isn't), but I haven't verified that `chainActive.Tip()` can't change while holding the lock. That's one possibility.

[...]

PR #962, opened 2021-07-26
[Chain] Prune entire orphaned chains.

Zannick:
### Problem

Pruned blocks sometimes cause other blocks in the index (mostly descendants of orphans) to have invalid parents, which may trip an assert `pindexWalk->pprev` (#692) on startup. Additionally, because block indexes *don't own their own hash values*, removing the index from `mapBlockIndex` where the hash is allocated may lead to a use-after-free corrupting the hash of a block written to disk (may be the cause of https://github.com/Veil-Project/veil/issues/930#issuecomment-819166828).

### Solution

Select pruning candidates in two rounds: first, based on height. Then, add all descendants of the blocks chosen in the first round.

For the memory issue, add a `shared_ptr` field to CBlockIndex that provides the hash value when the block is removed from mapBlockIndex. This will manage the memory of the hash for the lifetime of the object (which is forever anyway) and allows for efficient sharing and copying as usual (unique_ptr does not).

### Testing

Via regtest. See https://github.com/Veil-Project/veil/issues/692#issuecomment-886256882

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -109,8 +109,8 @@ const CBlockIndex* CBlockIndex::GetAncestor(int height) const
             heightWalk = heightSkip;
         } else {
             if (!pindexWalk->pprev) {
-                LogPrintf("%s: pindexWalk failed:  Hash: %s, Current Height: %d, Working Height: %d\n",
-                          __func__, pindexWalk->phashBlock->GetHex(), height, nHeight);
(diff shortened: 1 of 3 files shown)


codeofalltrades (approved):
ACK 20549463c13f1859ba2bdc16b2c292aa7e18309a

CaveSpectre11:
utACK   https://github.com/Veil-Project/veil/commit/20549463c13f1859ba2bdc16b2c292aa7e18309a

CaveSpectre11 (approved):

Issue #942, opened 2021-05-15
Orphans sometimes have mismatched hashes in the db


Zannick:
Blocks in the leveldb occasionally appear to have bad hashes. At startup many messages will appear in the log like this:
```
2021-05-14T03:40:58Z LoadBlockIndexGuts: Skipping Block 116: e4331ddbe300b6eb008d4f0313e155bd477038beb6f2787048d422659f2add85 (status=2)- Block Hash does not match Index Key
```
[slightly modified the checked-in code to show block height]

It's unclear whether this is related to #692 (which I did trigger on occasion) and this is quite complex so I am filing a separate bug.

With some additional debugging in `WriteBatchSync`, a search over the database produces some more information when blocks enter this state:
`2021-05-14T03:40:58Z (debug) WriteBatchSync: !!!! Search found bad block! 116 - Wrote: bf18646c65c8200e3c28a25fa7378070fbc70f95ccc646bbf8cf785f84112495, Read: e4331ddbe300b6eb008d4f0313e155bd477038beb6f2787048d422659f2add85`
Block indexes have a block hash cached in the index, and under certain circumstances, another hash of the index is created.

The cause of this difference is that orphaned blocks do not always have an updated `mapAccumulatorHashes` (all entries are set to the default 0)<sup>1</sup>, although they have a cached hash of what said map *should* be, in `hashAccumulators`. A hash can be retrieved in these ways:
*  `CBlockIndex::GetBlockHash` returns a cached hash.
    https://github.com/Veil-Project/veil/blob/b42559603a1cfd0c88dbe06d58ebc40e4c7822ce/src/chain.h#L394-L397
*  `CBlockIndex::GetBlockHeader` constructs a CBlockHeader object that will compute the hash. In doing this, it **hashes the current accumulator map**
   https://github.com/Veil-Project/veil/blob/b42559603a1cfd0c88dbe06d58ebc40e4c7822ce/src/chain.h#L386
*  `CDiskBlockIndex::GetBlockHash` constructs a header to compute the hash, and provides its **cached** version of `hashAccumulators`.
  https://github.com/Veil-Project/veil/blob/b42559603a1cfd0c88dbe06d58ebc40e4c7822ce/src/chain.h#L707

The problem here is inconsistency: `GetBlockHash` is not virtual, so its result depends on which class we consider the object to be while we read it--because one method recalculates `hashAccumulators` rather than use the saved value.

[...]

PR #963, opened 2021-07-26
[Chain] Keep hashAccumulators more up-to-date.

Zannick:
### Problem

Issue #942:
Orphaned blocks sometimes had inconsistent hashes between block index hash (generally the cached value) and the disk index hash (produced by taking a new hash of the data). On closer inspection, `hashAccumulators` was the only differing value.

### Solution

Make sure `hashAccumulators` is set or copied appropriately where relevant, particularly when `mapAccumulatorHashes` is set but not changed. The main culprit is likely `LoadBlockIndexGuts`, since it would be very easy to forget to update `hashAccumulators` before writing it out again.

Plus we use `hashAccumulators` instead of recalculating the map's hash as described in #942.

### Testing

On regtest, with lots of extra debug warnings in `WriteBatchSync` about bad hashes, minted zerocoins in forks, confirmed this makes the debug warnings go away.

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -254,14 +254,16 @@ const CBlockIndex* LastCommonAncestor(const CBlockIndex* pa, const CBlockIndex* 
 void CBlockIndex::AddAccumulator(libzerocoin::CoinDenomination denom, CBigNum bnAccumulator)
 {
     mapAccumulatorHashes[denom] = SerializeHash(bnAccumulator);
+    hashAccumulators = SerializeHash(mapAccumulatorHashes);
 }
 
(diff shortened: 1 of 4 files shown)


CaveSpectre11 (commented):
utACK https://github.com/Veil-Project/veil/pull/963/commits/a7941b80f9001e8e4f38966c3d8d428263279f21

CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/963/commits/a7941b80f9001e8e4f38966c3d8d428263279f21

codeofalltrades (approved):
ACK a7941b80f9001e8e4f38966c3d8d428263279f21

PR #966, opened 2021-09-19
[Mining] Clear the calculated hash speed at mining start.

Zannick:
This avoids having the hash speed affected by a gap in mining
or a change in the number of mining threads, without requiring
users to call `setminingalgo` on the same algorithm.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4989,6 +4989,7 @@ UniValue generatecontinuous(const JSONRPCRequest& request)
         result.pushKV("threads", 0);
         result.pushKV("message", "Mining stopped");
     } else {
+        ClearHashSpeed();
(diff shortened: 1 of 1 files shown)


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/966/commits/55c60f82403d9e8bcb191635a19cf66119505148

WetOne:
utACK 55c60f82403d9e8bcb191635a19cf66119505148

WetOne (approved):

seanPhill:
I've got this on startup (after a power outage) on #966 and #967. 

On the first one,
> dbus[1146]: Dynamic session lookup supported but failed: launchd did not provide a socket path, verify that org.freedesktop.dbus-session.plist is loaded!
It's a different dbus number on the second one.

Not sure if this happened when I started them up yesterday. (Possibly.)

PR #967, opened 2021-09-19
[Mining] Recent mining hash speed

Zannick:
### Issue

In `getmininginfo`, it isn't clear that it computes the `hashspeed` by counting all hashes made since the start of hashing and dividing by the elapsed time, which tends to smooth out fluctuations but also (prior to #966) can include periods of time when the miner is off or the number of threads different.

### Solution

This adds a field to `getmininginfo` called `hashspeed_recent` which is a sum of each thread's most recent round of hashing (where a round is either a fixed number of hashes or ends with a new block).

### Tested

On regtest chain with sha256d and randomx.

PR #879, opened 2020-12-06
[RPC][Mining] Add sanity checks to generatecontinuous

CaveSpectre11:
### Note
This is built on top of #878 and requires code added in #877 and #878 
### Problem
`generatecontinuous` can get users in trouble by not knowing what their settings will do to their resources

### Solution
Add checks in `generatecontinuous`:

- Warn if RandomX is set to less than 4 threads.
- Warn if Sha256D is more than one thread less than the number of cores
- Provide ability to override the warnings with a new optional parameter
- Disallow generatecontinuous true if already running
- Rework the result of the command to provide a JSON status

### Testing
 ```
$ veil-cli generatecontinuous true 1
{
  "success": true,
  "algorithm": "sha256d",
  "threads": 1,
  "message": "Mining started"
}
$ veil-cli generatecontinuous true 1
error code: -32603
error message:
Mining already active
$ veil-cli generatecontinuous false
{
  "success": true,
  "algorithm": "sha256d",
  "threads": 0,
  "message": "Mining stopped"
}
```
```
$ ./veil-cli setminingalgo randomx
{
  "success": true,
  "message": "Mining algorithm changed from sha256d to randomx"
}
$ veil-cli generatecontinuous true 1
error code: -8
error message:
Error: RandomX must be at least 4 threads
```
```
$ veil-cli setminingalgo sha256d
{
  "success": true,
  "message": "Mining algorithm changed from randomx to sha256d"
}
$ veil-cli generatecontinuous true 100
error code: -8
error message:
Error: Available cores: 4, limit sha256d to 3 threads
$ ./veil-cli generatecontinuous true 4 true
{
  "success": true,
  "algorithm": "sha256d",
  "threads": 4,
  "message": "Warning: Available cores: 4, limit sha256d to 3 threads
```

PR #969, opened 2021-10-02
[GUI|MINE] Mining Tab

WetOne:
**Problem**
Mining requires use of QT

**Root Cause**
No Mining interface in the QT implementation

**Solution**
Implement a Mining Tab to allow for mining to be activated and controlled through the GUI.. The mining tab will control mining activity initiated through the CLI.


src/qt/veil/miningwidget.h
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2019 The Veil developers
```
CaveSpectre11, line 1: NIT: Be sure to give appropriate copyright when adding files (e.g. 2021) or update to include 2021 if it's not done yet (there's a script we can run but might as well do it when we've edited a file)

src/qt/veil/miningwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    connect(model, SIGNAL(updateMiningFields()), this, SLOT(updateMiningFields()));
+}
+
+MiningWidget::~MiningWidget()
+{
+    delete ui;
+}
+
+void MiningWidget::onUpdateAlgorithm() {
```
CaveSpectre11, line 70: NIT.  I know it's a copied and edited file, but be sure to try and clean up a new file to a single style.
```
void module::method(params){
````
```
void module::method(params) {
````
```
void module::method(params)
{
````

File: src/qt/veil/miningwidget.cpp [hunk 3]
```diff
@@ -0,0 +1,147 @@
...
+    int nThreads = ui->numThreads->text().toInt();
+
+    if (!mineOn) {
+        if (nThreads > (nCores - 1)) {
+            openToastDialog("Too many threads", mainWindow->getGUI());
+            return;
+        }
+            
+        if ((nAlgo == MINE_SHA256D) && (nThreads > (nCores - 1)))
+            openToastDialog(QString::fromStdString("SHA256D limited to " + (nCores-1)), mainWindow->getGUI());
```
CaveSpectre11, line 112: Two comments here.  First:  I'm contemplating it, but for cli based mining, they -can- override this check; so I'm leaning towards a "do it anyway" type button in case they really want to do it.

Second:  The second "if" statement won't be hit.  If `(nThreads > (nCores - 1))` then it will pop "Too many threads" and won't ever get down to the check about the SHA256D limitation.  The only way it can get to the MIN_SHA256D check is if the second condition isn't met.    Quite simply, that first nThreads check shouldn't be there, since it is intended to restrict SHA256D (Per PR #879)

WetOne: We can implement an override in a later iteration.  This PR can lay down a baseline.  I'm inclined to leave the first check and remove the 2nd check (first check applies to ProgPow as well).  It is almost irrelevant as the GUI currently restricts the user.  I'm inclined to leave it for security check.  Thoughts?

src/qt/veil/miningwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+        if ((nAlgo == MINE_RANDOMX) && (nThreads < 4)) {
+            openToastDialog(QString::fromStdString("RandomX must be at least 4 threads"), mainWindow->getGUI());
```
Zannick, line 115: Is there some way to indicate that this message doesn't prevent mining but changes the threads and starts anyway? Actually, scratch that. I would prefer the GUI enforce that during selection instead of on enabling mining--it wouldn't have to be a pop-up there. And I see below you already have the range being set for randomx.

src/qt/bitcoingui.cpp
```diff
@@ -331,13 +331,22 @@ void BitcoinGUI::createActions()
     receiveCoinsAction->setShortcut(QKeySequence(Qt::ALT + Qt::Key_4));
     tabGroup->addAction(addressesAction);
 
+    QIcon iconMining(":/icons/ic-mined-white-png");
+    miningAction = new QAction(iconMining,tr("&Mining"), this);
+    miningAction->setStatusTip(tr("Mining"));
+    miningAction->setToolTip(miningAction->statusTip());
+    miningAction->setCheckable(true);
+    miningAction->setShortcut(QKeySequence(Qt::ALT + Qt::Key_6));
```
Zannick, line 339: If this is above the settings tab, it should have the lower shortcut.

src/qt/veil/miningwidget.cpp
Zannick: I'm fine with the first as a safety check (until the option to override is added), given that the GUI should prevent the casual user from hitting this.

src/qt/veil/miningwidget.cpp
WetOne: You pretty much hit it on the head.  The GUI is the first line of defense, performing a range check prior to enabling.  This check is for redundancy (in case someone finds a way to get a lower number).

src/qt/veil/miningwidget.cpp
WetOne: With a better understanding of the history of the checks I'm OK with removing the first one.  I'd like to do something where if the user exceeds the recommended number of threads (which would be Cores -1) a popup appears allowing the user to override.  I'm still inclined to have that be a separate PR.

I have the formatting changes ready but have not pushed them.  @CaveSpectre11, let me know if you want the thread limits changed in this PR or a different one.  After that I will either push the changes I have now or add the other changes and push it in one go.

src/qt/bitcoingui.cpp
WetOne: There are a couple errors on the key sequence.  Good catch.  I'm going to clean it up.  It will be the following:
Overview -> 1
Send -> 2
Receive-> 3
Mine -> 4
Address -> 5
Settings -> 6

This is the order of the icons top to bottom.

src/qt/veil/miningwidget.cpp
```diff
@@ -0,0 +1,149 @@
...
+void MiningWidget::onUpdateAlgorithm() {
+    if (mineOn) {
+        openToastDialog(QString::fromStdString("Mining must be stopped to change algorithm!"), mainWindow->getGUI());
+    } else {
+        currentMiningAlgo = ui->cmbAlgoSelect->currentIndex(); 
+        std::string algoStr = GetMiningType(currentMiningAlgo, false, false);
+        bool setAlgoResult = SetMiningAlgorithm(algoStr); 
+
+        if (setAlgoResult) {
+            openToastDialog(QString::fromStdString("Aglorithm Switch Success!"), mainWindow->getGUI());
```
Zannick, line 75: nit: typo

src/qt/veil/miningwidget.cpp
```diff
@@ -0,0 +1,149 @@
...
+    setMineActiveTxt(mineOn);
+}
+
+void MiningWidget::updateMiningFields() {
+    mineOn = GenerateActive();
+    currentMiningAlgo = GetMiningAlgorithm();
+
+    setMineActiveTxt(mineOn);
+
+    ui->lblHashRate->setText(QString("Mining at %1 H/s").arg(QString::number(GetHashSpeed())));
```
Zannick, line 127: In a future PR, we might want to change this to GetRecentHashSpeed() (#967)

Zannick (approved):
utACK 7703eb22ffbe3dd8c7f4535cf015175699acddb6

src/qt/veil/miningwidget.cpp
CaveSpectre11: Since this is on a message presented to the user, and not a rare message; I'm going to upgrade this above just a 'nit'.

Zannick (approved):
utACK 781c6a6501e6ea44a3db86a575e8d1d4a2705346

Issue #970, opened 2021-10-05
[Feature][Wallet] Allow sendXtoY to create multiple transactions


Zannick:
## Summary

Currently, attempts to send Veil with these commands can fail if it does not fit in a single transaction or block (and for RingCT, sometimes the user needs to run an additional command to unstick the funds). Instead, we plan to create multiple transactions for sending the requested funds. Spending zerocoin already does this. (We will not cover minting zerocoin here.)

## Outline

- [x] Create MultiTxReceipt wrapper class akin to ZerocoinSpendReceipt.
- [x] Create new function to pick inputs up to a certain amount.
- [ ] Add RPC option to use multi-tx. Automatically determine maximum number of inputs (or define a hardcoded fixed maximum) or accept the number as the option.
- [ ] Modify send to create multiple transactions, assign partial values to recipients, test accepting them to the mempool, and commit them.
    - [ ] ct/ringct outputs are split in two

## Optional additions

- [ ] Add config option to default to multi-tx.
- [ ] Can change from one tx can be used in another?
- [ ] If a transaction is still rejected for being too large, it could be automatically split further?

PR #972, opened 2021-10-19
[Consensus][RingCT] Correctly track and test RingCT spends.

Zannick:
## Problem

Creating multiple ringct spend transactions in a batch (as part of #970) does not work because of these two similar issues.

1.  When creating a transaction, the transaction record has dummy inputs which don't correspond to any actual output, so we don't actually track the pending spends.
2.  When verifying that we aren't double-spending any ringct output, we add every keyimage we see to the set to check against. But if we're testing that a transaction could be accepted to the mempool, then committing that transaction, the commit fails (until the next block).

## Solution

1.  Extract the spent coins' txo records to be marked as pending spend.
2.  Don't track the inputs permanently when testing that a transaction could be accepted to mempool. We still track within the transaction.

## Tested

On regtest with other changes for #970.


codeofalltrades (approved):
ACK ec15117aca1c478a88fae104846839de6acbd40f

PR #973, opened 2021-10-22
Lock cs_main before calling GetBlocksToMaturity

Zannick:
This is similar to another call, just somehow missed this one earlier.


CaveSpectre11 (approved):
utACK https://github.com/Veil-Project/veil/pull/973/commits/0987b3d0a7298d35592323c4163744d476451848

PR #978, opened 2021-12-02
[Refactor] Move parts of AddAnonInputs into helper functions.

Zannick:
Rearranges part of `AddAnonInputs_Inner` into separate helper functions so it's easier to understand and also to reuse the code elsewhere. I'm not sure I fully understood the code enough to properly name the helper functions, but here they are:

* ArrangeBlinds
  * Places the real inputs and creates dummy entries for the ring.
  * Optionally adds temporary dummy sigs.
* GetKeyImage
  * Attaches the key image for a real input.
  * I made this take one at a time for ease of calling from elsewhere.
* SetBlinds
  * Does some further stuff to the blinds.

### DIFF
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -3122,6 +3122,174 @@ bool AnonWallet::IsMyAnonInput(const CTxIn& txin, COutPoint& myOutpoint)
     return false;
 }
 
+bool AnonWallet::ArrangeBlinds(
+        std::vector<CTxIn>& vin,
+        std::vector<ec_point>& vInputBlinds,
+        std::vector<std::vector<std::vector<int64_t>>>& vMI,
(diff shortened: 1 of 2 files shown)


WetOne (approved):

WetOne:
utACK 42dfa73438bdfe535449e93fb6bcf93ba1d968d0

Zannick:
Fixed the following:
- SetBlinds was called with `nSigRingSize`, `nSigInputs` reversed, causing a segfault.
- SetBlinds was accepting some vector arguments by value; now all vectors are correctly passed by reference.
- I had moved some vectors into SetBlinds cause they appeared otherwise unused, but they were storing data for other vectors and so needed to still be scoped to a longer lifetime than the others.

codeofalltrades (approved):
ACK 91e7f972d29d3845787643fd462a94e57007f791

PR #979, opened 2021-12-02
[Trivial] Fix a size_t reference for clang 13

Zannick:
This was the only one necessary to build veild. I imagine any others have a `using` directive somewhere in their includes.


codeofalltrades (approved):
ACK 7e4c850c727b362035ba4af4f97c89fccb12a98b

PR #989, opened 2022-02-09
[Consensus] Update main chain POW end to supply end

Zannick:
Per the team, POW should be extended throughout the end of the supply.

As a change to the chainparams, this will be a mandatory change--versions without this update will reject POW blocks starting in November 2022, causing a fork. It may be prudent to add a protocol version bump as well to enforce separately.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -366,7 +366,7 @@ public:
         nKernelModulus = 100;
         nCoinbaseMaturity = 100;
         nProofOfFullNodeRounds = 4;
-        nLastPOWBlock = 2000000;
+        nLastPOWBlock = 9816000; // Continue POW until supply creation ends
         nHeightSupplyCreationStop = 9816000; //Should create very close to 300m coins at this time
         nTimeEnforceWeightReduction = 1548619029; //Stake weight must be reduced for higher denominations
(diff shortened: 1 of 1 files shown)


WetOne (approved):

seanPhill:
Is there any specific test I can do to confirm the extension of PoW to the end of supply?

Zannick:
The only thing I can think of would be to change this value for regtest to a much lower height and verify that PoW stops at that height.

seanPhill:
Aside from my adding the extension to all chains, I'm not really seeing where this value is compared in the code for the lower number to have had any effect whatsoever. Maybe it was never going to happen?

seanPhill:
> Aside from my adding the extension to all chains, I'm not really seeing where this value is compared in the code for the lower number to have had any effect whatsoever. Maybe it was never going to happen?

Ah, found it in `validation.cpp`!

After chainparams.h passes the value of nLastPOWBlock to LAST_POW_BLOCK it then uses that below.
```
    //Do not accept PoW blocks after the last pow height
    if (pindex->nHeight > chainparams.LAST_POW_BLOCK() && !block.IsProofOfStake())
        return state.DoS(100, error("%s: Proof of Work block added after the last allowed PoW height", __func__), REJECT_INVALID, "pow-after-cutoff");
```

PR #990, opened 2022-02-13
[RPC] Correct getblockstats fee calculation

Zannick:
Current implementation always performs a subtraction `ins - outs` to calculate fee, which is only correct for basecoin-to-basecoin, and results in zerocoin spends appearing as the fee. CT/RingCT and Zerocoin Minting/Spending include a ct_fee output that we can use (ZC spends have a ct_fee of 0 but at least that lets us skip the subtraction).


minerminer4949 (approved):
ACK 95c474077429da08cffd2e4e154fc4a822155a90

PR #991, opened 2022-03-13
[RPC][Wallet] Create multiple transactions with one command.

Zannick:
## Details

RPC send commands have a new argument `inputs_per_tx` that controls whether the amount specified can be split across multiple transactions. This value ranges from 0 (default) to 32 (an upper bound designed to prevent transactions larger than can be accepted). When set to 0, attempts to fit the send in one transaction as per previous behavior.

The sends are interpreted to mean that the recipient will receive the sent amount, but under the hood the transactions are created as if `subtractfeefromamount` was specified. In particular, if the send needs to be split into `n > 1` transactions, the outputs of the first `n - 1` are randomly reduced to account for the fee (and the shortfall from the inputs).

Coins are selected using a sliding window that prioritizes cleaning up small utxos (dust) over using fewer inputs, and thus transactions built this way will tend toward having more inputs. This is beneficial for ringct staking (keeping larger coins intact and reducing the amount of small utxos). The smallest utxo is always included if possible to prevent the amount of small utxos from increasing with usage.

## Tested
On regtest some months ago.

### DIFF
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -502,12 +502,12 @@ public:
                 break;
             }
             case OUTPUT_CT:
-                if (0 != pwalletAnon->AddBlindedInputs(wtx, rtx, recipients, !fCheckFeeOnly, nFeeRet, &coin_control, fail_reason))
+                if (0 != pwalletAnon->AddBlindedInputs(wtx, rtx, recipients, !fCheckFeeOnly, 0, nFeeRet, &coin_control, fail_reason))
                     fFailed = true;
                 break;
             case OUTPUT_RINGCT:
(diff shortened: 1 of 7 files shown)


Zannick:
`random_shuffle` is removed in C++17 so I'll look into replacing it with `std::shuffle` which is available but takes a slightly different 3rd argument.

Zannick:
Changed to use `Shuffle()` from random.h per other uses.

WetOne (approved):

WetOne:
utACK a872b41cc0a8fe2a09e48b2a5aaedd55ea048cc6

Veil-Project (approved):
ACK a872b41cc0a8fe2a09e48b2a5aaedd55ea048cc6

PR #992, opened 2022-03-19
[Logging] Add missing format specifier.

Zannick:
Discovered a malformed error message during debugging. This allows the error to properly be formatted.


codeofalltrades (approved):
utACK 3e948f272d6e562d02799cfba4e331b5f8a4d33e

PR #995, opened 2022-04-08
Proposed fix for (one type of) force_return errors.

Zannick:
## Issue

When dealing with some orphan transactions[^1], some addresses aren't in the address map, and so when we look them up we get an invalid return value to indicate we didn't find it. However, we don't check for the sentinel value and instead try to look up its type (it's a `boost::variant`) which can assert with forced_return if the `which_` field is outside the valid range.

[^1]: This only happened for me for months-old orphan blocks, and I can't tell if the address never existed or if it got corrupted at some point.

## Fix

We don't need to look up an address in the map when all we're going to do is use the key (which must be the same type per `boost::variant`).

## Tested

Windows 10 mainnet wallet.


codeofalltrades (approved):
ACK 61c6f10e65aac25ff9029a04a19629a4a1a126c0

PR #996, opened 2022-04-12
[QT] Proposed fix for (one type of) force_return errors

codeofalltrades:
## Issue

When dealing with some orphan transactions, some addresses aren't in the address map, and so when we look them up we get an invalid return value to indicate we didn't find it. 

## Fix

Verify the address is valid before trying to get the boost value.

## Tested

Windows 10 testnet wallet.


WetOne (approved):

WetOne:
utACK 1c07fe1f3a8fdede697edc16fca74d620f0ff53e

Issue #668, opened 2019-09-16
Stealth CT amounts received as 0.00000001 one satoshi of Veil per stealth input


seanPhill:

CryptoMaximalist:
This is the behavior of a locked wallet receiving CT funds.  The range proof which protects the privacy of the Tx amounts means a locked wallet can't access the private key to determine the amount, but it knows each output must be worth at least 1 satoshi.  CT and RingCT will always have at least 2 outputs because having only one would betray the output amount based on the input amounts

seanPhill:
That would have been nice to have disclosed in the (GUI and CLI) wallet. :o
How about then there be added a button or function that can unlock and rescan that one block for the transaction? -- Oh! Or how about Unlocked for Staking only ALSO add for scanning received transactions?  Going full RingCT is going to make this very important.
A single click solution would take barely a second or three.

Did I mention above that users reporting this problem stopped at least one exchange offering withdrawals to stealth addresses?

seanPhill:
I just did a little silent video showing me fix this problem as it happened. Bearing in mind that you can't just preemptively unlock your wallet before money arrives from a third party wallet.

https://vimeo.com/380165170

PR #1001, opened 2022-04-24
Fix wrong CT / RingCT amounts in transaction history when wallet locked

### LINKED ISSUE #922: Make automated rescanringctwallet and friendly message enabling exchanges to send to stealth without worrying recipients
Exchanges are only sending to basecoin. Users are worried when receiving stealth txs.

 
  Noobs need the process of confidently receiving stealth funds to be noob-friendly, even grandma, executive, or trainee / intern friendly.

  The Vinex exchange very early in Veil history tried sending withdrawals to SV addresses, but customers thought they only received a single Satoshi of Veil as their wallet was locked at the time. This lead Vinex to suspend withdrawals to stealth addresses due to support workload. 

 If the wallet is locked when processing the block containing the transaction, it will (in the recipient's wallet) have a placeholder value of one satoshi only. After unlocking it is necessary to `rescanringctwallet`. This needs to be automated if a received transaction has been detected, and a friendly message needs to be generated for the GUI QtWallet user and for RPC-using  exchanges alike. 

My write-up in Discord:

> The only exchange that did implement stealth withdrawals was Vinex, early on, but when people began receiving stealth transactions while their wallets were locked and therefore it just looked like they received a (placeholder amount of) 1 satoshi of Veil, support requests increased. It turns out that if your wallet is locked when you receive a stealth payment you have to unlock your wallet and do rescanringctwallet in the Debug Console (or the equivalent command in Terminal).
> 
> In the future, so that exchanges can send and receive stealth transactions I would like to see the wallet give a message like "You have received a payment. Please unlock your wallet so the amount can be decrypted." Then it should automatically rescanringctwallet, which only takes a few seconds in most cases. [There would be no need to add "stealth" to describe "a payment", because Veil transactions are normally stealth and eliminating basecoin transactions as a norm, such as from mining pools or to and from exchanges, will achieve this overwhelmingly.]

Much of this, I believe is already planned for the RingCT release, but I am posting this on the suggestion of a user, in relation at least to the friendly notification mentioned in the second paragraph. I don't see any specific related issues although I believe plans and intentions have been previously discussed. Please add any references if found.

us77ipis:
### Problem
When receiving a CT transaction while the wallet is locked (or unlocked for staking only), the amount displayed in the transaction history is wrong, and keeps wrong until a `rescanringctwallet` rpc is called and the wallet is restarted.

### Root Cause
- More concretely the wrong amount showed is -1 satoshi per output in the transaction, which comes from the default value in the following line: https://github.com/Veil-Project/veil/blob/7c8bee132ea6305b29a613679b7591a27d8edbd8/src/veil/ringct/outputrecord.h#L43
- Furthermore, for the case when the transaction contained a change output, there was a bug in the way the GUI calculates the transaction amount, which resulted in that default -1 satoshi also being added for the change output (which isn't owned by the receiving wallet), thus the final amount was in this case always displayed as 1 satoshi less than the actual transaction was.
- For incoming RingCT transactions the `ORF_OWNED` flag was not set correctly, leading to RingCT transactions being showed as "(n/a)" in the transaction history, if they arrived while the wallet was locked.

### Solution
- When a CT transaction arrives while the wallet is locked (or unlocked for staking only):
  - Instead of a wrong amount show **"Hidden"** in the transaction history entry.
  - In the tooltip of the transaction row show the message **"Unlock wallet to see hidden amount"**.
  - In the "Incoming transaction" GUI notification also show "Hidden" instead of a wrong amount.
- Fix `rescanringctwallet` to properly update corresponding entries in the transaction history GUI.
- Keep track of locked transactions and when the wallet is then unlocked, automatically do a **partial `rescanringctwallet`** which only rescans those tracked locked transactions, and thus should be fast also for wallets having a very long transaction history.

### Bounty Payment Address
`sv1qqpjsrc60t60jhaywj5krmwla52ska70twc7wun6qnee65guxhvtxegpqwhuxypra4jn3pq86s24ryltcw6g2ss4573hyqac9u4g23m9mvxpyqqqwny49k`

### Testing Results
- Tested by sending basecoin/CT/RingCT from one testnet wallet to CT/RingCT of another testnet wallet, where the received amount showed up first as "Hidden" and once the wallet was unlocked, it showed the correct amount.
- Tested unlocking of the wallet while the transactions are still unconfirmed.
- Tested restarting the wallet between transaction arrival and unlocking of the wallet (state consistency across wallet restarts).

Fixes #922, #668

### DIFF
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -106,7 +106,7 @@ public:
         int upperIndex = (upper - cachedWallet.begin());
         bool inModel = (lower != upper);
 
-        if(status == CT_UPDATED)
+        if(status == CT_UPDATED || status == CT_UPDATED_FULL)
         {
             if(showTransaction && !inModel)
                 status = CT_NEW; /* Not in model, but want to show, treat as new */
@@ -120,6 +120,23 @@ public:
 
         switch(status)
         {
+        case CT_UPDATED_FULL:
+        case CT_DELETED:
+            if(!inModel)
+            {
+                const char *statusStr = status == CT_DELETED ? "CT_DELETED" : "CT_UPDATED_FULL";
+                qWarning() << "TransactionTablePriv::updateWallet: Warning: Got " << statusStr << ", but transaction is not in model";
+                break;
+            }
+            // Removed -- remove entire transaction from table
+            parent->beginRemoveRows(QModelIndex(), lowerIndex, upperIndex-1);
+            cachedWallet.erase(lower, upper);
+            parent->endRemoveRows();
+            if (status == CT_DELETED)
+                break;
+            // Fall through is intentional here to prevent code duplication, since CT_UPDATED_FULL
+            // is equivalent to first CT_DELETED and then CT_NEW.
+            inModel = false;
         case CT_NEW:
             if(inModel)
             {
@@ -150,17 +167,6 @@ public:
                     parent->endInsertRows();
                 }
             }
-            break;
-        case CT_DELETED:
-            if(!inModel)
-            {
-                qWarning() << "TransactionTablePriv::updateWallet: Warning: Got CT_DELETED, but transaction is not in model";
-                break;
-            }
-            // Removed -- remove entire transaction from table
-            parent->beginRemoveRows(QModelIndex(), lowerIndex, upperIndex-1);
-            cachedWallet.erase(lower, upper);
-            parent->endRemoveRows();
             break;
         case CT_UPDATED:
             // Miscellaneous updates -- nothing to do, status update will take care of this, and is only computed for
@@ -553,7 +559,12 @@ QVariant TransactionTableModel::txWatchonlyDecoration(const TransactionRecord *w
 
 QString TransactionTableModel::formatTooltip(const TransactionRecord *rec) const
 {
-    QString tooltip = formatTxStatus(rec) + QString("\n") + formatTxType(rec);
+    QString tooltip;
+    if (rec->status.isLocked) {
+        tooltip += tr("Unlock wallet to see hidden amount");
+        tooltip += "\n";
+    }
+    tooltip += formatTxStatus(rec) + QString("\n") + formatTxType(rec);
     if(rec->type==TransactionRecord::RecvFromOther || rec->type==TransactionRecord::SendToOther ||
        rec->type==TransactionRecord::SendToAddress || rec->type==TransactionRecord::RecvWithAddress)
     {
@@ -657,6 +668,8 @@ QVariant TransactionTableModel::data(const QModelIndex &index, int role) const
         return walletModel->getAddressTableModel()->labelForAddress(QString::fromStdString(rec->address));
     case AmountRole:
         return qint64(rec->credit + rec->debit);
+    case AmountLockedRole:
+        return rec->status.isLocked;
     case FeeRole:
         return "Fee " + ( rec->getFee() == 0 ? "0.00 VEIL" : BitcoinUnits::formatWithUnit(BitcoinUnits::VEIL, rec->getFee(), false, BitcoinUnits::separatorAlways)) ;
     case TxHashRole:
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -5021,7 +5021,8 @@ void AnonWallet::RescanWallet()
         CTransactionRecord* txrecord = &mi->second;
 
         int nHeightTx = 0;
-        if (!IsTransactionInChain(txid, nHeightTx, Params().GetConsensus())) {
+        bool isTransactionInChain = IsTransactionInChain(txid, nHeightTx, Params().GetConsensus());
+        if (!isTransactionInChain) {
             //This particular transaction never made it into the chain. If it is a certain amount of time old, delete it.
             if (GetTime() - txrecord->nTimeReceived > 60*20) {
                 //20 minutes too old?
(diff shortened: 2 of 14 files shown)


us77ipis:
The current changes of the PR already work and solve the problem. However, it is still marked as draft because currently when the wallet is unlocked it does the full `rescanringctwallet` in the UI thread, which I don't think is a good idea. We could do one of the following things (or both?) to optimize this:
- Do `rescanringctwallet` in another thread, when the wallet is unlocked.
- Don't do a full `rescanringctwallet`, but instead keep track of "locked" transactions and then only specifically rescan those when the wallet is unlocked.

us77ipis:
Btw. what will the bounty be for solving this soon three year old issue with as far as I can read in #922 quite high impact for Veil users :wink: ?

seanPhill:
I've got this PR running on Testnet, and firstly ran the released standard released version, to compare, but this time, unlocking has so far taken 2.5 hours, and is continuing to show the macOS beach ball of waiting, and no progress in the debug log. :( Admittedly, this wallet is normally slow (because the wallet.dat file is moderately large, and it has done a huge number of stakes and spends), but not like this. I have been able to rescanringctwallet with no problem, but with PR1001 it's just not completing. This is programmed to only do a single rescanringctwallet upon unlock, right?
There was probably at least thirty or forty ringct incoming transactions that had not got a value except for "0.00000001" (as displayed on the released version), and "0.00000000" when opened in PR1001.
Screen Shot 2022-04-29 at 9 37 57 am
The beachball of waiting is invisible in the screenshot, but it is still there.

us77ipis:
That sounds like a locking issue. I think I forgot to lock also `cs_main`. @seanPhill Please try again with the changes I just pushed.

us77ipis:
> There was probably at least thirty or forty CT (from incoming zerocoin spends) incoming transactions that had not got a value except for "0.00000001" (as displayed on the released version), and "0.00000000" when opened in PR1001.

That's the expected behavior for transactions that had already arrived before. For the "Hidden" text to be displayed the transaction has to have income through the wallet that has the changes of this PR. However, once you unlock the wallet (and it doesn't hang up) also the old "0.00000000" values should be replaced by the correct ones.

src/qt/transactiontablemodel.cpp
```diff
@@ -120,6 +120,20 @@ class TransactionTablePriv
 
         switch(status)
         {
+        case CT_UPDATED_FULL:
+        case CT_DELETED:
+            if(!inModel)
+            {
+                qWarning() << "TransactionTablePriv::updateWallet: Warning: Got CT_DELETED, but transaction is not in model";
```
Zannick, line 127: This warning should say CT_UPDATED_FULL instead if that's the status.

src/qt/transactiontablemodel.cpp
```diff
@@ -120,6 +120,20 @@ class TransactionTablePriv
...
+                break;
+            }
+            // Removed -- remove entire transaction from table
+            parent->beginRemoveRows(QModelIndex(), lowerIndex, upperIndex-1);
+            cachedWallet.erase(lower, upper);
+            parent->endRemoveRows();
+            if (status == CT_DELETED)
+                break;
+            inModel = false;
```
Zannick, line 136: Please do not implicitly fall through. I recommend inverting the if statement if you meant to break here after inModel = false.

src/veil/ringct/anonwallet.cpp
```diff
@@ -4025,13 +4025,18 @@ bool AnonWallet::UnlockWallet(const CExtKey& keyMasterIn)
...
     AnonWalletDB wdb(*walletDatabase);
     if (!wdb.ReadNamedExtKeyId("stealthchange", idChangeAccount)) {
         //If the change account is not created yet, then create it
         if (!CreateStealthChangeAccount(&wdb))
             return error("%s: failed to create stealth change account ", __func__);
     }
+    }
+
+    LOCK2(cs_main, pwalletParent->cs_wallet);
+    RescanWallet();
```
Zannick, line 4039: This may be a good opportunity to mark RescanWallet with `EXCLUSIVE_LOCKS_REQUIRED` annotations if it always needs these locks, that way clang can catch issues like that in the future.

Zannick (changes requested):

src/qt/transactiontablemodel.cpp
us77ipis: Fall through is intentional here to prevent code duplication, since `CT_UPDATED_FULL` is equivalent to first `CT_DELETED` and then `CT_NEW`.

Zannick: Ok, then please add a comment explaining the fallthrough is intentional.

us77ipis: Okay, I'll do that

seanPhill:
Tested a second time with the changes at the point in time of my thumbs up above.
Fully worked as expected. This wallet took just a few minutes, but a smaller wallet would have been quicker.
Screen Shot 2022-04-29 at 4 01 54 pm
Screen Shot 2022-04-29 at 4 03 04 pm
Screen Shot 2022-04-29 at 4 04 07 pm
Screen Shot 2022-04-29 at 4 04 29 pm
Screen Shot 2022-04-29 at 4 04 55 pm
Screen Shot 2022-04-29 at 4 05 35 pm

us77ipis:
This PR is now ready for extensive testing and review.

us77ipis:
I don't know how the translation stuff currently works. Since I added the new strings "Hidden" and "Unlock wallet to see hidden amount", should some translations be added? If yes, how exactly is that handled?

seanPhill:
I just had an issue with this function tonight. [Don't know if this is related to the requested changes.] 

While it has normally worked fine for me, my testnet wallet had been locked (and / or closed) and receiving a lot of transactions for about twenty hours, and maybe I unlocked it a bit soon after starting it up, but when I did, the transactions this time did not correct themselves. I was still getting the mouseover message about needing to unlock to see the correct amounts. I waited a few minutes, and tried locking the wallet again and unlocking it again, and it still did not correct the transactions list. ... I then manually did `rescanringctwallet`, and it promptly fixed all the amounts.

This is a wallet with over 55,000 transaction rows in the course of twenty days from fresh seed. The issue was with the 62 received CT amounts over the course of 20 hours, while the wallet was either locked or turned off.

us77ipis:
> [Don't know if this is related to the requested changes.]

I already pushed the requested changes (and marked them as solved), don't know why it still shows "Changes requested".

> I then manually did rescanringctwallet, and it promptly fixed all the amounts.

If full `rescanringctwallet` worked, then that is an issue with the optimization, thus for whatever reason those transactions didn't get tracked as locked transactions.

seanPhill:
@Zannick can you confirm that your requested changes have been made? ... -> code review. Everything should be ready.

Zannick (approved):
utACK 6333d875984d0cecaceefa6d2f7a830755274a44

Confirmed that the changes I requested were addressed. ("Changes requested" always shows if someone's last review was "Request changes", sorry it took so long to update the review.)

WetOne:
utACK 6333d875984d0cecaceefa6d2f7a830755274a44

WetOne (approved):

seanPhill:
This PR solves the issue that caused an exchange in early 2019 to reverse their decision to allow stealth withdrawals.
I hope that this PR can be included in the next release, which will enable exchanges to enable stealth withdrawals. @codeofalltrades

Issue #1015, opened 2022-07-03
RingCT Staking 


Zannick:
This is a tracking issue for the RingCT Staking project. I'm thinking of it in terms of the following breakdown, with parts of it sharing some similarities to either or both of Zerocoin staking and regular RingCT transactions:

## Staking Weight

Zerocoin staking power is based on the denomination "value" of the zerocoin, with a small reductions in power at each level above the first: a 100-coin has a weight of 90, a 1000-coin has a weight of 800, and a 10000-coin has a weight of 7000.

RingCT UTXOs have unknown values; to properly apply weights with minimal loss of privacy we have to make brackets similar to Zerocoin denominations. Previous work on this settled on using powers of 16 (in sats). I'll add some reductions in power to mimic the Zerocoin weights; we want to ensure that having 16 in a particular bracket will outweigh 1 in the next bracket (not counting dust).

bracket | min | weight factor | weight
--|--|----|--
1|0.00000017
2|0.00000257
3|0.00004097
4|0.00065537
5|0.01048577
6|0.16777217
7|2.68435457|100%
8|42.94967297|95%|40.8021893215
9|687.19476737|91%|625.3472383067
10|10,995.11627777|71%|7,806.5325572167
11|175,921.86044417|50%|87,960.930222085
12|2,814,749.76710657|30%|844,424.930131971
13|45,035,996.27370497|10%|4,503,599.627370497

The first three scale factors were chosen to closely match the equivalent values in zerocoin, the latter three were arbitrary extensions.

- [x] Staking brackets
- [ ] Staking weight reduction

## Transaction Creation

Blocks with Proof of Stake have two transactions: the coinbase transaction (generic), and the stake transaction. For zerocoin, the stake transaction takes one zerocoin as input, and mints one zerocoin of the same value plus the value of the coinbase split into 10s.

RingCT anon inputs are hidden amongst other anon UTXOs. RingCT sends generally output two UTXOs with a random split of the value, plus a RangeProof. However, change returned to oneself is usually a single output. In all cases, the value is hidden.

**Design decision**: RingCT staking transactions will return 3 outputs: one will be the same value as the original input, the other two will split the reward.

Much of my work so far has been refactoring the staking and RingCT code to allow for a smoother implementation reusing these components.

- [x] RingCT stake transaction input handling
- [ ] RingCT stake transaction outputs handling (in progress)
- [ ] RingCT stake transaction rangeproof(s)
[...]

PR #1019, opened 2022-10-30
Implement RingCT Staking

Zannick:
## Summary

This PR implements the functionality necessary for the creation and validation of blocks via Proof of Stake using unspent RingCT outputs, as described in #1015 and taking much inspiration (and code) from #678.

This does not yet enable RingCT staking (by placing the default start at block 50 million).

## Details

Potential RingCT inputs are divided into **brackets**; the stake weight of a coin corresponds to the minimum value of its bracket, with some higher levels further reduced, similar to zerocoin (see #1015 for details). A RingCT output becomes eligible for staking after 1000 blocks, same as zerocoin.

A RingCT stake block always has two key transactions, the **coinbase** and the **coinstake**. The coinbase contains either an empty standard output (most blocks), or the standard outputs from the superblock, plus two RingCT outputs which provably sum to the block stake reward. The coinstake contains a single RingCT input (the staking coin), and three outputs: an empty standard output indicating that it's a coinstake, a data output, and a RingCT output of the same value as the input. The rangeproof for this transaction proves that the input is large enough to pass the stake difficulty, by providing a min_value equal to the minimum value of a bracket. For additional privacy, this is not the bracket the coin is in but the lowest possible bracket that still passes the stake difficulty.

The block is signed using the same key as the staking coin.

## Other changes

* The wallet automatically creates a new stealth address for RingCT staking.
* Fixes an issue where RingCT marked the wrong thing as spent[^1], and also adds a way to reset the pending spend status for stakes that fail after the transaction is created.
* The base_uint class now has a `safeMultiply` function that informs the caller whether the operation overflowed the datatype.
* Fixes some small oversights like GetRangeProofInfo returning an int 0 or 1 when it could not be confusing and just return a bool, some error() calls with too many arguments, and a lack of default values for rangeproof parameter overrides.
* A lot of refactoring, many for clarity.

[^1]: Possibly I should split this out for a separate review, though I'm not sure what it affects.

## Potential improvements (for the project as a whole)

Cache the pubkey used to mark the RingCT stake uniqueness (#678 does this), since it may be repetitive across stake attempts to otherwise access it.

Remove the brackets and use a flatter scale with some high cutoff like 1M, given that we can reduce the rangeproof min_value anyway.

Add separate config settings for staking via Zerocoin or RingCT.

Add UI elements for the GUI wallet.

Add new tests/fix existing tests. :)

## What next?

This PR needs a thorough review, including a good security review that it won't be possible to generate free coins, and testing in dev. We can discuss further features as well. Finally, we have to pick a height at which to enable on testnet (causing a fork).


src/validation.cpp
```diff
@@ -4903,33 +4933,78 @@ static CDiskBlockPos SaveBlockToDisk(const CBlock& block, int nHeight, const CCh
...
-    } else {
-        return error("%s: dynamic_cast of stake ptr failed", __func__);
+    switch (stake->GetType()) {
+        case STAKE_ZEROCOIN:
+            ZerocoinStake* stakeZerocoin;
+            try {
+                stakeZerocoin = dynamic_cast<ZerocoinStake*>(stake);
+            } catch (std::bad_cast) {
+                return false;
+            }
```
us77ipis, line 4945: `dynamic_cast` on a pointer does not throw `std::bad_cast` on failure but returns a null pointer instead.

src/validation.cpp
```diff
@@ -4903,33 +4933,78 @@ static CDiskBlockPos SaveBlockToDisk(const CBlock& block, int nHeight, const CCh
...
+
+            return ContextualCheckZerocoinStake(pindex, stakeZerocoin);
+        case STAKE_RINGCT:
+            PublicRingCTStake* stakeRCT = nullptr;
+
+            try {
+                stakeRCT = dynamic_cast<PublicRingCTStake*>(stake);
+            } catch (std::bad_cast) {
+                return false;
+            }
```
us77ipis, line 4955: `dynamic_cast` on a pointer does not throw `std::bad_cast` on failure but returns a null pointer instead.

us77ipis (changes requested):
Just quickly skimmed through the code and found these...

src/validation.cpp
Zannick: Done, thanks.

Zannick:
devnet will start ringct staking alongside zerocoin staking at height 25000.

Zannick:
DevNet mining is underway. Sean notes: 
> In the GUI the RingCT stakes are incorrectly labeled as "Mined on 3txz...longStealthAddress". It's easy to identify them with listtransactions.

We should update those transactions to be RingCT stakes. The transactions themselves will look similar to mining transactions; we'll have to look at the stake transaction next to it.

seanPhill:
> DevNet mining is underway. Sean notes:
> 
> > In the GUI the RingCT stakes are incorrectly labeled as "Mined on 3txz...longStealthAddress". It's easy to identify them with listtransactions.
> 
> We should update those transactions to be RingCT stakes. The transactions themselves will look similar to mining transactions; we'll have to look at the stake transaction next to it.

A second case in the display of (RingCT stake winning) transactions in the GUI is that (maybe if the winning RingCT utxo is seen by the wallet first) it displays as "Zerocoin Stake", but shows the value of the RingCT utxo. Users would be confused as their wallet will say they _gained_ 18001.99083067 Veil, (magically) for example, but their balance will of course have only gained the amount of the block reward (50 Veil on devnet, and always ten Veil on mainnet).

seanPhill:
While syncing or resyncing we might need to look into _this_ error too. (Much repeated!)
> ERROR: ProcessNewBlock: AcceptBlock FAILED (bad-block-sig, PoS block signature not valid (code 16))(-1)

seanPhill:
> While syncing or resyncing we might need to look into _this_ error too. (Much repeated!)
... 
> Currently I have TWO happily synced and staking devnet wallets [only].

WHEN a wallet resynced from snapshot or otherwise has several blocks forked it is failing to stop requesting the same block from the connect=192.168.0... peer with ` ERROR: ProcessNewBlock: AcceptBlock FAILED (bad-block-sig, PoS block signature not valid (code 16))(-1)` even though I've already tried invalidateblock (a slightly earlier bad block) as the replacement for that block is rejected, while the perfectly happy forced-connected node that is synced and staking along with the other peer, is accumulating MILLIONS of banscore, as the connection is force, so it won't disconnect.

BUT, the cause of the third wallet forking immediately after starting up after loading the snapshot, without syncing from the running wallets, is that it was not a locked wallet, and it didn't wait to sync after starting!! ... Now this _third_ wallet is locked and HAS synced and has no gazillion of errors accumulating in the debug log.

PR #1020, opened 2022-11-01
[Mining][RPC] Fix `getnetworkhashps` rpc

### LINKED ISSUE #1018: [Bug][RPC][Mining] The network hashrate returned by `getmininginfo` rpc is wrong
The problem is that the function that should return the network hashrate for one of the three mining algos, calculates the average hashrate using the `nChainWork` _"Total amount of work (expected number of hashes) in the chain up to and including this block"_ which **includes the work of all three algos** and if I see correctly even also PoS work.
https://github.com/Veil-Project/veil/blob/4107c99a2029a07f89ad0b1e0f68c03ec55b7350/src/rpc/mining.cpp#L159

This is also the reason why all the returned network hashrates for the three different algos are always almost the same even though their difficulties vary a lot.

us77ipis:
The hashrates returned by the `getnetworkhashps` rpc were always wrong, because the calculation was based on the entire work of the chain (including all three mining algos and also proof-of-stake).

Fixes #1018

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -54,12 +54,35 @@ unsigned int ParseConfirmTarget(const UniValue& value)
     return (unsigned int)target;
 }
 
+enum class MiningType {
+    Sha256D,
+    ProgPow,
+    RandomX,
+    PoS,
+    X16RT,
+    Invalid
+};
+
+MiningType ParseMiningType(const std::string &type) {
+    if (type == "sha256d")
+        return MiningType::Sha256D;
+    if (type == "progpow")
+        return MiningType::ProgPow;
+    if (type == "randomx")
+        return MiningType::RandomX;
+    if (type == "pos")
+        return MiningType::PoS;
+    if (type == "xr16t")
+        return MiningType::X16RT;
+    return MiningType::Invalid;
+}
+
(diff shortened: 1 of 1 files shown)


src/rpc/mining.cpp
```diff
@@ -70,75 +70,97 @@ static UniValue GetNetworkHashPS(int lookup, int height, std::string type) {
     if (pb == nullptr || !pb->nHeight)
         return 0;
 
+    bool isPoS = type == "pos";
+    bool isSha256D = type == "sha256d";
+    bool isRandomX = type == "randomx";
+    bool isProgPow = type == "progpow";
+    bool isX16RT = type == "xr16t";
```
Zannick, line 77: Nit: this could probably be turned into an enum value and then used as a switch in the if-else blocks below.

Zannick (approved):

src/rpc/mining.cpp
us77ipis: True, that's the cleaner way, just changed that.

Zannick (approved):
utACK 15eeef7de68c101715a09a50cc8c85c235b330d5

seanPhill:
I've been running this for some weeks.

PR #1021, opened 2022-11-04
Bugfix: race condition of two threads using same RandomX VM

us77ipis:
Race condition between https://github.com/Veil-Project/veil/blob/4107c99a2029a07f89ad0b1e0f68c03ec55b7350/src/pow.cpp#L462 and https://github.com/Veil-Project/veil/blob/4107c99a2029a07f89ad0b1e0f68c03ec55b7350/src/chain.cpp#L304
which use the same RandomX VM but the first one correctly acquires the lock `cs_randomx_validator` before, while the second doesn't acquire the lock.

This bug caused the daemon / wallet to randomly crash.

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -294,7 +294,9 @@ uint256 GetRandomXBlockHash(const int32_t& height, const uint256& hash_blob ) {
     // Get the keyblock for the height
     auto temp_keyblock = GetKeyBlock(height);
 
-    // The hash we are calculating is in the same realm at the current validation caches
+    {
+    LOCK(cs_randomx_validator);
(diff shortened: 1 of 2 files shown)


Zannick (approved):

Zannick (approved):
ACK 140b2fe2bae389fae1c19e27c83f2a688c26f3e0

seanPhill:
I have been running this for some weeks, but haven't done a lot of RandomX mining.
I am stepping up this mining on testnet.

seanPhill:
I ran this on testnet and mined a lot of RandomX without any problems. Debug log entries look fine.

PR #1022, opened 2022-11-04
[RPC] Return RandomX seed and ProgPow epoch also if no miningaddress is set

us77ipis:
The `getblocktemplate` rpc, should always include the RandomX seed (if a RandomX template is requested) and the ProgPow epoch (if a ProgPow template is requested), also if no miningaddress was specified in the config.


Zannick (approved):

Zannick (approved):
utACK df354a8db31287886302494f10e4371b97d749fc

barrystyle:
Wouldn't this be a bit pointless if `pprpcheader` isn't returned?
This hash cannot be calculated without the miningaddress set, as it is required for the coinbase transaction to be formed in CreateNewBlock (in miner.cpp).

us77ipis:
For example a mining pool which builds the blocks from scratch needs the epoch information but not the `pprpcheader` nor does it need to set a miningaddress in the daemon because it is set externally while building the coinbase transaction like you said.

PR #1023, opened 2022-11-18
[Chain] Eliminate unnecessary overhead testing for stale block indexes.

barrystyle:
Stale block indexes will only occur if the chaintip gets thrashed around aggressively, or during extended periods of minting; never during InitialBlockDownload and even then this only needs to be checked periodically.

Performance increase during block synchronization is very noticeable once applied.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -2061,5 +2061,10 @@ bool AppInitMain()
         InitRandomXLightCache(chainActive.Height());
     }
 
+    // Check for stale block indexes every five minutes
+    scheduler.scheduleEvery([]{
+        PruneStaleBlockIndexes();
+    }, DEFAULT_INDEXCHECK_INTERVAL * 1000);
+
     return true;
 }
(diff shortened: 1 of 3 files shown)


seanPhill:
It does seem to be faster syncing (testnet), but I haven't tested and timed a full sync yet.

seanPhill:
I got about a block per 0.1 seconds syncing 4000 blocks on mainnet on a late-2013 iMac.

seanPhill:
Syncing 10001 mainnet blocks on a late-2013 Mac Pro, with `threadbatchverify=4` also yielded close to a tenth of a second per block, in about 20 minutes.

2022-12-20T03:15:16Z UpdateTip: new best=2b5b20c22072ad49d685332652207f4d84f50207999b2af4047eb5bf8557227a height=2043182 type=PoS version=0x30000000 tx=4293744 date='2022-12-13T03:14:36Z' cache=2.6MiB(15813txo)

2022-12-20T03:35:15Z UpdateTip: new best=40ea16b029bf674318948e7e18218a18a93fe2164582149fb17685c0cebd0a41 height=2053183 type=PoS version=0x30000000 tx=4311996 date='2022-12-20T03:35:25Z' cache=6.5MiB(39697txo)


10,001 blocks synced in 20 minutes (minus one second)

1199 seconds for 1000 blocks

1199/10001=0.11988801 seconds per block

PR #1026, opened 2023-04-09
Update build-factory.yml

ohcee:
Was having problems finding any action runners to start the workflow and came across this https://github.com/actions/runner-images/issues/6002. So I updated to the the oldest usable image 20.04. When finished I noticed that x86 Linux 32 bit failed. I assume there is no support after 18.04. Everything else is working fine.

### DIFF
--- a/.github/workflows/build-factory.yml
+++ b/.github/workflows/build-factory.yml
@@ -8,7 +8,7 @@ env:
 jobs:
   create-source-distribution:
     name: Create Source Distribution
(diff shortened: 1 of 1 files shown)


ohcee:
looks like the PPA doesn't have support for 20.04 `The repository 'http://ppa.launchpad.net/bitcoin/bitcoin/ubuntu focal Release' does not have a Release file.`

seanPhill:
This PR replaced by PR1027. Thanks.

PR #1028, opened 2023-04-19
Issue # 943 automintoff by default

ohcee:
This fix should have `automint` turned off by default. I have tested on mac with no issues. All the other wallet functionally seem to work with automint disabled. I have also changed the automint page to reflect the change and to help the user understand that they can still mint to the right of the page like always. As well as from the debug console.

### DIFF
--- a/src/qt/veil/forms/settingsminting.ui
+++ b/src/qt/veil/forms/settingsminting.ui
@@ -271,8 +271,8 @@ border:0;
 font-size:18px;</string>
               </property>
               <property name="text">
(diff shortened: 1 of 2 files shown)


seanPhill:
This works well! Tested with veil.conf as
automintoff=1 still not autominting
automintoff=0 it automints perfectly
# automintoff=0 commented out, therefore using default, ... not autominting. Success.

There are more enhancements I want, but they belong in a new issue.
Here's a preview of what I have in mind, but this PR is ready to commit.
Zerocoin Minting Page proposed enhancements

PR #1029, opened 2023-05-01
[Build] Allow building on later model Ubuntu/Debian

barrystyle:
Patch for libevent due to header differences in later glibc and patch for boost, to allow compilation with g++ 11


seanPhill (approved):
This code appears fine and worked without error.

PR #1035, opened 2023-06-02
[GUI] Add a link to the "Looks like you are new here!" screen by a "?" icon in the navigation bar

### LINKED ISSUE #1016: [Feature request] (low priority) Add a link to the "Looks like you are new here!" screen by a "?" icon in the navigation bar?
Sometimes people claim that this sort of day one help is not in the wallet, but the problem may simply be that the "Looks like you are new here!" screen disappears once some coins are received, and even though the FAQ page can be discovered via Settings, Options, Frequently Asked Questions, people forget it is there, and likely didn't read it in the first place, while the other information on the "Looks like you are new here!" page just doesn't exist anywhere else in the wallet.

So, maybe a nice, big question mark can be added below the gear icon in the nav bar?

Screen Shot 2022-07-04 at 7 10 32 pm

Necrosys:
Fixes #1016

**Screenshots**
* New wallet, Overview page, no transactions yet
new_wallet_overview_page_no_trx_yet
* New wallet, Overview page, immature transaction only
new_wallet_overview_page_immature_trx
* New wallet, Help page
new_wallet_help_page
* Old wallet, Overview page
old_wallet_overview_page
* Old wallet, Help page
old_wallet_help_page


ohcee (commented):
ACK working and looking as it should. Tested on MacOS with a balance/transactions and without balance.

seanPhill (approved):
Code looks good. More code changes than what I had envisioned, but function implementation is even more elegant than what I had in mind! :)

seanPhill:
@Necrosys please supply a Veil stealth address so we can tip you for this PR (and the previous one).

PR #1037, opened 2023-07-11
Local dependencies

steel97:
Because dependencies sometimes unavailable (see: https://github.com/Veil-Project/veil/actions/runs/5515437920/jobs/10055763646) I added local copy of most of dependencies.
I'am not sure if it's best solution, but it should work


Veil-Project:
I'd prefer to add this only to the 1.4.1 branch for the release, and leave the master branch to have all the .mk makefiles sources corrected.

PR #1051, opened 2025-12-06
Batch Zerocoin Database Write to Improve Sync Speed

ohcee:
This change improves sync performance by replacing multiple separate LevelDB writes for Zerocoin data with a single batched write. Previously, ConnectBlock() wrote spends, mints, and pubcoin spends individually, causing excessive disk I/O and slower block validation. The new function `WriteBlockZerocoinData()` combines these into one operation.

Key Points:
- Reduces LevelDB write overhead
- Enhances sync speed without configuration changes
- Improves performance on both older and modern hardware
- No consensus compatibility changes

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -823,6 +823,6 @@ bool CZerocoinDB::WriteBlockZerocoinData(
              (unsigned int)countMints,
              (unsigned int)countPubcoinSpends);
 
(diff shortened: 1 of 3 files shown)


seanPhill (approved):
ACK

seanPhill:
We'll run this on testnet for a while, but it should be good.

PR #1053, opened 2026-01-09
[Fix / Upgrade] PoS Staking tx inclusion, watchonly upgrades, new stealthkey rpc

blondfrogs:
This PR contains several fixes and improvements for watch-only address handling, stealth key management, and a critical PoS block creation bug.                                                 
                                                                                                                                                                                                  
  Bug Fixes                                                                                                                                                                                       
                                                                                                                                                                                                  
  Fix PoS blocks dropping first mempool transaction (7e90fcf)                                                                                                                                     
  - PoS blocks were silently dropping the first transaction from the mempool                                                                                                                      
  - The coinstake transaction was overwriting position 1 in the block's transaction list, where addPackageTxs() had already placed a mempool transaction                                          
  - Added a dummy coinstake placeholder to reserve position 1 before mempool transactions are added                                                                                               
  - This ensures all pending transactions are properly included in PoS blocks                                                                                                                     
                                                                                                                                                                                                  
  Fix watchonly status showing 0 transactions (1c4c6344)                                                                                                                                          
  - getwatchonlystatus was reporting 0 transactions even when transactions existed                                                                                                                
  - Legacy bug caused first transaction to be stored at index 0 with count also stored as 0                                                                                                       
  - Added check to detect transactions at index 0 when count is 0                                                                                                                                 
                                                                                                                                                                                                  
  Fix watchonly status and indexing (af9de02)                                                                                                                                                     
  - Fixed sync status check to compare against heightWhenImported instead of scanFromHeight                                                                                                       
  - Added progress information (scanned height, scan from height, imported height, percent complete)                                                                                              
  - Fixed getwatchonlytxes indexing to use proper 0-based user-facing API with 1-based database indexing                                                                                          
  - Added input validation for starting_index parameter                                                                                                                                           
                                                                                                                                                                                                  
  New Features                                                                                                                                                                                    
                                                                                                                                                                                                  
  Import stealth keys RPC and dump improvements (49affbf)                                                                                                                                         
  - Added importstealthkeys RPC to import stealth addresses with full private keys (scan + spend secrets)                                                                                         
  - Added stealth key support to dumpwallet RPC - now exports stealth address private keys                                                                                                        
  - Fixed metadata saving - ephemeral pubkeys are now always saved (needed to regenerate keys later)                                                                                              
  - Fixed several instances of RegenerateKey to use GetKey for proper key retrieval                                                                                                               
                                                                                                                                                                                                  
  Remove watchonly address improvements (c43a625)                                                                                                                                                 
  - Enhanced removewatchonlyaddress to accept stealth address string directly (in addition to scan_secret + spend_public)                                                                         
  - Now properly removes all associated transaction data from the database                                                                                                                        
  - Returns count of transactions removed                                                                                                                                                         
  - Added database method EraseWatchOnlyAddress for complete cleanup

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -167,6 +167,14 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
     pblocktemplate->vTxFees.push_back(-1); // updated at end
     pblocktemplate->vTxSigOpsCost.push_back(-1); // updated at end
 
+    // Add dummy coinstake tx as second transaction for PoS blocks
+    // This reserves position 1 so addPackageTxs doesn't put mempool txs there
+    if (fProofOfStake) {
+        pblock->vtx.emplace_back();
+        pblocktemplate->vTxFees.push_back(-1);
+        pblocktemplate->vTxSigOpsCost.push_back(-1);
+    }
+
     CMutableTransaction txCoinStake;
     CBlockIndex* pindexPrev;
     //Do not pass in the chain tip, because it can change. Instead pass the blockindex directly from mapblockindex, which is const.
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -227,6 +227,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "importlightwalletaddress", 2, "created_height"},
     { "importlightwalletaddress", 4, "num_prefix_bits"},
     { "importlightwalletaddress", 6, "bech32"},
+    { "importstealthkeys", 2, "rescan"},
     { "getanonoutputs", 0, "inputsize"},
     { "getanonoutputs", 1, "ringsize"},
     { "getkeyimages", 0, "txdata"},
@@ -237,9 +238,6 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "getnewaddress", 4, "makeV2"},
     { "getwatchonlytxes", 1, "starting_index"},
     { "getwatchonlytxes", 2, "batch_size"},
-
-
-
 };
 
 class CRPCConvertTable
--- a/src/veil/ringct/anonwallet.cpp
+++ b/src/veil/ringct/anonwallet.cpp
@@ -320,6 +320,7 @@ bool AnonWallet::GetKey(const CKeyID &address, CKey &keyOut) const
     if (mapKeyPaths.count(address)) {
         if (!RegenerateKey(address, keyOut))
             return error("%s: failed to regenerate keyid %s", __func__, address.GetHex());
+        return true;
     }
 
     if (pwalletParent->GetKey(address, keyOut))
@@ -350,8 +351,8 @@ bool AnonWallet::GetPubKey(const CKeyID &address, CPubKey& pkOut) const
     LOCK(pwalletParent->cs_wallet);
     if (mapKeyPaths.count(address)) {
         CKey keyTemp;
-        if (!RegenerateKey(address, keyTemp))
-            return error("%s: FIXME: regenerate key failed, but should not have!", __func__);
+        if (!GetKey(address, keyTemp))
+            return error("%s: failed to get key", __func__);
         pkOut = keyTemp.GetPubKey();
     }
 
@@ -391,7 +392,7 @@ bool AnonWallet::GetStealthAddressScanKey(CStealthAddress &sxAddr) const
 
 bool AnonWallet::GetStealthAddressSpendKey(CStealthAddress &sxAddr, CKey &key) const
 {
-    return RegenerateKey(sxAddr.GetSpendKeyID(), key);
+    return GetKey(sxAddr.GetSpendKeyID(), key);
 }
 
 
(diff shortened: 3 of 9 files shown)


ohcee:
utACK

seanPhill (commented):
I appreciate this bug fix (c43a625 "remove watchonly address fixes") and will be testing it on testnet pronto.
utACK

seanPhill (commented):
utACK af9de02 "watchonly status and indexing"

seanPhill (commented):
utACK ( 49affbf "importstealth rpc, stealthkey support in dump rpc, metadata saved")
Love your code quality and documentation!

seanPhill (commented):
utACK ( 7e90fcf "Fix PoS blocks dropping first mempool transaction" )
This is a great catch! I had not realised why I often had to push through transactions on testnet with a proof of work block. On mainnet it wasn't a huge problem, as there were always proof of work blocks. Testing this on testnet shortly.

In fact it wasn't extremely obvious as just looking at my spreadsheet of the superblock transactions, they've been confirmed in staked blocks slightly more than 50% of the time, and I haven't analysed this on other transactions such as budgets or my own many spends, statistically, so I hardly ever noticed it on mainnet, just when spending testnet funds, or noticed the wait _occasionally_ when broadcasting a budget spend, and didn't attribute it to anything like this.

seanPhill (commented):
utACK ( 1c4c634 "Fix watchonly status showing 0 transactions" )
💯

Veil-Project:
Very much appreciated. Will merge this as soon as testnet testing has been done.

ohcee:
Thanks for the awesome work @blondfrogs.

seanPhill:
I should remind people that Ubuntu 24.04 or later is needed for this. I went out of my way on a previous build to support an old Ubuntu, but I won't this time.

Those with old Ubuntu will get a message like this:
```
./veild: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./veild)
./veild: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./veild)
./veild: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./veild)
./veild: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by ./veild)
./veild: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by ./veild)
./veild: /lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.13' not found (required by ./veild)
```

It's best to just upgrade the host, from 20.04 LTS in the case I just tried.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.6 LTS
Release:	20.04
Codename:	focal
```

seanPhill:
@blondfrogs  On an Ubuntu server upgraded from 20.04 to 24.04 LTS I have a couple of testnet wallets, both with bloated  100-200 MB-ish wallet.dats. The one getting IPv4 peers is with the standard / automatic testnet settings, and the one set on IPv6 has a custom RPC port, username and password, and that all worked fine, and maybe this server is a bit undersized than some of my others, but they start and sync fine, but they are stopping after being unlocked, with the last line in the debug log being `UnlockZerocoinWallet: cwallet set zerocoin masterseed to …`
Any idea of what is going wrong?

```
$ ./veild -testnet
Segmentation fault (core dumped)
```

Restarting the released version 1.4.1.0.
Unlocked. Didn't crash. `getwalletinfo` … still running.

Second instance also running fine (on the released version), unlike the PR version.

seanPhill (commented):
Please see my notes above, based on testing.

ohcee (changes requested):
**Environment:** M4 Mac Mini, macOS, mainnet synced (~block 3,848,700+), QT console. All tests run live on mainnet.

---

### Verified

- `importstealthkeys` — verified, round-trip confirmed with `dumpprivkey`
- `dumpprivkey` stealth — verified, returns all four fields correctly
- `removewatchonlyaddress` by address string — verified
- `getwatchonlystatus` sync fix — verified
- `AddWatchOnlyAddress` scan start fix — verified
- `anonwallet.cpp` key lookup fixes — verified

---

### Issues

**Issue #1 — `transactions_removed` is never returned**

Commit `c43a625` explicitly claims "Returns count of transactions removed." It never appears in the response. `EraseWatchOnlyAddressData` reads `txCount` internally but the signature has no output parameter so the count is discarded before it reaches the RPC.

**Issue #2 — Erase loop misses transactions at index 0**

```cpp
for (int i = 1; i <= txCount; i++)
```

The first transaction for a fresh address is written at index 0 via the `current_count = -1` path in `WriteWatchOnlyTx`. The loop never erases it.

---

### Incorrect Description

Commit `49affbf` says "Added stealth key support to dumpwallet RPC." Only `dumpprivkey` was changed. Confirmed live `dumpwallet` output still contains the old note that stealth keys are not included.

---

### Not Tested

**PoS coinstake fix (`7e90fcf`)** I do not have mature zerocoin coins available for staking on this setup so this could not be verified end to end.

PR #1052, opened 2025-12-17
Optimize Zerocoin processing during block validation (batch writes + accumulator fast-path)

ohcee:
### Summary

This PR improves block validation and sync performance by reducing unnecessary Zerocoin database work during chain synchronization, while remaining fully consensus-safe and compatible with existing wallets.

The changes focus on:
- Writing all Zerocoin data for a block in a single batch
- Skipping empty Zerocoin database writes
- Avoiding repeated accumulator rebuilds after Light Zerocoin activation, except where required for correctness

***No protocol rules, data formats, or wallet behavior are changed.***

---

#### 1. Zerocoin DB batching
- Introduces `WriteBlockZerocoinData()` to write spends, mints, and pubcoin spends in a single LevelDB batch.
- Adds a fast-path that returns immediately when there is no Zerocoin data to write for a block.

#### 2. Accumulator validation optimization
- **Pre-Light Zerocoin**: retains existing behavior (validate every block).
- **Post-Light Zerocoin**:
  - Accumulators are only rebuilt on checksum boundary blocks.
  - Accumulator values are written only when needed for future spend verification.
- This preserves correctness while eliminating redundant work during sync.

#### 3. Safety and compatibility
- No consensus logic is altered.
- Zerocoin staking remains fully supported.
- RingCT behavior is unchanged.

---

### Performance Results

Observed improvements during sync testing:

- **macOS (Apple Silicon)**: consistent reduction in block connect time and higher blocks/hour over long runs
- **Older Ubuntu systems**: major reduction in Zerocoin database write time  
  - Zerocoin DB writes reduced from ~347 ms/blk to ~88 ms/blk
- Community side-by-side testing shows significantly faster historical catch-up compared to the current wallet

Exact gains vary by hardware, but improvements are consistent across systems.

---

### Testing

- Tested on macOS and Linux
- Tested from snapshots and during long continuous syncs

## Testers may optionally compare against the previous Zerocoin batching PR to observe the incremental improvement from the fast-path and accumulator changes. 

---

### Notes

This PR is designed to benefit all users automatically.  
No configuration flags or user action is required.

PR #1056, opened 2026-06-03
Zerocoin sync optimization

ohcee:
## Summary

Builds on the Zerocoin sync optimization from #1051 with additional performance improvements to accumulator validation and database writes, plus the critical correctness fix that caused the previous attempt (#1052) to be closed.

## What was wrong with #1052

The previous PR was closed because valid Zerocoin spends were being rejected (`InvalidChainFound` at block ~3,606,027). This was likely caused by a `prune=` setting in the test config removing block data that accumulator validation requires. However the fix in this PR forcing `ValidateAccumulatorCheckpoint()` whenever Zerocoin spends are present is the correct approach regardless, as it ensures correctness even in edge cases.

### 1. Zerocoin DB fast-path
- `WriteBlockZerocoinData()` returns immediately when all maps are empty
- Outer guard in `ConnectBlock` skips the call entirely when nothing to write
- Most post-Light Zerocoin blocks have no Zerocoin data, this saves real work on every such block

### 2. Accumulator validation optimization
- **Pre-Light Zerocoin**: unchanged, validates every block
- **Post-Light Zerocoin**:
  - Skips entirely when no spends and not a checksum boundary block
  - Forces validation when Zerocoin spends are present (the critical correctness fix)
  - Only writes `DatabaseChecksums()` on checksum boundary blocks

### 3. Safety and compatibility
- No consensus logic altered
- Zerocoin staking fully supported
- RingCT behavior unchanged
- No configuration flags required

## Performance Results

Tested on M4 Mac Mini (Apple Silicon) syncing from a Dec 2025 snapshot (~232k blocks):

| Metric | #1051 (current master) | This PR | Improvement |
|--------|----------------------|---------|-------------|
| Connect block | 11.44 ms/blk | 8.39 ms/blk | ~27% |
| Writing zerocoin | 3.77 ms/blk | 2.17 ms/blk | ~42% |

The improvement is most visible on blocks with no Zerocoin activity, where the accumulator is skipped entirely and zerocoin DB write cost drops from ~3.77ms to ~0.04ms per block.

## Testing

- Synced ~232k blocks from Dec 2025 snapshot on M4 Mac Mini
- No `InvalidChainFound` errors
- No Zerocoin spend validation failures
- Benchmarked with `debug=bench` in `veil.conf`

## Notes

This PR is designed to benefit all users automatically. No configuration flags or user action required.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -778,6 +778,13 @@ bool CZerocoinDB::WriteBlockZerocoinData(
     const uint256& hashBlock,
     bool fWritePubcoinSpends)
 {
+    // Fast path: nothing to write for this block
+    if (spendInfo.empty() &&
+        mintInfo.empty() &&
+        (!fWritePubcoinSpends || mapPubcoinSpends.empty())) {
+        return true;
+    }
+    
     CDBBatch batch(*this);
     size_t countSpends = 0;
     size_t countMints = 0;
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2864,14 +2864,56 @@ bool CChainState::ConnectBlock(const CBlock& block, CValidationState& state, CBl
                          REJECT_INVALID, "bad-cb-amount");
     }
 
-    // Ensure that accumulator checkpoints are valid and in the same state as this instance of the chain
-    int64_t nTimeAccumulate = GetTimeMicros();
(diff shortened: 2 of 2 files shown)


seanPhill (approved):
👍🏻 Nice work! Thanks for the testing!

PR #1058, opened 2026-06-03
Fix/spendzerocoinmints help 1045

### LINKED ISSUE #1045: CLI help for spendzerocoinmints correction needed
`help spendzerocoinmints` says that "address" is optional, and that if a target address is omitted it will spend to a new change address.

```
Spend zerocoin mints to a VEIL address.

Requires wallet passphrase to be set with walletpassphrase call.

Arguments:
1. mints_list     (string, required) A json array of zerocoin mints serial hashes
2. "address"     (string, optional, default=change) Send to specified address or to a new change address.
```

When tested **accidentally**, `spendzerocoinmints '["TheZerocoinSerialHash"]'` instead gave
`No address provided (code -4)`

which is perfectly fine by me!

Let's just correct the help output.

ohcee:
Fixes #1045
The help text incorrectly described address as optional with a default change address fallback. In practice, omitting it returns -4: No address provided. This corrects both the usage line and the argument description to reflect actual behavior.

### DIFF
--- a/src/wallet/rpczerocoin.cpp
+++ b/src/wallet/rpczerocoin.cpp
@@ -703,13 +703,13 @@ UniValue spendzerocoinmints(const JSONRPCRequest& request)
     UniValue params = request.params;
     if (request.fHelp || params.size() < 1 || params.size() > 2)
         throw std::runtime_error(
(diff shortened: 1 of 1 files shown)


seanPhill (approved):
👍🏻

PR #1062, opened 2026-06-29
Reject stealth and RingCT coinbase outputs after rollback height

nik343:
Veil's code had a critical vulnerability that allowed malformed coinbase blocks to be accepted starting at height 3890103. These blocks contained stealth outputs in the coinbase transaction, which should never have been accepted by consensus.

The network stall happened between block 3890306 and 3890307 from 2026-06-28 07:42:30 UTC to 2026-06-28 17:35:08 UTC (according to my personal node), due to an attempt to spend those outputs to basecoin. The mempool accepted the transactions, but block validation saw a supply increase and rejected the block template. Nodes with poisoned mempools could not produce valid blocks.

Last clean block:

`3890102`
`5651c2cf36c6178512afe27340db440064020a2d7dae58d8513ee12aa45a7c90`

Existing synced nodes will still need to invalidate the first bad block or reindex/reorg onto the corrected chain:

```
invalidateblock 913780c878021677e2c5fc471016c8db5e1acf5e705ed48c413525bff85009fc
clearmempool
```

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -331,6 +331,7 @@ public:
                 { 706000, uint256S("0xf98ece86f185e15af9a9d6b554c54f36f0b8b19d11e5e98ffe3b7578a0c8e2f9")},
                 { 840000, uint256S("0x9b46be33e4a84456e7c4e4785bad8646cb7cf1b6192adfd4c2916e768254f621")},
                 { 1744442, uint256S("0xe464d7547e1a21fe2471ddc4a6da535e2e9011ed570ab687897c7b7c7d9ed83f")},
+                { 3890102, uint256S("0x5651c2cf36c6178512afe27340db440064020a2d7dae58d8513ee12aa45a7c90")},
             }
         };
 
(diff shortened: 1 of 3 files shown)


seanPhill (commented):
utACK
Will be testing it in a minute or two, when the Actions have completed.

seanPhill (commented):
I've run it on macOS and it is waiting at the target block. Started up fine.

I had done
```
invalidateblock 913780c878021677e2c5fc471016c8db5e1acf5e705ed48c413525bff85009fc
clearmempool
```
before shutting down the previously running node.

I will be proceeding on several Linux nodes, not waiting for `master`.

The Block Explorer may be on the rejected chain until it is restarted with this PR.

seanPhill (approved):
ACK

Started in macOS and in Linux. The node is waiting at the target block and refusing the invalid block, but that would have happened anyway due to previously invalidating the bad block before shutting down the previous wallet.

However, I have also seen a node where attempting to reconsider the bad block has it rejected for the intended reason:
`ERROR: ConnectBlock(): coinbase contains stealth or RingCT outputs`

ohcee (approved):
Tested and confirmed ACK.

Fresh-sync on this build (node ~15k blocks behind, no manual steps): synced up and stopped exactly at 3890102, getbestblockhash = 5651c2cf...a45a7c90, refused to follow 3890103+.
Checkpoint hash confirmed on node.
Activation height 3890103 matches the audit (first poisoned block).

PR #1064, opened 2026-07-10
[Staking][Zerocoin][GUI] Show confirmations until staking becomes active (revives #913)

ohcee:
# [Staking][Zerocoin][GUI] Show confirmations until staking becomes active (revives #913)

**Draft — stacked on #1063.** This branch is based on the dark-mode branch, so the diff shows those commits too until #1063 merges; the change under review here is only the top commit (`[Staking][Zerocoin][GUI] Show confirmations until staking becomes active`).

Revives #913 by @Rock-N-Troll (his authorship is preserved on the commit), rebased onto current master. Issue reference: #889.

## What it does

The staking toggle in the status bar currently gives no feedback about *why* staking isn't active. This adds `CzTracker::GetConfirmationsRemainingForStaking()`:

- `-2` — tracker not yet initialized → "Fetching zerocoin status..."
- `-1` — no zerocoin in the wallet → "You need some zerocoin in order to stake"
- `0` — at least one zerocoin mature enough to stake → "Enabling..."
- `N` — confirmations remaining until the most mature zerocoin can stake → "Staking requires N more confirmations", updating as blocks arrive

Also clears `mapHashedBlocks` when the staking loop idles without stakeable coins (edge case from the original PR discussion: the wallet spends its only mature zerocoin while the staking loop is running).

## Changes from the original #913

- Rebased across five years of master: `MilliSleep` → `UninterruptibleSleep` in miner.cpp; in `veilstatusbar.cpp`, kept the unlock-for-staking dialog flow from #924 (which merged after #913 was written and supersedes its toast-only handling of locked wallets).
- Both indentation issues from the original review (miner.cpp, veilstatusbar.cpp) are fixed.

## Testing

- Builds on macOS (Apple Silicon); CI matrix on this PR covers the rest
- Functional testing on testnet in progress: mint zerocoin → watch the countdown → confirm "Enabling..." / "Staking Enabled" once mature

All credit to **@Rock-N-Troll** for the feature. Review feedback welcome — will mark ready for review after zerocoin countdown testing completes on testnet.

PR #1067, opened 2026-07-28
Guard null wallet and zerocoin tracker in staking status UI

ohcee:
updateStakingCheckbox() and setStakingText() dereferenced GetMainWallet() and GetZTrackerPointer() without checking either. GetMainWallet() returns an empty shared_ptr when the main wallet is not resolved, and GetZTrackerPointer() returns zTracker.get() on a unique_ptr that is null until the tracker is constructed, so either can segfault from an ordinary GUI status refresh.

Route both call sites through a helper returning -1 when the wallet or tracker is unavailable. -1 is what the tracker itself returns when the wallet holds no zerocoin, so the checkbox reads as not-ready rather than ready, and the existing "You need some zerocoin in order to stake" branch already handles that value.

setStakingText() also called pwallet->IsStakingActive() unguarded; that one is fixed inline since it's a plain wallet deref rather than the tracker chain.

The tracker dereferences came in with #1064. Worth landing before #1053 rebases onto master, since that branch is being tested on a multi-wallet mainnet setup with no zerocoins.

### DIFF
--- a/src/qt/veil/veilstatusbar.cpp
+++ b/src/qt/veil/veilstatusbar.cpp
@@ -69,12 +69,25 @@ void VeilStatusBar::onBtnSyncClicked(){
 
 #ifdef ENABLE_WALLET
 bool fBlockNextStakeCheckSignal = false;
+
+// Confirmations remaining before staking can begin, or -1 when the wallet or its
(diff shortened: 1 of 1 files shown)


seanPhill (approved):
Looks good. Thanks!

PR #1070, opened 2026-08-04
PoS tx inclusion, watch-only upgrades and stealth import, #1053 rebased onto master with fixes and tests

ohcee:
This carries #1053 forward so it can merge. It is @blondfrogs' branch, rebased onto current master (post-1.4.3.0, includes #1062), plus the two fixes from blondfrogs#2 and a regression test suite. All of blondfrogs' commits are preserved with his authorship.

**From #1053 (blondfrogs):**
- Fix for PoS blocks silently dropping the first mempool transaction (present since Oct 2018 — every PoS block on mainnet has been affected)
- Watch-only status/indexing fixes and `importstealthkeys` support

**Added here:**
- Fix a null dereference when assembling PoS block templates: the dummy coinstake placeholder reserved `vtx[1]` before the coinstake ran, and `CBlock::PowType()` → `IsProofOfStake()` dereferences `vtx[1]` whenever `vtx.size() > 1`. Every PoS template assembly segfaulted once the wallet had a stakeable zerocoin. The placeholder now goes in after coinstake creation, and `IsProofOfStake()` null-checks `vtx[1]`.
- Fix watch-only indexing, erasure and removal count; `importstealthkeys` now rescans the chain so pre-import funds are found.
- `src/test/pos_placeholder_tests.cpp` and `src/test/watchonly_tests.cpp` lock both fixes against regression (these ran green on blondfrogs#2, CI 8/8).

**Testing:** deterministic regtest repro of the staking crash, 3-node regtest matrices for staking A/B, watch-only and stealth-import flows, live testnet validation on 2026-07-27 confirming staked blocks now include mempool transactions, and community testnet nodes currently running these fixes. @seanPhill additionally verified no crashes across Windows/Linux with these fixes (blondfrogs#2 thread).

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -62,6 +62,8 @@ BITCOIN_TESTS =\
   test/multisig_tests.cpp \
   test/net_tests.cpp \
   test/netbase_tests.cpp \
+  test/pos_placeholder_tests.cpp \
+  test/watchonly_tests.cpp \
   test/pmt_tests.cpp \
   test/policyestimator_tests.cpp \
   test/pow_tests.cpp \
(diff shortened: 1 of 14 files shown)


seanPhill (approved):
ACK

PR #1071, opened 2026-08-07
Bootstrap repair: replace fork era fixed seeds (still Bitcoin's list) and prune dead DNS seeders

ohcee:
Right now Veil's ability to onboard a new user rests on one A record. If `node03.veil-project.com` stops resolving, a freshly installed wallet finds zero peers, never syncs, and doesn't tell you why. There's no fallback either, because the fallback has never worked.

A node has two ways to find its first peer: the DNS seeds in `chainparams.cpp`, and the fixed seeds compiled into the binary for when DNS fails. I tested both today, here's what I found.

**Mainnet DNS seeds, 1 of 14 working:**

| Hostname | Resolves to | P2P on 58810 |
|---|---|---|
| node01 | 172.67.146.205 | no, it's a Cloudflare proxy address so it can't carry P2P |
| node02 | 213.136.83.170 | no |
| **node03** | **209.126.8.91** | **yes** |
| node04 | 213.136.88.79 | no |
| node05 through node11 | no DNS record | |
| `veilseed.codeofalltrades.com` | no DNS record | |
| `veilseed.veil-stats.com` | no DNS record | |
| `veil-seed.pontificatingnobody.com` | nameservers unreachable | |

**Testnet DNS seeds, 0 of 11 working.** `testnode01` and `testnode02` only resolve to Cloudflare proxy addresses. `testnode03` through `testnode09` and both third party hostnames have no DNS record at all. So testnet has no bootstrap path whatsoever, which probably explains a lot about how painful it is to get a testnet node going.

**Fixed seeds, never worked once.** `pnSeed6_main` has 1,268 entries and every single one is on port 8333. That's Bitcoin's port, not Veil's 58810. `contrib/seeds/nodes_main.txt` is Bitcoin v0.17.0's seed list, carried over at the fork and never regenerated, so the compiled in fallback has never been able to reach a Veil node. I checked those addresses today and they're dead on 58810 and on 8333 too. They're seven year old Bitcoin nodes.

**Why this went unnoticed:** `contrib/seeds/generate-seeds.py` quietly assigns port 8333 to any entry written without an explicit port, so a stale list regenerates into a header that looks perfectly fine. And since node03 kept answering, bootstrap looked like it worked. You only find out the day the last DNS seed goes down, and then new installs stop working everywhere at once with no fix short of a release.

This PR fixes both layers, and adds a DNS seed that keeps itself current so the list can't quietly rot again.

**Changes:**
- `contrib/seeds/nodes_main.txt`: replaced with 10 mainnet nodes, each verified with a full `version`/`verack` handshake against mainnet magic, all at chain tip when checked (user agents 1.4.1 through 1.4.2.99)
- `contrib/seeds/nodes_test.txt`: one verified testnet node, since testnet had nothing
- `src/chainparamsseeds.h`: regenerated with `contrib/seeds/generate-seeds.py`
- `src/chainparams.cpp`: dropped the third party seeder hostnames from mainnet, testnet and devnet. The `codeofalltrades.com`, `veil-stats.com`, `pontificatingnobody.com` and `asoftwaresolution.com` subdomains are all NXDOMAIN or have dead nameservers. Every `veil-project.com` hostname is kept.
- `src/chainparams.cpp`: added one mainnet DNS seed, `seed.veil-info.org`, see below

**A DNS seed that keeps itself current**

The reason this list decayed to one working entry is that every seed is a hand maintained A record. `seed.veil-info.org` works differently. It's answered by [Veil-Project/Veil-DNS-Seeder](https://github.com/Veil-Project/Veil-DNS-Seeder), which crawls the network, tests what it finds, tracks availability over 2h/8h/1d/7d windows, and only hands out addresses that recently completed a real handshake. Dead nodes age out on their own and new ones get discovered without anyone touching a zone file.

That crawler hadn't been able to reach a Veil node since May 2021. Its `PROTOCOL_VERSION` was 70028 while nodes enforce 70029 after `nTimeKIfork`, so every handshake got rejected. Fixes are in a companion PR to that repo. It's been running since 2026-08-07:

```
$ dig +short seed.veil-info.org
169.58.67.195
62.72.43.203
209.126.9.235
57.131.32.128
```

Its first pass confirmed 9 reachable nodes out of 53 addresses tried. One of them (`109.123.233.94`) wasn't in any hand built list, and it's in the fixed seeds here.

I run this host. If the project would rather own it, happy to hand the seeder over, help stand up a second instance somewhere for redundancy, or just drop this hostname and keep the fixed seed changes. Whatever you prefer. A crawler backed seed under `veil-project.com` would be the better long term home anyway, and the same delegation works there.

Ports are written out explicitly in `nodes_*.txt` because `generate-seeds.py` defaults bare entries to 8333/18333, which is exactly how the Bitcoin list survived the fork unnoticed. One mainnet entry runs on a custom port and `SeedSpec6` carries that per entry.

**Verifying:** every listed IP answers a P2P handshake. Quick check:

```
while read n; do nc -vz -G 3 ${n%:*} ${n#*:}; done < contrib/seeds/nodes_main.txt
```

**One thing worth doing that isn't code, and it helps more people than this PR does:** everything here only reaches users who run a build that has it. Every wallet already installed keeps querying `node01` through `node11.veil-project.com`. Whoever holds that DNS can fix bootstrap for all of them today without waiting for a release:

- point the dead `node05` through `node11` records at live nodes (DNS only, never proxied), or
- `CNAME` one of them to `seed.veil-info.org` so existing binaries pick up crawler maintained answers automatically

Also worth noting `node01` and `testnode01` currently resolve to Cloudflare proxy addresses, which can't carry P2P traffic. Those two want un proxying either way.

---

For release notes: fresh installs can find peers with no manual `addnode`, even if all the seed DNS is down.


contrib/seeds/nodes_test.txt
seanPhill, line 1: We should probably set up some onion seeders.

seanPhill (approved):
ACK

